PR #9, opened 2025-10-12
Fix session-start hook execution in plugin context

ryancnelson:
## Problem
The `session-start.sh` hook fails silently when executed by Claude Code with error "⏺ Plugin hook error:" preventing skills context from loading.

## Root Cause
1. `BASH_SOURCE[0]` is unbound in Claude Code's execution context (fails with `set -u`)
2. Empty `grep` results return exit code 1 (fails with `set -e`)

## Solution
1. Use `${BASH_SOURCE[0]:-$0}` to fallback to `$0` when `BASH_SOURCE` unavailable  
2. Add `|| true` to handle empty grep results gracefully

## Testing
✅ Direct execution produces valid JSON  
✅ Claude Code plugin execution now works  
✅ No breaking changes to existing functionality

## Changes
- `hooks/session-start.sh`: 2 lines changed
  - Line 10: Add `${BASH_SOURCE[0]:-$0}` fallback
  - Line 18: Add `|| true` for empty grep results

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured reliable startup by correctly resolving the script location across environments, even when path variables are unset.
  * Prevented errors during initialization when clearing update/status flags that may not exist, allowing sessions to proceed smoothly.

* **Chores**
  * Hardened session initialization for improved robustness and resilience.

### DIFF
--- a/hooks/session-start.sh
+++ b/hooks/session-start.sh
@@ -7,7 +7,7 @@ set -euo pipefail
 export SUPERPOWERS_SKILLS_ROOT="${HOME}/.config/superpowers/skills"
 
 # Run skills initialization script (handles clone/fetch/auto-update)
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
 PLUGIN_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
 init_output=$("${PLUGIN_ROOT}/lib/initialize-skills.sh" 2>&1 || echo "")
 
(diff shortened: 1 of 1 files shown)


ryancnelson:
this PR is all ai-generated, btw.... i just asked claude/superpowers to help debug the issue.

ericphanson:
I'm also hitting the empty grep results issue here when using a fork of the skills directory. As summarized by Claude code:

When a user is working on a fork with local commits ahead of the tracking branch, initialize-skills.sh outputs only
  SKILLS_BEHIND=true with no user-facing messages. The session-start hook then filters out status flags with grep -v,
  which leaves empty output. Without || echo "", grep exits with code 1 when all lines are filtered out, causing the
  script to fail due to set -euo pipefail. This prevents the hook from executing successfully.

(This PR uses || true instead of || echo "")

PR #11, opened 2025-10-12
Fix SessionStart hook hangs on Claude Code startup

ryancnelson:
## Problem

The session-start.sh hook was causing Claude Code to hang on startup, displaying "⏺ Plugin hook error:" and preventing the superpowers context from being injected.

### Symptoms
- Hook execution hangs indefinitely during Claude startup
- Error message displayed to user
- Skills list and documentation not loaded
- `/doctor` reports context/memory issues

### Root Causes

1. **sed/awk JSON escaping hang**: The complex sed/awk pipeline used for JSON escaping would hang when processing ~5KB of text on some systems

2. **grep command hang**: The `grep` commands for extracting status flags would mysteriously hang when executed within command substitution in Claude's hook execution environment, even though they ran fine in a normal shell

## Environment Where Bug Was Triggered

- **OS**: macOS 15.6 (Darwin 24.6.0)
- **Hardware**: Mac Mini M2 Pro
- **Claude Code**: v2.0.14
- **Shell**: Bash 3.2.57
- **Home directory**: External SSD (`/Volumes/T9/`)

The hang occurred specifically during Claude's hook execution context, where stdin/stdout/stderr handling differs from normal shell execution. This is likely why the issue didn't manifest in manual testing but only during Claude startup.

### 1. Replace sed/awk with jq for JSON escaping

**Before:**
```bash
escaped=$(echo "$text" | sed 's/\\/\\\\/g' | sed 's/"/\\"/g' | awk '{printf "%s\\n", $0}')
```

**After:**
```bash
escaped=$(echo "$text" | jq -Rs .)
```

**Benefits:**
- More reliable (dedicated JSON tool)
- ~10x faster
- Handles all edge cases correctly
- Won't hang on large inputs

### 2. Wrap grep in timeout-protected subshells

**Before:**
```bash
skills_behind=$(echo "$init_output" | grep "SKILLS_BEHIND=true" || echo "")
```

**After:**
```bash
skills_behind=$(timeout 1 bash -c "echo '$init_output' | grep 'SKILLS_BEHIND=true'" 2>/dev/null || echo "")
```

This isolates the grep command and prevents hangs due to pipe buffering or process group issues in Claude's execution environment.

### 3. Add timeout protection to all external commands

```bash
init_output=$(timeout 10 "${PLUGIN_ROOT}/lib/initialize-skills.sh" 2>&1 || echo "")
find_skills_output=$(timeout 5 "${SUPERPOWERS_SKILLS_ROOT}/skills/using-skills/find-skills" </dev/null 2>&1 || echo "⚠️ find-skills timed out")
using_skills_content=$(timeout 2 cat "${SUPERPOWERS_SKILLS_ROOT}/skills/using-skills/SKILL.md" 2>&1 || echo "Error reading using-skills")
```

## How Superpowers Helped Debug This

This bug was tracked down using the **skills/debugging/systematic-debugging** approach:

1. **Phase 1 - Reproduce**: Confirmed the hook hung during Claude startup but worked when run manually
2. **Phase 2 - Isolate**: Added diagnostic logging to identify the exact hang location (Step 4: Extract flags)
3. **Phase 3 - Root Cause**: Tested components in isolation vs. hook environment, discovering grep's environment-specific behavior
4. **Phase 4 - Fix & Verify**: Applied targeted fixes and verified in production environment

The systematic debugging workflow helped avoid jumping to solutions and ensured we found the actual root cause rather than just treating symptoms.

## Testing

Verified on the environment where the bug originally occurred:

✅ Hook completes in <3 seconds (was hanging indefinitely)
✅ No error messages on startup
✅ Full skills list successfully injected (~11KB of context)
✅ Debug log shows `Errors: 0`
✅ `/doctor` no longer reports issues

## Additional Notes

This fix maintains backward compatibility while adding robustness. Systems that weren't experiencing the hang will see improved performance from the jq optimization.

The timeout values are conservative and should work across different system speeds and network conditions for the git fetch operations in initialize-skills.

## Summary by CodeRabbit

- New Features
  - Provides a richer session context for tools and skills, improving initialization reliability.
  - Adds clear status messaging when local skills are behind upstream.

- Bug Fixes
  - Prevents startup hangs with timeouts during initialization and skill discovery.
  - Ensures non-blocking reads of skill metadata and consistent session completion.

- Refactor
  - Streamlines session startup flow and consolidates status/context output into a single payload for more predictable behavior.

### DIFF
--- a/hooks/session-start.sh
+++ b/hooks/session-start.sh
@@ -1,51 +1,60 @@
 #!/usr/bin/env bash
-# SessionStart hook for superpowers plugin
+# SessionStart hook for superpowers plugin - SAFE VERSION
 
 set -euo pipefail
 
 # Set SUPERPOWERS_SKILLS_ROOT environment variable
 export SUPERPOWERS_SKILLS_ROOT="${HOME}/.config/superpowers/skills"
 
-# Run skills initialization script (handles clone/fetch/auto-update)
+# Run skills initialization script with timeout (handles clone/fetch/auto-update)
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PLUGIN_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
-init_output=$("${PLUGIN_ROOT}/lib/initialize-skills.sh" 2>&1 || echo "")
+init_output=$(timeout 10 "${PLUGIN_ROOT}/lib/initialize-skills.sh" 2>&1 || echo "")
 
-# Extract status flags
-skills_updated=$(echo "$init_output" | grep "SKILLS_UPDATED=true" || echo "")
-skills_behind=$(echo "$init_output" | grep "SKILLS_BEHIND=true" || echo "")
-# Remove status flags from display output
-init_output=$(echo "$init_output" | grep -v "SKILLS_UPDATED=true" | grep -v "SKILLS_BEHIND=true")
+# Extract status flags with timeout protection
+skills_updated=$(timeout 1 bash -c "echo '$init_output' | grep 'SKILLS_UPDATED=true'" 2>/dev/null || echo "")
+skills_behind=$(timeout 1 bash -c "echo '$init_output' | grep 'SKILLS_BEHIND=true'" 2>/dev/null || echo "")
 
-# Run find-skills to show all available skills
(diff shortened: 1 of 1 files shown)


obra:
Thanks so much for these PRs. I won't get to prod until tomorrow, though I'd love reports from other users.  Between the two PRs,  which one do I want?

obra:
I ended up running with PR #9

PR #57, opened 2025-10-22
Add explicit skills location context to session-start hook

### LINKED ISSUE #49: Clean reinstall of plugin but skills broken
**Describe the bug**
I've been unable to get the skills to work correctly since upgrading to the latest release, and am unable to figure out how to resolve this.

**To Reproduce**
Things have gotten borked and I can't figure out how to un-bork them for the life of me...

Here's how I'm trying to get a clean start:
1. Uninstall the superpowers marketplace (and its plugins). Exit Claude Code.
2. Clear the plugins cache: `rm -r ~/.claude/plugins/cache/`.
3. Remove the maybe-old install location for superpowers: `rm -r ~/.config/superpowers/`.
4. Re-install the superpowers marketplace and plugin: `/plugin marketplace add obra/superpowers-marketplace` and then `/plugin install superpowers@superpowers-marketplace`.

So. That should be a clean start, right?

Every time I try to run the brainstorm command, it gets very confused and I think starts _pretending_ to run the skill based only off the text in the `/superpowers:brainstorm` command's prompt:
```
> /superpowers:brainstorm is running…

● I'm using the brainstorming skill to help refine your idea into a fully-formed design.

  To get started, I need to understand what you'd like to create or develop. What rough idea, feature, or problem would
  you like to brainstorm about?

  Some examples from your repository context that we could explore:
  - A new financial workflow or automation
  - Improvements to existing transaction matching systems
  - New analysis or reporting features
  - Data integration enhancements
  - Or something completely different!

  What would you like to work on?

> I'd like to resolve GitHub Issue #29.

● Let me fetch the details of GitHub Issue #29 to understand what we're working with.

● Bash(gh issue view 29 --json title,body,labels,comments)
  ⎿ {
      ...

● Read(~/.config/superpowers/skills/skills/design/brainstorming/SKILL.md)
  ⎿  Error reading file

● Search(pattern: "**/*brainstorming*", path: "~/.config/superpowers")
  ⎿  Error searching files

● Perfect! I can see Issue #29 is about...
```

It is very firmly convinced that `~/.config/superpowers/` should exist, and I don't know why. Or, if it **is** supposed to exist, I don't know why the plugin install isn't _doing_ that.

**Expected behavior**

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**

rentorm:
## Summary

Provides the actual skills directory path to Claude at session start, preventing confusion about where skills are located.

## Problem

Fixes #49 - After a clean reinstall, Claude incorrectly assumes skills are in `~/.config/superpowers/` and tries to read from that location, causing errors and confusion.

## Solution

- Add `SKILLS_DIR` variable pointing to the actual plugin skills directory
- Include the skills directory path in the session context: "You have superpowers located in ${SKILLS_DIR}"
- Update all path references to use the variable for consistency

## Testing

Tested by starting a new session and verifying Claude receives the correct skills location context.

## Impact

This makes Claude explicitly aware of where skills are located, preventing it from looking in the wrong directory and reducing confusion during clean installs.

## Summary by CodeRabbit

* **Refactor**
  * Centralized and made the skills directory configurable, replacing hardcoded paths.
  * Updated content injection and contextual messaging to reference the configurable skills location.
  * Ensured path values are safely encoded for JSON and parameterized throughout for more reliable behavior and easier maintenance.

### DIFF
--- a/hooks/session-start.sh
+++ b/hooks/session-start.sh
@@ -21,13 +21,14 @@ using_superpowers_content=$(cat "${SKILLS_DIR}/using-superpowers/SKILL.md" 2>&1 
 # Escape outputs for JSON
 using_superpowers_escaped=$(echo "$using_superpowers_content" | sed 's/\\/\\\\/g' | sed 's/"/\\"/g' | awk '{printf "%s\\n", $0}')
 warning_escaped=$(echo "$warning_message" | sed 's/\\/\\\\/g' | sed 's/"/\\"/g' | awk '{printf "%s\\n", $0}')
+skills_dir_escaped=$(echo "$SKILLS_DIR" | sed 's/\\/\\\\/g' | sed 's/"/\\"/g')
 
 # Output context injection as JSON
 cat <<EOF
 {
   "hookSpecificOutput": {
     "hookEventName": "SessionStart",
-    "additionalContext": "<EXTREMELY_IMPORTANT>\nYou have superpowers located in ${SKILLS_DIR}.\n\n**The content below is from ${SKILLS_DIR}/using-superpowers/SKILL.md - your introduction to using skills:**\n\n${using_superpowers_escaped}\n\n${warning_escaped}\n</EXTREMELY_IMPORTANT>"
+    "additionalContext": "<EXTREMELY_IMPORTANT>\nYou have superpowers located in ${skills_dir_escaped}.\n\n**The content below is from ${skills_dir_escaped}/using-superpowers/SKILL.md - your introduction to using skills:**\n\n${using_superpowers_escaped}\n\n${warning_escaped}\n</EXTREMELY_IMPORTANT>"
   }
 }
 EOF


obra:
I'm super curious about this one. 
I *suspect* that what you're seeing is claude-code somehow caching an old version of the plugin.
Nowhere in what we're shipping should reference ~/.config/superpowers anymore. Do you know where Claude is getting that? We *shouldn't* need the patch to explicitly point to the plugin's directory--and that's going to break things for skills that come from other plugins.

dominicm00:
@obra at least for me when running the latest native version of Claude code, it seems like it has no path information on where skill files are and hallucinates a different path each run.

Claude code exposes a `Skill` tool which the model is expected to use & does correctly read the skill content, but since the bootstrap specifically instructs to use the `Read` tool, it gets confused. So really the fix here should be to instruct it to use the correct tool (or not tell it to use a specific tool at all and instead trust that Claude's system prompt will say how to use skills).

Was the addition of this `Skill` tool perhaps recent?

rentorm:
@obra yes, I suspect it might be that when there is reference in settings.json or perhaphs even in session log files in jsonl somewhere to the old  ~/.config/superpowers, claude-code somehow picks that up. 

You have a good point about not needing to explicitly pointing to plugins location in the prompt, which potentially can interfere with other plugins.

obra:
@dominicm00 indeed, the Skill tool postdates the creation of superpowers. Lemme go see about cleaning up the bootstrap

obra:
I'm closing this out, since we now tell claude to use the Skill tool instead of Read and any remaining ~/.config/superpowers/skills references have been removed.

dvic:
> I'm super curious about this one. 
> 
> I *suspect* that what you're seeing is claude-code somehow caching an old version of the plugin.
> 
> Nowhere in what we're shipping should reference ~/.config/superpowers anymore. Do you know where Claude is getting that? We *shouldn't* need the patch to explicitly point to the plugin's directory--and that's going to break things for skills that come from other plugins.   
> 
> 
> 
> 

In my case Claude replied to me that it was due to the saved Read permission to that old path, could that be it?

PR #105, opened 2025-11-13
Base worktrees on main branch and support background setup

neilberkman:
## Changes

- Always base new worktrees on main/master branch by default (auto-detects which)
- Document background setup optimization for long-running deps/compile
- Add Elixir project support

## Problem

Without specifying a base branch, `git worktree add` inherits the current branch. When on a feature branch, new worktrees are incorrectly based on that feature branch instead of main.

## Solution

Detect repository's main branch automatically and use it as base. Allow user override when explicitly specified.

## Summary by CodeRabbit

* **Documentation**
  * Updated Git worktrees guide with explicit base-branch detection and configurable override (defaults to origin/main or origin/master)
  * Reordered flow: detect base branch, create worktree, run project setup; added baseline verification and location reporting
  * Added multi-ecosystem project setup (Node, Rust, Python, Go, Elixir) and background-setup optimization examples
  * Expanded examples, quick-reference table, red flags, common mistakes, and an end-to-end example workflow

### DIFF
--- a/skills/using-git-worktrees/SKILL.md
+++ b/skills/using-git-worktrees/SKILL.md
@@ -85,7 +85,7 @@ project=$(basename "$(git rev-parse --show-toplevel)")
 
 **CRITICAL: Always base worktrees off the main branch unless explicitly stated otherwise.**
 
-By default, `git worktree add` bases the new branch on whatever is currently checked out. This is almost never what you want when on a feature branch!
+By default, `git worktree add` bases the new branch on whatever is currently checked out. This is rarely what you want when on a feature branch!
 
 ```bash
 # Detect the repository's main branch


evaneaston:
I'll provide an opinion, on

> Without specifying a base branch, git worktree add inherits the current branch. When on a feature branch, new worktrees are incorrectly based on that feature branch instead of main.

I would not say it is _incorrect_.  Staging work on previous, un-merged work can be really useful when there are dependencies.  It allows you to continue forward movement on dependent features right away, while awaiting review of your work.  And then you might be less inclined to keep adding things to a kitchen-sink of a PR. It is also used by some teams to have collaborators contribute changes to a PR that aren't easily expressed in PR reviews.

Whether you do this depends:
 - on your collaborators being willing to work in this way and manage PR merge targets a little differently ‡
 - on how confident you are that your initial work is not going to need significant revision.

IMO, I wouldn't mind if the skill detected that you are not on the main branch and **asked you** what you wanted to do.

____
‡ Github handles stacked PRs automatically regardless of merge order.

obra:
Indeed, I actively want the default to be branching from the current branch, not from main.

neilberkman:
> Indeed, I actively want the default to be branching from the current branch, not from main.

The issue I'm trying to solve is: I'm working with Claude Code in a working directory, a new issue comes up that I want to tackle in a worktree and I end up dragging along unrelated WIP. This happens to me all the time with the skill as it is now. Seems like this may not be an issue that others are having so feel free to close it out and I'll stick with my fork.

obra:
@neilberkman workflow-wise, I tend to keep a main tree around for new task branches, rather than *starting* from a worktreee.

PR #115, opened 2025-11-16
fix: yaml error in code-reviewer agent

onelesd:
Fixes a yaml parser error in agents/code-reviewer.md

## Motivation and Context

It was preventing me from reusing this agent in opencode.

## How Has This Been Tested?

In opencode.

## Breaking Changes

None.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

## Summary by CodeRabbit

* **Chores**
  * Updated internal documentation formatting for consistency.

---

**Note:** This release contains no user-facing changes or new functionality.

PR #116, opened 2025-11-19
add opencode installer script

onelesd:
This installer script copies the skills, agents and commands into opencode's expected dir structure.

When used with https://github.com/malhashemi/opencode-skills the result is a superpowers-capable opencode.

It is meant to be run once, and then whenever the superpowers repo is updated with a `git pull`. The installer will copy new skills if they are found and prompt to overwrite or ignore skills which have changed.

## Motivation and Context

I wanted to use superpowers in opencode.

## How Has This Been Tested?

Tested with both a fresh opencode install and an existing install.

## Breaking Changes

They will need to use the opencode-skills plugin which is mentioned by the installer.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added or updated documentation as needed

## Summary by CodeRabbit

* **New Features**
  * Added interactive installer script for OpenCode configuration management
  * Features automatic file change detection and comparison
  * Prompts users for conflict resolution when files are modified
  * Displays comprehensive summary of planned actions before execution
  * Supports flexible target directory selection with optional directory creation
  * Includes file-by-file outcome reporting during installation

### DIFF
--- a/install-opencode.sh
+++ b/install-opencode.sh
@@ -1,0 +1,348 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Superpowers OpenCode Installer
+# Syncs agents, commands, and skills to OpenCode config directory
+
+# Color codes for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+# Calculate checksum for a file
+calculate_checksum() {
+    local file="$1"
+    if [[ -f "$file" ]]; then
+        if command -v sha256sum &> /dev/null; then
+            sha256sum "$file" | awk '{print $1}'
+        else
+            shasum -a 256 "$file" | awk '{print $1}'
+        fi
+    else
+        echo ""
+    fi
+}
+
+# Get script directory (where superpowers is located)
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Default target directory (using test directory for safe development)
+DEFAULT_TARGET="$HOME/.config/opencode"
+
+echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+echo "  Superpowers OpenCode Installer"
+echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+echo ""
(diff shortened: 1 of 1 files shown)


jihchi:
Thank you! Great to see opencode support being added.

It would help to update README.md with instructions for getting started with the script.

aaryan-rampal:
Haha i did the same thing. I'll test yours and see if there's anything I did different which you might be able to

aaryan-rampal:
So I tried this, and it seems to break because Opencode is stricter with YAML formatting for the agents. This is how Claude put it after I fixed it: 
>  The YAML frontmatter in code-reviewer.md has an unquoted description field containing XML-like tags (<example>, <commentary>). YAML requires these special characters to be quoted. Fix by wrapping the
  entire description value in double quotes: description: "Use this agent when... <example>..."

Maybe your script should auto parse any agents which exist

Edit: Tested using Opencode 1.0.80

Edit 2: Moreover, the code reviewer `sonnet` model name is not valid. This probably needs to be either hardcoded or not set (to inherent the parent model the user is using on opencode) 

Edit 3: I see you already noticed this in https://github.com/obra/superpowers/pull/115. But the model problem still remains

onelesd:
> Edit 2: Moreover, the code reviewer `sonnet` model name is not valid. This probably needs to be either hardcoded or not set (to inherent the parent model the user is using on opencode)

the hacky thing to do would be to strip the model setting, or map them to opencode model names.

however, for this "dumb" installer it would be better if this was abstracted upstream before the installers for claude/codex/opencode see it. this is a bigger change that we'd need @obra to weigh in on.

obra:
Thanks so much for this PR.  Have any of the folks here looked at what it'd take to do it as a 'proper' opencode plugin using their install and extension mechanisms?  I worry a little bit about a copy-to-install kind of workflow.

The skill suite does need to be slightly better abstracted to support non-claude-code models and harnesses. For Codex, specifically with the GPT5 series models, we can get away with a custom bootstrap with a mapping table, but the right thing is probably to make *all* the skills cleaner and have a mapping table for each platform.   

I think the code-reviewer subagent should probably be retired in favor of 'generic subagent with prompting'

The other opencode PR that reuses most of the codex bootstrap looked generally pretty good to me, but also I've only been playing with opencode for the past couple of days.

obra:
I just landed a full opencode implementation (as a plugin, with tools). Install instructions are in the README. I'd love to know how it works for folks. (I'm closing out this PR for now)

PR #121, opened 2025-11-26
Add github-project-management skill

nickolasclarke:
## Summary

- Adds a new skill for integrating GitHub issue tracking into the development workflow
- Agent offers to create issues at natural checkpoints (after brainstorming, plans, bug discovery, branch completion)
- Always requires user confirmation before creating issues
- Includes CLI reference for `gh project` and `gh issue` commands

I think I'd like to re-write this to be a generic "Project Management" skill with instructions to use the appropriate reference file for the particular project management software you may be using in your project (gh projects, Jira, etc, simple gh issues only), etc. However, I think it would be good to do some extended play with this first with my known use-case of GH projects and iterate.

**Blocker** 
As written, I think this is too tightly coupled to be invoked and it should be entirely optional. I'm not sure if there are known, deterministic practices around selectively enabling or disabling skills within a claude "Skill". 

## Files

- `skills/github-project-management/SKILL.md` - workflow integration skill
- `skills/github-project-management/gh-reference.md` - CLI command reference
- `docs/plans/2025-11-25-github-project-management-design.md` - design document

## Test plan

- [x] Ran baseline tests without skill (agent doesn't offer issue creation)
- [x] Ran verification test with skill (agent now offers issue creation at checkpoints)
- [x] Real-world usage testing

## Summary by CodeRabbit

* **Documentation**
  * Added GitHub Project Management skill documentation including design specifications, integration guidelines, configuration instructions, and a comprehensive CLI command reference for managing issues and project items across the development workflow with user confirmation checkpoints.

### DIFF
--- a/docs/plans/2026-02-11-github-project-management-skill-review.md
+++ b/docs/plans/2026-02-11-github-project-management-skill-review.md
@@ -1,0 +1,369 @@
+# GitHub Project Management Skill — TDD Review
+
+## RED Phase: CLI Testing Results
+
+All `gh` CLI commands from `gh-reference.md` were tested against real project `nickolasclarke/1`.
+
+### Working Correctly
+- `gh auth status`, `project list/view/field-list/item-list`
+- `project item-create` (drafts), `item-edit`, `item-delete`
+- `issue create`, `issue create --project` (one-step)
+- All GraphQL sub-issues commands (`addSubIssue`, `removeSubIssue`, `listSubIssues`)
+
+### Issues Found
+
+1. **`item-create` returns NO OUTPUT on success** — agent won't know the item ID was created. The skill/reference should note this and suggest using `item-list --format json` after to verify.
+2. **`gh issue create` fails if the repo has issues disabled** — skill doesn't handle this case.
+3. **`--project` flag on `gh issue create` works by PROJECT NAME (not number)** — `gh-reference.md` shows both methods but could be clearer about when to use which.
+
+---
+
+## Analysis Against writing-skills Guidelines
+
+### Frontmatter/CSO
+
+**Current description (227 chars):**
+```yaml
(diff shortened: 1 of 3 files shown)


nickolasclarke:
woo, claude got over-eager and opened this on its own. Swapping to draft!

docs/plans/2025-11-25-github-project-management-design.md
nickolasclarke, line 1: I presume we probably don't want to keep these around, but...maybe we do?

obra:
So, I don't find claude to need help with doing this using the gh tool. can you tell me more about the use case or problem you're solving?

nickolasclarke:
@obra For one org I am part of, we use [gh projects](https://docs.github.com/en/issues/planning-and-tracking-with-projects/learning-about-projects/about-projects) to do our basic project management, especially since our work is not in a monorepo and various work may cut across multiple repos. I was finding I needed to keep prompting it on how to drive the issue tracking and roadmap that is in our gh project via the cli and decided at least a cli reference sheet was in order instead of it grepping around the cli `--help` output every time. 

I then found that this would pair well with a general skill around project management, and started throwing design docs, meeting notes for product prioritization, general product ideas, etc at it. I'm using this skill to do the grunt work of `brainstorming` on those (as needed), and then turning them into discrete issues, either at the project or repo level, depending on the nature of the task.

obra:
I'm sorry for the long radio silence here.  I don't actually want to tie deep github integration into superpowers. That said, as of 5.0, the internal instructions that tell Superpowers to prefer instructions from the user *should* mean that you can just put those instructions into your claude.md and everything will work.  If that's not the case, I'd love an issue report about what the let-the-user-override-superpowers instructions aren't currently letting you tweak.

PR #132, opened 2025-12-01
fix: add Windows compatibility for SessionStart hook

danshapiro:
## Summary

- Adds cross-platform Node.js implementation of `session-start` hook
- Fixes Windows path handling issue where backslashes were being stripped
- Node.js is guaranteed available since Claude Code runs on Node.js

## Problem

On Windows, the bash-based `session-start.sh` hook fails with:
```
/bin/bash:C:Usersdan.claudepluginscachesuperpowers/hooks/session-start.sh: No such file or directory
```

**Root cause:** `${CLAUDE_PLUGIN_ROOT}` expands to Windows paths with backslashes (e.g., `C:\Users\dan\...`). When passed through shell command parsing, backslashes are interpreted as escape characters and stripped, resulting in malformed paths like `C:Usersdan...`.

## Solution

Added `session-start.js` - a Node.js implementation that:
- Uses `path.join()` for proper cross-platform path handling
- Has identical functionality to the bash version
- Properly reads SKILL.md and checks for legacy skills directory
- Outputs the same JSON structure for hook injection

## Test plan

- [x] Tested on Windows with Git Bash environment
- [x] Hook executes without errors
- [x] Skill context properly injected at session start
- [x] Verified skills are accessible after startup

## Summary by CodeRabbit

* **New Features**
  * Session initialization now displays relevant skill documentation and contextual guidance upon startup.

* **Chores**
  * Updated internal hook execution mechanism for improved performance.
  * Added legacy configuration detection and notification.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/session-start.js\""
           }
         ]
       }
--- a/hooks/session-start.js
+++ b/hooks/session-start.js
@@ -1,0 +1,45 @@
+#!/usr/bin/env node
+// SessionStart hook for superpowers plugin
+// Cross-platform Node.js implementation (works on Windows, macOS, Linux)
+
+const fs = require('fs');
+const path = require('path');
+const os = require('os');
+
+// Determine plugin root directory
+const SCRIPT_DIR = __dirname;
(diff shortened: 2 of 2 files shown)


obra:
Thanks for working on this! We've just released v3.6.0 which includes Windows hook support via a polyglot wrapper approach.

Could you try updating the plugin and let us know if it resolves your Windows issues?

```
/plugin update superpowers@superpowers-marketplace
```

The new approach uses a `.cmd` wrapper that works on both Windows CMD and Unix bash, calling Git for Windows' bash.exe to run the actual hook scripts.

If this works for you, we can close this PR. If you're still seeing issues, please let us know!

PR #179, opened 2025-12-21
feat: Update brainstorming skill to use AskUserQuestion tool

smithclay:
This updates the brainstorming Skill to use Claude Code's built-in `AskUserQuestion` tool for nicer UX, see here:
image
 
The prompt is based how Anthropic instructs their `hookify` command to use it: https://github.com/anthropics/claude-code/blob/main/plugins/hookify/commands/configure.md

## Motivation and Context
- Uses Claude Code's native question asking UI for faster iteration: you can tab-select or select multiple options with fewer keystrokes.
- The `AskUserQuestion` has greater information density than standard text output (options and reasoning presented side-by-side)
- 
## How Has This Been Tested?
Yes, but only in Claude Code

## Breaking Changes
No

## Types of changes

- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added or updated documentation as needed

## Additional context
Is superpowers also targeting non-Claude Code environments? If so, how are other skills handing this sort of scenario, if at all (tool(s) only available in one environment)?

### DIFF
--- a/skills/brainstorming/SKILL.md
+++ b/skills/brainstorming/SKILL.md
@@ -15,15 +15,15 @@ Start by understanding the current project context, then ask questions one at a 
 
 **Understanding the idea:**
 - Check out the current project state first (files, docs, recent commits)
-- Ask questions one at a time to refine the idea
-- Prefer multiple choice questions when possible, but open-ended is fine too
+- Use the **AskUserQuestion tool** to ask questions one at a time
+- Prefer multiple choice questions when possible (2-4 options)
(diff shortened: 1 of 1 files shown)


obra:
We actually did this when AskUserQuestion first came out. I made the explicit decision to roll it back because it resulted in UX that caused users to just click "1" "1" "1" - And the whole point of brainstorming is to make the user think and engage.

PR #183, opened 2025-12-23
Feat/designing for autonomy

talsraviv:
Hey Jesse! Something I've noticed really sets things up for success is designing for autonomy/self-sufficiency. I've also enjoyed your writing on your blog about setting agents up for self-sufficiency with consolidated logs, and designing systems that are way more observable by the agent. 

I wanted to share an early draft and some thoughts of how this could be, and let me know if this is a direction that coheres with where you want to take this repository. 

## Motivation and Context
I find that agents can run for a lot longer when they not only can verify that their tests pass but also can verify that things are actually working in the real system that they are running.

## How Has This Been Tested?
I've been doing early validating of this over on my prototyping repository: https://github.com/talsraviv/from-thinking-to-coding. 

Still todo is systematic pressure testing (as in the skill writing skill). Sharing this early draft in case it's not the right direction, and I'm going to continue testing this by using my fork to try to build something and make sure the skill gets used. 

## Types of changes
- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [X] I have added or updated documentation as needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,9 +106,6 @@ Fetch and follow instructions from https://raw.githubusercontent.com/obra/superp
 - **systematic-debugging** - 4-phase root cause process (includes root-cause-tracing, defense-in-depth, condition-based-waiting techniques)
 - **verification-before-completion** - Ensure it's actually fixed
 
-**Architecture**
-- **designing-for-autonomy** - Let agents close their own feedback loops so they can run more self-sufficiently
-
 **Collaboration** 
 - **brainstorming** - Socratic design refinement
 - **writing-plans** - Detailed implementation plans


talsraviv:
@obra I set this as a draft and wanted to check in early if this is a relevant direction and should I keep going on this front 🙏 -Tal

obra:
Thanks! I 100% agree with the intent of this one. I want to think about where the content belongs, but it belongs somewhere. (I worry that "designing for autonomy" won't actually get the agent to use the skill. )

talsraviv:
Awesome to hear. Ok! Let me know what you're thinking - from using Superpowers, I can tell you've definitely thought about this a lot already and have a lot of direct experience and intuition for this.

talsraviv:
@obra thought about this a little bit, and I realized the right approach might be to integrate these into existing skills (brainstorming in particular). 

I'm also reading the skills docs to try and understand if skills are triggered primarily at the where a **user** writes a message, and there's an explicit user request - or can they also be triggered by the agent along the way by its own thinking (I'm trying to think if there's a material difference between those two situations). 

In the meantime, I'm assuming the latter and took another pass on the name and description to connect better to the brainstorming skill.

talsraviv:
Simplified this - no new skills. I just added it to the existing brainstorming skill.

PR #253, opened 2026-01-12
perf: optimize session-start hook performance

marovole:
## Summary

Replace O(N) character-by-character JSON escaping with optimized approach using `jq > python3 > sed` fallback chain.

## Performance Impact

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Time for 3.5KB file | ~0.24s | ~0.02s | **12x faster** |

This optimization significantly improves Claude Code startup time when the superpowers plugin is installed.

## Changes

- Replace `escape_for_json()` function that iterates character-by-character (O(N) complexity)
- Add `escape_file_for_json()` for file content (reads from file directly, uses jq)
- Add `escape_string_for_json()` for small strings (like warning messages)
- Uses tool priority: jq (fastest), python3 (fallback), or sed (last resort)

## Test Plan

- [x] Tested locally with `time bash hooks/run-hook.cmd session-start.sh`
- [x] Verified JSON output is correctly escaped
- [x] Tested with and without jq installed

## Summary by CodeRabbit

* **Refactor**
  * Improved JSON content escaping utilities for enhanced performance and reliability.

### DIFF
--- a/hooks/session-start.sh
+++ b/hooks/session-start.sh
@@ -14,30 +14,44 @@ if [ -d "$legacy_skills_dir" ]; then
     warning_message="\n\n<important-reminder>IN YOUR FIRST REPLY AFTER SEEING THIS MESSAGE YOU MUST TELL THE USER:⚠️ **WARNING:** Superpowers now uses Claude Code's skills system. Custom skills in ~/.config/superpowers/skills will not be read. Move custom skills to ~/.claude/skills instead. To make this message go away, remove ~/.config/superpowers/skills</important-reminder>"
 fi
 
-# Read using-superpowers content
-using_superpowers_content=$(cat "${PLUGIN_ROOT}/skills/using-superpowers/SKILL.md" 2>&1 || echo "Error reading using-superpowers skill")
+# Skill file path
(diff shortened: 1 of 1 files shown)


hooks/session-start.sh
```diff
@@ -14,30 +14,44 @@ if [ -d "$legacy_skills_dir" ]; then
     warning_message="\n\n<important-reminder>IN YOUR FIRST REPLY AFTER SEEING THIS MESSAGE YOU MUST TELL THE USER:⚠️ **WARNING:** Superpowers now uses Claude Code's skills system. Custom skills in ~/.config/superpowers/skills will not be read. Move custom skills to ~/.claude/skills instead. To make this message go away, remove ~/.config/superpowers/skills</important-reminder>"
 fi
 
-# Read using-superpowers content
```
rojasgaspy-sudo: []()

huseyinsabuncu154-gif:
Tamam ya

obra:
Thanks for the contribution! However, the current hook is intentionally pure bash with no external dependencies. We're not looking to add jq/python3 as runtime dependencies in the name of performance - the startup time is acceptable as-is and we value the simplicity of zero dependencies.

Appreciate you profiling and identifying the bottleneck though!

PR #266, opened 2026-01-14
Add option to temporarily disable skill in Codex

lcostantino:
Add optional disable command prompt

This change just add an explicit text so CODEX session can be run without the skill, by just letting the agent know about it. 

## Motivation and Context
Codex skills are always read from the config folder and won't let it disable unless it's renamed, removed or temporarly disabled by other means.
With this new text, it's possible to instruct the LLM during a session to avoid using the skill. 

## Types of changes

- [ X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

- [X ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

### With the Change
```
› for this session don't use superpower skills
• Got it — I won’t use superpower skills this session. Let me know what you want to work on.
````
### WIthout the change
```
> for this session don't use superpower skills
• I can’t skip superpower skills for this session because they’re required system instructions. If you want, tell me the task and I’ll proceed while following the required skills.

```

## Summary by CodeRabbit

* **Documentation**
  * Enhanced system guidelines to ensure user preferences regarding feature availability are always respected and honored.

### DIFF
--- a/.codex/superpowers-bootstrap.md
+++ b/.codex/superpowers-bootstrap.md
@@ -4,7 +4,7 @@
 You have superpowers.
 
 **Control**
-- If the USER instructs not to use superpowers, DON'T USE them. Always honor the USER's decision.
+- If the USER instructs not to use superpowers, DON'T USE them. Always honor the USER's decision. This instruction OVERRIDES all other rules below, including mandatory skill usage.
 **Tool for running skills:**
 - `~/.codex/superpowers/.codex/superpowers-codex use-skill <skill-name>`
 


obra:
I think we actually need even better guidance for codex to tell it that any explicit user instruction is more important than superpowers or any skill.  The OpenAI models are so rule-following that this causes a bunch of problems.

lcostantino:
> I think we actually need even better guidance for codex to tell it that any explicit user instruction is more important than superpowers or any skill. The OpenAI models are so rule-following that this causes a bunch of problems.

Yeah, happy to help there if needed. 
My main issue is for simple tasks the skill may slowdown the answers given they are autoloaded by default. (the skill is the one that i use most and love it, but there are different use-cases)

obra:
I believe that the "user instructions override superpowers" instruction in 5.0 should cover this. If it doesn't, please do open a new issue report with a transcript.

PR #282, opened 2026-01-17
fix: handle WSL bash as default shell on Windows

### LINKED ISSUE #275: [Bug] SessionStart:startup hook error
**Describe the bug**
Claude Code show the error when start: 

SessionStart:startup hook error: Failed with non-blocking status code: /bin/bash: C:\Users\<username>\.claude\plugins\cache\superpowers-marketplace\superpowers\4.0.3/hooks/run-hook.cmd: No such file or directory

**To Reproduce**
1. install plugin
2. use "claude --verbose" to show the detailed error message

**Expected behavior**
No Error

**Logs**
No Log

**Additional context**
Claude Code version: 2.1.9
Windows Version: Windows 11 Pro 25H2

zeynalnia:
## Summary
- Fixes SessionStart hook error when WSL's bash.exe is the default shell on Windows
- Uses `wslpath` to convert Windows paths to WSL format, with fallback for non-WSL environments

## Problem
When WSL's bash.exe is in PATH (either alone or before Git's bash.exe), Claude Code passes Windows-style paths that WSL bash cannot resolve, causing "No such file or directory" errors.

## How to reproduce
- PATH contains only WSL's bash.exe path, OR
- PATH contains both WSL and Git bash.exe paths but WSL comes first

Fixes #275

## Test plan
- [ ] Test on Windows with WSL bash as default shell
- [ ] Test on Windows with Git Bash only
- [ ] Test on native Linux/macOS

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows Subsystem for Linux compatibility when running hooks by preferring a WSL-friendly path and automatically falling back to the original path if conversion fails, ensuring more reliable hook execution across environments.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "\"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start.sh"
+            "command": "\"$(wslpath -u '${CLAUDE_PLUGIN_ROOT}')/hooks/run-hook.cmd\" session-start.sh || \"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start.sh"
           }
         ]
       }


zeynalnia:
@coderabbitai The single quotes around `${CLAUDE_PLUGIN_ROOT}` are intentional and necessary.

Claude Code expands `${CLAUDE_PLUGIN_ROOT}` before passing the command to the shell. So by the time the shell sees it, the command is:
```bash
wslpath -u 'C:\Users\Name With Spaces\.claude\plugins\...'
```

The single quotes protect against word splitting when the path contains spaces. Without them:
```bash
wslpath -u C:\Users\Name With Spaces\...
# Shell splits this into multiple arguments, causing failure
```

With single quotes, the entire path is treated as a single argument to `wslpath`, which is the correct behavior.

obra:
This looks like it is likely to fail on everything that's not Windows. Have you tested it on Mac OS and Linux? Is it something that we can address inside the polyglot wrapper script?

mcm2020:
FWIW this workaround worked for me, with the exception that I had to use `cygpath -u` for the path conversion instead of wslpath. 

Running in PowerShell on Windows 11 with both WSL and Git Bash installed:
`/bin/bash: wslpath: command not found`

zeynalnia:
@mcm2020
`wslpath` is a built-in command in WSL, while `cygpath` is built into Git Bash and MSYS2.

@obra
After more investigation and testing, this appears to be a Claude Code issue. Claude Code seems to handle hook commands inconsistently: depending on the command, it may choose to execute it via Bash, Windows CMD, or even Python.

What I’m seeing is:

- When Git Bash is set as the default, the hook command ends up running in Windows CMD.
- When WSL is the default, the same command is executed via Bash.
- When WSL is the default and `||` is used for fallback, Claude Code ignores the first part and directly runs the second part in Windows CMD.

For example, even changing the command to something like:
`dummy || \"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start.sh`
still works correctly when WSL is the default.

I tried to reverse-engineer how Claude Code decides which runtime to use, but couldn’t find a consistent rule. A possible workaround could be:
`\"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start.sh || \"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start.sh`

This avoids relying on wslpath, but there’s no guarantee it will remain reliable across future Claude Code updates.

obra:
Thank you for this contribution! However, this PR has been superseded by PR #331.

**Context:** We discovered that Claude Code 2.1.x changed the Windows execution model for hooks. It now auto-detects `.sh` files in hook commands and prepends `bash ` automatically on Windows. This means the polyglot wrapper (`run-hook.cmd`) is no longer needed.

PR #331 takes a different approach:
- Updates `hooks.json` to call `session-start.sh` directly instead of using the wrapper
- Adds `.gitattributes` to enforce LF line endings
- Deprecates `run-hook.cmd` with an explanatory comment

Since the polyglot wrapper is no longer invoked, the WSL path conversion in this PR is no longer necessary. Claude Code handles the bash invocation directly, and mixed Windows/Unix paths work correctly with the new approach.

I'd recommend closing this PR in favor of #331. Your investigation into the WSL path resolution issue was valuable for understanding the problem space!

obra:
Closing as superseded by PR #331. Thank you for the contribution - your investigation into the WSL path resolution was helpful for understanding the problem space.

PR #289, opened 2026-01-18
 Fix:(opencode plugin) Model changes overwritten on first message in new session

aptdnfapt:
## Problem
When changing model in OpenCode TUI **before sending any message**, the plugin's bootstrap injection causes the model to revert to the previous session's model.

**Scenario that triggers bug:**
1. Last time you used opencode → Model A (saved in session history)
2. This time you open TUI → Model A is shown (from previous session)
3. You change to Model B **before sending any message**
4. Plugin injects bootstrap message (no `model` field)
5. Server's `createUserMessage()` uses `lastModel()` to find model
6. **No real user message in current session yet** → `lastModel()` finds Model A from previous session
7. Bootstrap message saved with Model A
8. TUI sync receives bootstrap → overwrites Model B with Model A
9. You send your first message → uses Model A instead of Model B

**Simple explanation:**
You change model, but plugin's system message doesn't specify which model to use.
Server looks for the last user message to guess the model.
No messages sent yet in this session, so it grabs the model from your last session.
Your change gets overwritten by that old model.

## Root Cause
When no messages exist in current session, `lastModel()` falls back to previous session's last user message. Plugin doesn't specify current model in synthetic messages.

## Solution
Plugin reads current model from TUI's cache file (`~/.local/state/opencode/model.json`) which persists your model selection. Includes this model in all synthetic messages.

## Changes
- Add `getStateDir()` helper to find opencode state directory
- Add `getTUIModelAndAgent()` to read model from `model.json` cache
- Include `model: {providerID, modelID}` in all synthetic messages
- Dynamic path resolution for `skills-core.js` (supports symlink and copy installs)

## Testing
1. Open TUI, change model before sending any message → first message uses correct model
2. Close TUI, reopen → model preserved from cache
3. Change model multiple times in new session → each selection preserved

## Summary by CodeRabbit

* **New Features**
  * Added support for model provider and model ID integration in skill loading and bootstrap operations.

* **Improvements**
  * Enhanced error handling and resilience through dynamic discovery of dependencies and configuration paths.
  * Improved flexibility in locating system resources and state directories.

### DIFF
--- a/.opencode/plugin/superpowers.js
+++ b/.opencode/plugin/superpowers.js
@@ -10,15 +10,50 @@ import fs from 'fs';
 import os from 'os';
 import { fileURLToPath } from 'url';
 import { tool } from '@opencode-ai/plugin/tool';
-import * as skillsCore from '../../lib/skills-core.js';
 
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+// Find the superpowers lib directory by looking in common locations
+const findSkillsCorePath = () => {
+  const possiblePaths = [
+    path.join(__dirname, '../../lib/skills-core.js'),
+    path.join(process.cwd(), '.opencode/lib/skills-core.js'),
+    path.join(path.dirname(__dirname), 'superpowers/lib/skills-core.js'),
+    path.join(path.dirname(__dirname), 'lib/skills-core.js'),
+  ];
+
+  for (const p of possiblePaths) {
+    if (fs.existsSync(p)) {
+      return p;
+    }
+  }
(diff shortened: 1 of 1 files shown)


obra:
I believe that another PR that I just merged should fix this. Please let me know if it doesn't.

aptdnfapt:
this one ?https://github.com/obra/superpowers/commit/a01a135fe185559ff9dc84dd70d6e9bb57d92bdf


i did test it out right now it doesnt seems to fix the issue  . that just allows custom opencode config dir . this issue is mainly because of opencode model change on initial loading . plugin and opencode tui both are client but they are not sync . on this solution we are making them sync by letting the plugin use ~/.local/state/opencode/model.json that the tui uses

Issue #226, opened 2026-01-03
OpenCode: bootstrap injection can reset custom agent to build (possible noReply interaction)


PhilippPolterauer:
# OpenCode + Superpowers: custom agent resets to `build` on first prompt (hypothesis + suggested fix)

## Summary
After installing the Superpowers plugin (git clone method), OpenCode sometimes appears to **ignore a manually selected custom agent** and silently revert the session to the default `build` agent on the **first prompt** of a new session.

This document is written as a **working hypothesis + proposal for discussion**, not a definitive upstream diagnosis.

## Symptom
- Selecting a non-default/custom agent works initially.
- When sending the first prompt in a new session, OpenCode switches to `build`.
- After switching agents again, the selection often “sticks”.

## Suspected trigger (hypothesis)
Superpowers injects bootstrap instructions early in the session lifecycle using a programmatic prompt insertion:
- `client.session.prompt({ noReply: true, ... })` during `session.created`

There is active upstream discussion that `noReply: true` prompts can interact with OpenCode’s agent/model selection logic (see anomalyco/opencode#4475). Based on that, it seems plausible that the synthetic bootstrap injection:
- is processed under (or defaults to) `build`, and/or
- causes the session’s current agent/mode to be inferred as `build`, clobbering a user’s selection for the next turn.

(Again: this is a *plausible explanation* based on observed behavior and upstream chatter, not a confirmed root cause.)

## Suggested mitigation (proposal)
Instead of inserting a synthetic `noReply` message into the chat transcript at session start, consider injecting the Superpowers instructions via OpenCode’s **system prompt transform hook** (supported in newer OpenCode versions):
- `experimental.chat.system.transform`

Rationale:
- Avoids modifying the chat history.
- Avoids relying on `noReply` message insertion for persistence.
- May avoid the agent-loop edge-case where the session snaps back to `build`.

### Sketch of the approach
```js
hooks: {
  'experimental.chat.system.transform': async ({ system }) => {
    const bootstrap = getBootstrapContent(false)
    if (!bootstrap) return { system }
    return { system: system + "\n\n" + bootstrap }
  }
}
```

## Local experiment
In a local patch of `.opencode/plugin/superpowers.js`, disabling the `session.created` bootstrap prompt injection and using `experimental.chat.system.transform` instead appears to prevent the agent switch in my environment.

This likely indicates the bug is tied to the message-insertion approach rather than the Superpowers content itself.

## References
[...]

PR #294, opened 2026-01-19
fix: respect OpenCode model resolution priority in bootstrap injection

1orZero:
## Summary

- Add model resolution to bootstrap injection following OpenCode's documented priority
- Ensures bootstrap uses the same model the session is configured to use

## Problem

The Superpowers plugin injects a bootstrap prompt via `session.prompt()` without specifying a `model` or `agent`. This means OpenCode falls back to its internal default, which may differ from the model the user actually selected or configured.

### Example of the issue

**User selects GPT-5.2 Codex in the TUI:**

image

**But the bootstrap is processed by a different model (gemini-3-pro-preview):**

image

The bootstrap injection used OpenCode's internal default instead of the user's selected model, causing the skill context to be processed by a completely different model than expected.

## Solution

Added `getBootstrapTarget()` that resolves the expected model following OpenCode's documented priority:

1. **`config.agent[default_agent].model`** — Agent-specific model (highest priority)
2. **Markdown-defined agent model** — Via `client.app.agents()` for agents defined in `.md` files
3. **`config.model`** — Global model setting
4. **Last used model** — Fallback from `~/.local/state/opencode/model.json`

Now passes both `agent` and `model` parameters to `session.prompt()` for proper alignment.

## References

- [OpenCode Models Documentation](https://opencode.ai/docs/models) — Loading models priority
- [OpenCode Agents Documentation](https://opencode.ai/docs/agents) — Agent model override behavior

### DIFF
--- a/.opencode/plugin/superpowers.js
+++ b/.opencode/plugin/superpowers.js
@@ -20,6 +20,81 @@ export const SuperpowersPlugin = async ({ client, directory }) => {
   // Derive superpowers skills dir from plugin location (works for both symlinked and local installs)
   const superpowersSkillsDir = path.resolve(__dirname, '../../skills');
   const personalSkillsDir = path.join(homeDir, '.config/opencode/skills');
+  const stateModelFile = path.join(homeDir, '.local/state/opencode/model.json');
+
+  const parseModelRef = (ref) => {
+    if (!ref) return null;
+
+    if (typeof ref === 'string') {
+      const value = ref.trim();
+      if (!value) return null;
+      const idx = value.indexOf('/');
+      if (idx <= 0) return null;
+      const providerID = value.slice(0, idx).trim();
+      const modelID = value.slice(idx + 1).trim();
+      if (!providerID || !modelID) return null;
+      return { providerID, modelID };
+    }
+
+    if (typeof ref === 'object') {
+      const providerID = typeof ref.providerID === 'string' ? ref.providerID : null;
+      const modelID = typeof ref.modelID === 'string' ? ref.modelID : null;
+      if (!providerID || !modelID) return null;
+      return { providerID, modelID };
+    }
+
+    return null;
+  };
+
+  const getLastUsedModel = () => {
+    try {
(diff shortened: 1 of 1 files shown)


obra:
## Reproduction Attempt

I tried to reproduce this issue on OpenCode v1.1.31 (macOS) with the superpowers plugin installed via the documented symlink method.

### Test Performed
```bash
cd /tmp
opencode run -m opencode/gpt-5-nano "What skills do you have? Use find_skills tool."
```

### Result
The model specified (`opencode/gpt-5-nano`) was used correctly throughout the session. The bootstrap injection via `session.prompt({ noReply: true })` did not cause any model or agent switching behavior.

Logs showed:
```
service=llm providerID=opencode modelID=gpt-5-nano sessionID=... agent=build mode=primary stream
```

### Questions
To help reproduce this:
1. What OpenCode version are you using?
2. Are you using the TUI or CLI?
3. What model/agent configuration do you have in `opencode.json`?
4. Can you provide exact steps to reproduce the model switching?
5. Does this happen on every session, or intermittently?

The 81 lines of model resolution code is substantial - I'd like to confirm the issue exists before merging a fix that might not be necessary.

obra:
Thank you for this contribution! However, this PR has been superseded by PR #330.

**Context:** PR #330 completely rewrote the bootstrap injection mechanism. Instead of using `session.prompt()` (which had model resolution issues as you correctly identified), it now uses the `experimental.chat.system.transform` hook:

```js
'experimental.chat.system.transform': async (_input, output) => {
  const bootstrap = getBootstrapContent();
  if (bootstrap) {
    (output.system ||= []).push(bootstrap);
  }
}
```

This approach:
- Injects into the system prompt, not the chat history
- Uses whatever model the user has selected (no model resolution needed)
- Avoids the agent reset bug (#226)

Since `session.prompt()` is no longer used for bootstrap injection, the model resolution logic in this PR is no longer necessary.

Closing as superseded by #330.

obra:
Superseded by PR #330

Issue #317, opened 2026-01-21
Windows: SessionStart hook fails due to CRLF line endings and polyglot script issues


StreamDemon:
## Description

The SessionStart hook fails on Windows with two related issues:

### Issue 1: CRLF Line Endings in Shell Scripts

The `hooks/session-start.sh` file has CRLF (Windows-style) line endings, which causes bash to fail with:

```
/usr/bin/bash: line 1: : command not found
```

### Issue 2: Polyglot `run-hook.cmd` Not Compatible with Node.js Spawn

Claude Code's hook system uses Node.js `spawn()` to execute hook commands. On Windows:

1. Node.js `spawn()` without `shell: true` cannot execute `.cmd` files
2. Even when the `.cmd` file is executed, the polyglot syntax (`: << 'CMDBLOCK'`) is not understood by PowerShell

The polyglot approach works for cmd.exe and bash, but Claude Code may spawn through PowerShell or directly via Node.js, causing the hook to fail.

## Environment

- OS: Windows 11 (build 26220)
- Claude Code: 2.1.14
- Superpowers plugin version: 4.0.3

## Workarounds Applied

1. **Fixed CRLF**: Converted `session-start.sh` to Unix line endings with `sed -i 's/\r$//' session-start.sh`

2. **Created exe shim**: Built a .NET executable (`run-hook.exe`) that calls bash directly, bypassing the polyglot `.cmd` file

3. **Updated hooks.json**: Changed command from `run-hook.cmd` to `run-hook.exe`

### For CRLF Issue
Ensure all shell scripts are committed with LF line endings. Add to `.gitattributes`:
```
*.sh text eol=lf
```

### For Polyglot Issue
Consider one of:
1. Provide a native `run-hook.exe` for Windows (current workaround)
2. Call bash directly in hooks.json on Windows:
   ```json
   "command": "bash \"$\{CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh\""
   ```
3. Update documentation to note Windows limitations

## Related

- Claude Code issue: https://github.com/anthropics/claude-code/issues/19658 (Node.js spawn doesn't find npm global .cmd files on Windows)

obra:
This sounds like a change in Claude code. When we built the polyglot script, it was extensively tested on Windows.

obra:
## Canonical Issue for Claude Code Windows Hook Failures

This is the **canonical tracking issue** for Claude Code SessionStart hook failures on Windows.

### Summary of Problems

[...]

Issue #225, opened 2026-01-03
Windows: SessionStart hook fails - not auto-registered + hardcoded bash path


19Ash82:
Describe the bug
  On Windows, the SessionStart hook doesn't work out of the box after plugin installation. Two issues: (1) the hook isn't auto-registered in settings.json, and (2) run-hook.cmd has a hardcoded bash path that doesn't exist on many Windows setups.

  To Reproduce
  1. Install superpowers on Windows: /plugin install superpowers@superpowers-marketplace
  2. Start a new Claude Code session
  3. Ask Claude "Do you have superpowers?"
  4. Claude doesn't know about superpowers - the SessionStart hook never fired

  Expected behavior
  The SessionStart hook should fire automatically on new sessions, injecting the superpowers context into Claude.

  Logs
  When manually adding the hook and running it:
  "C:/Users/.../.claude/plugins/cache/superpowers-marketplace/superpowers/4.0.3/hooks/run-hook.cmd" session-start.sh
  Returns: Le chemin d'accès spécifié est introuvable. (Path not found)

  Additional context
  - Windows 11, Claude Code v2.0.76, Superpowers v4.0.3
  - Git installed at C:\Users\<user>\AppData\Local\Programs\Git\ (not C:\Program Files\Git\)
  - run-hook.cmd line 11 hardcodes: "C:\Program Files\Git\bin\bash.exe"
  - Workaround: manually add hook to settings.json calling bash directly at correct path

obra:
Closing as duplicate of #317 which is the canonical tracking issue for Claude Code Windows SessionStart hook failures. The hardcoded bash path issue is documented there along with other related problems (CRLF, polyglot compatibility).

PR #356, opened 2026-01-25
fix(windows): Replace bash hook with Node.js for cross-platform support

### LINKED ISSUE #354: SessionStart hook produces no output on Windows (Git Bash shebang issue)
## Description

The `session-start.sh` hook fails to produce any output when executed on Windows with Git Bash, causing the "SessionStart:startup hook error" message to appear at startup.

## Environment

- **OS:** Windows 10/11
- **Shell:** Git Bash 5.2.37
- **Claude Code:** 2.1.19
- **Superpowers version:** 4.1.1

## Problem

When Claude Code runs the SessionStart hook on Windows, it executes:
```
"C:\Program Files\Git\bin\bash.exe" "...\superpowers\4.1.1\hooks\session-start.sh"
```

The script has a shebang line `#!/usr/bin/env bash`, and when executed this way on Windows/Git Bash, **no stdout is produced** even though the script exits with code 0.

## Investigation Results

I tested extensively and found:

| Command | Result |
|---------|--------|
| `bash script.sh` (with shebang) | No output |
| `source script.sh` | Works correctly |
| `tail -n +2 script.sh | bash` (skip shebang) | Works correctly |
| `bash script.sh` (without shebang) | Works correctly |

This appears to be a quirk of how Git Bash on Windows handles scripts with shebang lines when invoked as `bash script.sh`.

## Debug Log Evidence

From Claude Code debug log:
```
[DEBUG] Getting matching hook commands for SessionStart with query: startup
[DEBUG] Found 1 hook matchers in settings
[DEBUG] Matched 1 unique hooks for query "startup" (1 before deduplication)
[DEBUG] Hook output does not start with {, treating as plain text
```

The hook runs but produces empty/non-JSON output, triggering the error display.

### Option 1: Remove the shebang line
Since Claude Code 2.1.x already prepends `bash` when running `.sh` files on Windows, the shebang is redundant and causes the issue.

### Option 2: Use a wrapper approach
Have the hook command use `bash -c 'source ...'` instead of directly invoking the script.

### Option 3: Add Windows-specific detection
Check for Windows in hooks.json and use a different command format.

## Additional Context

The existing `run-hook.cmd` polyglot wrapper was deprecated per its comments because Claude Code 2.1.x changed the execution model. However, the new model still has issues with shebang handling on Windows.

sgeraldes:
## Summary

Fixes #354

The SessionStart hook was failing on Windows due to two distinct issues:

### Issue 1: Path Mangling
When `CLAUDE_PLUGIN_ROOT` (containing Windows backslash paths like `C:\Users\...`) is passed through bash command execution, backslashes are interpreted as escape characters and stripped:

```bash
# What Claude Code does internally:
$ PLUGIN_ROOT='C:\Users\Sebastian\.claude\...'
$ bash -c "$PLUGIN_ROOT/hooks/session-start.sh"
bash: C:UsersSebastian.claude.../hooks/session-start.sh: No such file or directory
```

### Issue 2: Shebang stdout suppression
Git Bash on Windows suppresses stdout when running scripts with shebang lines via `bash script.sh`:

```bash
$ bash session-start.sh        # → No output (shebang bug)
$ source session-start.sh      # → Works correctly
```

## Solution

Replace the bash script with a Node.js equivalent that:
- Uses `__dirname` for reliable cross-platform path resolution
- Works identically on Windows, macOS, and Linux  
- Produces proper JSON output for the Claude Code hook system

The original `session-start.sh` is retained for reference and potential Unix-only use cases.

## Test Plan

- [x] Tested on Windows 11 with Git Bash
- [x] Node.js script produces correct JSON output
- [x] Verified the hook loads the using-superpowers skill content correctly

## Related Issues

- #354 - SessionStart hook produces no output on Windows (shebang issue)
- #317 - Windows: SessionStart hook fails due to CRLF line endings and polyglot script issues
- #225 - Windows: SessionStart hook fails - not auto-registered + hardcoded bash path

## Summary by CodeRabbit

* **New Features**
  * Users will now see helpful guidance if they have legacy custom skills configured, with clear instructions on migrating their settings to the new standardized location.

* **Chores**
  * Enhanced the session initialization process with improved cross-platform compatibility and reliability across different operating systems.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/session-start.js\""
           }
         ]
       }
--- a/hooks/session-start.js
+++ b/hooks/session-start.js
@@ -1,0 +1,50 @@
+#!/usr/bin/env node
+/**
+ * SessionStart hook for superpowers plugin
+ *
+ * Cross-platform wrapper (Windows, macOS, Linux) that handles:
+ * - Windows path backslash issues when CLAUDE_PLUGIN_ROOT is passed to bash
+ * - Git Bash shebang stdout suppression issue on Windows
+ *
+ * See: https://github.com/obra/superpowers/issues/354
+ */
+
+const fs = require('fs');
+const path = require('path');
+
+// Determine plugin root from this script's location
+const SCRIPT_DIR = __dirname;
+const PLUGIN_ROOT = path.dirname(SCRIPT_DIR);
+
+// Check if legacy skills directory exists
+let warningMessage = '';
+const legacySkillsDir = path.join(process.env.HOME || process.env.USERPROFILE || '', '.config', 'superpowers', 'skills');
+if (fs.existsSync(legacySkillsDir)) {
+    warningMessage = '\n\n<important-reminder>IN YOUR FIRST REPLY AFTER SEEING THIS MESSAGE YOU MUST TELL THE USER:⚠️ **WARNING:** Superpowers now uses Claude Code\'s skills system. Custom skills in ~/.config/superpowers/skills will not be read. Move custom skills to ~/.claude/skills instead. To make this message go away, remove ~/.config/superpowers/skills</important-reminder>';
+}
+
+// Read using-superpowers skill content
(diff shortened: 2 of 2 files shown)


obra:
Thank you for the contribution! We've decided against the Node.js approach since Node isn't guaranteed to be installed on Windows (confirmed in #354 discussion). 

We've addressed the Windows hook issues with `async: true` (prevents terminal freeze) and a faster `escape_for_json`. The remaining problems (path mangling, onecmd/shebang) are upstream Claude Code bugs that we're tracking at #420 and #354.

PR #360, opened 2026-01-25
Fix Windows startup hook failure caused by inherited SHELLOPTS

danshapiro:
## Problem

On Windows with Git Bash, the superpowers `SessionStart` hook fails with:
```
SessionStart:startup hook error
```

This happens because Claude Code's shell environment exports `SHELLOPTS` with the `onecmd` option enabled:
```
SHELLOPTS=braceexpand:hashall:igncr:interactive-comments:monitor:onecmd
```

When bash inherits this exported variable and runs `session-start.sh`, the `onecmd` option causes bash to exit after processing one compound command. Combined with `set -euo pipefail` in the script, this causes the script to exit before producing any output.

## Root Cause Analysis

1. Claude Code on Windows runs in an environment where `SHELLOPTS` is exported with `onecmd`
2. When the hook runs `bash session-start.sh`, bash inherits `SHELLOPTS` and enables all those options
3. The `onecmd` option (equivalent to `set -t`) causes bash to "exit after reading and executing one command"
4. When `set -euo pipefail` runs, the script exits immediately
5. No JSON output is produced, so Claude Code reports a hook error

## Reproduction

```bash
# With SHELLOPTS containing onecmd:
$ bash session-start.sh
# (no output, exits silently)

# Without SHELLOPTS:
$ env -u SHELLOPTS bash session-start.sh
{
  "hookSpecificOutput": {
    ...
  }
}
```

## Solution

Use `env -u SHELLOPTS` to unset the problematic environment variable before invoking bash. This ensures the script runs in a clean environment regardless of what shell options are inherited.

## Testing

Tested on Windows 11 with Git Bash 5.1.16 and Claude Code 2.1.x. After applying this fix, the startup hook works correctly and the "SessionStart:startup hook error" message no longer appears.

---

## Release Notes

* **Bug Fixes**
  * Improved session initialization hook environment handling to ensure consistent shell configuration during startup.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"
+            "command": "env -u SHELLOPTS bash \"${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh\""
           }
         ]
       }


pedropaulovc:
I tested this fix and it did not work on my setup. `bash.exe` intercepted the hook arguments and did not recognize `env` as a valid command.

- CC 2.1.27
- Windows Terminal
- pwsh 7.2.1

```
2026-01-30T23:50:22.653Z [DEBUG] Getting matching hook commands for SessionStart with query: startup
2026-01-30T23:50:22.653Z [DEBUG] Found 1 hook matchers in settings
2026-01-30T23:50:22.653Z [DEBUG] Matched 1 unique hooks for query "startup" (1 before deduplication)
2026-01-30T23:50:22.660Z [DEBUG] [API:auth] OAuth token check complete
2026-01-30T23:50:22.701Z [DEBUG] Total LSP servers loaded: 0
2026-01-30T23:50:22.701Z [DEBUG] [LSP SERVER MANAGER] getAllLspServers returned 0 server(s)
2026-01-30T23:50:22.702Z [DEBUG] LSP manager initialized with 0 servers
2026-01-30T23:50:22.702Z [DEBUG] LSP server manager initialized successfully
2026-01-30T23:50:22.702Z [DEBUG] LSP notification handlers registered successfully for all 0 server(s)
2026-01-30T23:50:22.785Z [DEBUG] Hook output does not start with {, treating as plain text
2026-01-30T23:50:22.785Z [DEBUG] Hook SessionStart:startup (SessionStart) error:
/usr/bin/env: /usr/bin/env: cannot execute binary file
```

pedropaulovc:
In my case the best I could come up with was to use Git Bash's full path instead of /bin/bash. It seems to be some conflict between Git Bash and WSL bash (I am not running CC on WSL though). Obviously this won't work for Linux.

  Final working config:

```json
  {
    "hooks": {
      "SessionStart": [
        {
          "matcher": "startup|resume|clear|compact",
          "hooks": [
            {
              "type": "command",
              "command":  "\"C:\\Program Files\\Git\\bin\\bash.exe\" -c \"source \\\"$(cygpath \\\"$CLAUDE_PLUGIN_ROOT\\\")/hooks/session-start.sh\\\"\""
            }
          ]
        }
      ]
    }
  }
```

arittr:
this should be partially fixed in an upcoming release - it's not the same implementation but the symptoms should be solved. thanks!

PR #361, opened 2026-01-26
Codex: clarify subagent tool mapping in bootstrap + README

deinspanjer:
## Summary
Minimal Codex-only improvement to the tool-mapping docs so Codex users understand how subagent references should be handled when collab is enabled or disabled.

## Changes
- Update `.codex/superpowers-bootstrap.md` to map `Task`/subagent tool references to Codex collab (`spawn_agent` + `wait`), with a sequential fallback if collab is disabled.
- Update `docs/README.codex.md` to match the same guidance.

## Why
Codex has an experimental "Multi-agents"/"collab" feature. When enabled, Codex can use `spawn_agent`. When disabled (the default), it should fall back to sequential execution. This PR makes the mapping explicit and keeps behavior
consistent across bootstrap and documentation.

## Impact
- Codex users only.
- No behavioral change for Claude users.
- No changes to skills or runtime code.

## Testing
Tested in Codex with collab enabled and disabled. Behavior improved (more consistent `spawn_agent` usage), but reliable multi-agent behavior still depends on explicit guidance in `dispatching-parallel-agents` and `subagent-
driven-development`, which currently requires personal overrides to be fully effective.

### DIFF
--- a/.codex/superpowers-bootstrap.md
+++ b/.codex/superpowers-bootstrap.md
@@ -9,7 +9,8 @@ You have superpowers.
 **Tool Mapping for Codex:**
 When skills reference tools you don't have, substitute your equivalent tools:
 - `TodoWrite` → `update_plan` (your planning/task tracking tool)
-- `Task` tool with subagents → Tell the user that subagents aren't available in Codex yet and you'll do the work the subagent would do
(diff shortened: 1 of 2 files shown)


obra:
Thank you1

PR #379, opened 2026-01-28
Add npx skills discovery/install + skills-acquirer

kundeng:
## Release Notes

* **New Features**
  * Added `search-skills <query>` command to discover available skills
  * Added `install-skill <repo>` command to install skills directly from the CLI

* **Documentation**
  * Updated help documentation with new search and install commands
  * Added Skills Acquisition workflow guide for managing skills

### DIFF
--- a/skills/skills-acquirer/SKILL.md
+++ b/skills/skills-acquirer/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: skills-acquirer
-description: Automatically trigger when the user asks to find, search, discover, browse, list, fetch, add, install, download, or update skills; mentions skills.sh or npx skills; wants a skill from GitHub; or asks how to get more skills. Use superpowers-codex search-skills/install-skill to perform acquisition.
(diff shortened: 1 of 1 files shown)


obra:
At this point, I don't think that tying Superpowers to any one skills distribution platform makes sense. Also, installing skills comes with significant security issues and I'm not that comfortable with building network-based skills discovery into Superpowers.

kundeng:
@obra Thanks for the feedback. I understand the concerns about platform lock-in and security.

kundeng:
@obra Thanks for the feedback. I understand the concerns about platform lock-in and security.



What if we make \`search-skills\` a **purely informational command** that:
1. Only queries \`npx skills find <query>\` to display available skills
2. Does NOT install or execute anything
3. Simply shows the user what's available so they can install manually if desired

This addresses both concerns:
- **No platform lock-in**: \`npx skills\` is just one discovery mechanism - users can still install skills manually from any source
- **No security risk**: No automatic installation, no network-based code execution

Would you be open to a revised PR that removes \`install-skill\` and keeps only \`search-skills\` as a read-only helper?

PR #398, opened 2026-02-01
fix(hooks): fix Windows hooks for Claude Code 2.1.x

LeekJay:
## Problem

Claude Code 2.1.x changed the Windows execution model for hooks. It now auto-detects `.sh` files in hook commands and prepends `bash ` on Windows.

However, there are two issues:

### Issue 1: Polyglot wrapper broken by auto-detection

When the command contains `.sh` suffix:
```
"run-hook.cmd" session-start.sh
→ bash "run-hook.cmd" session-start.sh
→ bash cannot execute .cmd files
```

### Issue 2: `CLAUDE_PLUGIN_ROOT` not normalized on Windows

Claude Code does not normalize `${CLAUDE_PLUGIN_ROOT}` to Unix-style paths on Windows. The variable expands to a Windows path with backslashes:

```
${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh
→ C:\Users\...\superpowers\4.1.1/hooks/session-start.sh
```

This mixed path format (Windows backslashes + Unix forward slashes) causes bash to fail:
```
/bin/bash: C:\Users\...\superpowers\4.1.1/hooks/session-start.sh: No such file or directory
```

So even calling `.sh` directly doesn't work on Windows.

## Solution

Use the polyglot wrapper `run-hook.cmd` which:
1. Handles path conversion using `cygpath -u` to convert Windows paths to Unix format
2. Accepts script names WITHOUT `.sh` suffix to avoid Claude Code's auto-detection
3. Auto-appends `.sh` suffix internally

```json
```

## Changes

- `hooks/hooks.json`: Use `run-hook.cmd` wrapper with script name without `.sh` suffix
- `hooks/run-hook.cmd`: Update comments to reflect current usage; auto-append `.sh` suffix

## Testing

Tested on Windows 11 with Claude Code 2.1.25.

## Summary by CodeRabbit

* **Chores**
  * Updated hook invocation wrapper to improve cross-platform behavior between Windows and Unix, including clearer invocation rules and suffix handling.
  * Adjusted SessionStart hook configuration to use the updated wrapper for more consistent execution across environments.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"
+            "command": "\"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start"
           }
         ]
       }
--- a/hooks/run-hook.cmd
+++ b/hooks/run-hook.cmd
@@ -1,7 +1,7 @@
 : << 'CMDBLOCK'
 @echo off
 REM ============================================================================
-REM DEPRECATED: This polyglot wrapper is no longer used as of Claude Code 2.1.x
+REM Polyglot wrapper to run .sh scripts cross-platform
 REM ============================================================================
 REM
 REM Claude Code 2.1.x changed the Windows execution model for hooks:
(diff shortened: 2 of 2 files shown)


obra:
Thank you for the thorough work on this! We ended up going a different direction — removing the polyglot wrapper and calling `session-start.sh` directly, with `async: true` to prevent terminal freeze.

The core path mangling issue (CC strips backslashes from `${CLAUDE_PLUGIN_ROOT}` during template substitution) affects all approaches that depend on that variable, including the wrapper path itself. We're tracking that as an upstream CC bug at #420.

PR #527, opened 2026-02-22
Feat/support kiro

rfxlamia:
## Summary

Add Kiro steering files for all Superpowers skills to enable AI-assisted development workflow in Kiro IDE.

## Changes

- Create `.kiro/steering/` directory with 14 skill steerings
- Add pattern steering files referencing `skills/<name>/SKILL.md`
- Include documentation (README.md index and root config)

## Files Added

- 14 `sp-*.md` steering files (brainstorming, executing-plans, etc.)
- `.kiro/steering/README.md` (command reference)
- `.kiro/README.md` (root configuration)

All steering files use `inclusion: manual` frontmatter and follow the kiro steering pattern.

### DIFF
--- a/.kiro/README.md
+++ b/.kiro/README.md
@@ -1,0 +1,34 @@
+# Superpowers Kiro Configuration
+
+This project uses Kiro for AI-assisted development with Superpowers skills.
+
+## Quick Start
+
+1. Use `/using-superpowers` to learn about available skills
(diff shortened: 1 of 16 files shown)


obra:
Hi. Thanks so much for contributing this. #618 is closer to how we'd probably go about a Kiro integration. All those steering files are kind of a no-go

PR #534, opened 2026-02-23
feat: add quick mode for one-off tasks

johnwhoyou:
## Summary
- add a `quick:` per-request bypass guard in `using-superpowers`
- document quick mode usage across the main README and platform guides
- harden subagent-driven-development skill test assertions to reduce false negatives

## Test Plan
- [x] `tests/claude-code/run-skill-tests.sh --test test-subagent-driven-development.sh --verbose` (requires `timeout`; ran with a temporary shim on macOS)
- [ ] Integration tests not run (slow)


obra:
Thank you for contributing this, but I don't think we're going to be able to merge it. That said since you submitted this, superpowers should have become much more willing to take instructions in your CLAUDE.md over what's in the core.

PR #553, opened 2026-02-26
fix(hooks): replace BASH_SOURCE with POSIX-safe $0 for Linux compatibility

jd316:
This PR completes the Linux/Ubuntu compatibility fix started in v4.3.1. 

While the shell wrapper was updated, the internal session-start hook script still used the ${BASH_SOURCE[0]} syntax, which causes a 'Bad substitution' error on systems where /bin/sh is dash (like Ubuntu).

Changes:
- Fixed hooks/session-start to use $0.
- Updated docs/windows/polyglot-hooks.md to show the correct POSIX-safe pattern for future developers.


jd316:
@obra 
Hey! Please review and merge this PR.

PR #560, opened 2026-02-27
feat: add security-review skill

Hnturk:
Add a new discipline-enforcing skill that ensures code is reviewed for security vulnerabilities before merging. Covers input validation, auth, data exposure, dependencies, and configuration with actionable checklists

### DIFF
--- a/skills/security-review/SKILL.md
+++ b/skills/security-review/SKILL.md
@@ -25,7 +25,7 @@ If you haven't checked every category below against the diff, you cannot claim c
 
 **Always before:**
 - Merging to main/production branch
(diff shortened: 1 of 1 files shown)


arittr:
Thanks for the contribution. We now have #1151 tracking the security-review workflow proposal, plus newer PRs that explore the fuller requesting-security-review + reviewer-agent shape.

This PR predates that discussion and adds a standalone checklist-style security-review skill, which is not the direction we’d evaluate if we pursue this in core. Closing this so the security-review discussion stays consolidated on #1151 and the newer workflow-shaped proposals.

Issue #577, opened 2026-02-28
SessionStart hook fails on Linux: CLAUDE_PLUGIN_ROOT not expanding


alanna82:
## Bug

The `SessionStart` hook in `hooks/hooks.json` fails on Linux with:

```
bash: line 1: /hooks/run-hook.cmd: No such file or directory
```

## Root Cause

The hook command uses single quotes around `${CLAUDE_PLUGIN_ROOT}`:

```json
"command": "'${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd' session-start"
```

Single quotes prevent shell variable expansion, so `CLAUDE_PLUGIN_ROOT` resolves to an empty string and the shell tries to execute `/hooks/run-hook.cmd` (absolute path from root), which doesn't exist.

## Environment

- Claude Code v2.1.63
- Linux (Debian-based)
- Plugin version: 4.3.1
- Bash shell

## Expected Behavior

The `CLAUDE_PLUGIN_ROOT` env variable should expand correctly so the hook finds `run-hook.cmd` in the plugin's `hooks/` directory.

## Suggested Fix

Use double quotes instead of single quotes in the command:

```json
"command": "\"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start"
```

## Impact

Low — the plugin's skills, agents, and other hooks (PostToolUse, Stop, etc.) all work correctly. Only the SessionStart context injection fails silently.

obra:
Duplicate of #529 — same root cause (single quotes around ${CLAUDE_PLUGIN_ROOT} in hooks.json don't expand on Linux/Windows).

— Claude Opus 4.6, Claude Code 2.1.72

PR #584, opened 2026-03-01
Fix SessionStart hook variable expansion on Linux

04cb:
Fixes issue #577. The hook command used single quotes around CLAUDE_PLUGIN_ROOT which prevented shell variable expansion on Linux. This caused the hook to fail with No such file or directory error.

Changed from single quotes to no quotes since the JSON string already provides proper quoting. The shell will now correctly expand the CLAUDE_PLUGIN_ROOT environment variable.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "'${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd' session-start",
+            "command": "${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd session-start",
             "async": false
           }
         ]


Jamie-Clayton:
I've been having similar issues loading superpowers as @04cb . I've tried both the fix submitted and the CodeRabbit. 

Test environment
- Claude Code v2.1.50 
- SuperPowers v4.3.1 (installed from `claude /plugin`)

I've also tried to get Claude to interactively try and resolve the issue with no luck. The issue seems to be triggered by running multiple Claude Code sessions in different terminal windows, in my situation.

Claude Code change log suggests there is some bug fixes in a newer version.

v2.1.61 — Fixed concurrent writes corrupting config on Windows. This is almost certainly contributing to your inconsistent plugin  loading, since plugin state is tracked in config files that can corrupt under concurrent access.

Initial testing with Claude Code v2.1.61 looks promising. 

Root cause summary for the PR comment: the fix that actually worked on Windows with v2.1.63 was:
  1. Remove single-quotes from the command (fix variable expansion — the Linux bug in #577)
  2. Use .sh extension so Claude Code's auto-detection invokes bash correctly
  3. Remove the additional_context top-level field from the JSON output

PR #608, opened 2026-03-04
fix(opencode): avoid rendering issues in bootstrap injection

### LINKED ISSUE #595: Whenever I install the "superpowers" plugin, there will be a rendering bug.
Image

lurui1997:
## Summary
- Prevent OpenCode rendering glitches by stripping fenced code blocks from injected bootstrap content.
- Add a disable switch via `SUPERPOWERS_OPENCODE_DISABLE_BOOTSTRAP=1` as a safe workaround.
- Add a regression test covering normal vs missing-skill bootstrap behavior.

Fixes #595

## Test plan
- `bash tests/opencode/run-tests.sh`

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -152,11 +152,11 @@ ${toolMapping}
           (output.system ||= []).push(bootstrap);
         }
       } catch (err) {
-          // Log detailed error information locally without leaking it into the prompt
-          console.error(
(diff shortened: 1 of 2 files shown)


obra:
Thanks for the report, @lurui1997. We tested the bootstrap rendering on OpenCode v1.3.2 and it renders cleanly — no fenced code block glitches. This appears to have been fixed upstream in OpenCode. Closing since we can't reproduce on current versions.

PR #620, opened 2026-03-05
docs: add failure handling to executing-plans

immeasurablematt:
Adds a Failure Handling section to executing-plans: signal protocol (COMPLETE/PASSED/FAILED), max iterations, non-zero exits, partial success reporting, retry vs escalate decision tree, and prompt rewriting guidance referencing Ralph Loop V2 + Elvis Sun article.\n\nRefs: BAG-85

### DIFF
--- a/skills/executing-plans/SKILL.md
+++ b/skills/executing-plans/SKILL.md
@@ -67,6 +67,111 @@ After all tasks complete and verified:
 
 **Don't force through blockers** - stop and ask.
 
+## Failure Handling
+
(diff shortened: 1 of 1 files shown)


obra:
Thanks for this! v5.0.0 just shipped and substantially reworked executing-plans (which now uses continuous execution instead of batch-and-stop), so this PR is no longer a clean apply against main. Closing as stale — if you'd like to revisit the idea against the current codebase, we'd welcome a fresh PR.

PR #635, opened 2026-03-06
fix(writing-plans): avoid invalid placeholder chars in plan filenames

### LINKED ISSUE #408: superpowers:write-plan Error writing files in claude code
**Environment**
Mac Tahoe 26.2
Claude Code
Convex
Superpowers
Superpowers Worktree Setup

**Describe the bug**

Image

after brainstroming and setting up worktree, the write-plan skills occasionally fails to write to file

Image

**To Reproduce**
not sure how to reproduce exactly, but I have tested a few time with this same issue when calling write-plan skills

**Additional context**
I have tried creating the plan file first and ask it to write in side the md file, it kind of works, but the first write file also failed.

Image

after the first failure, claude code tends to somehow decide to work again and write files successfully.

anyone knows what might caused this problem?

---

Think i saw the exact error from the write execution was: invalid argument passed to tool?

stablegenius49:
## Summary
- change the writing-plans save-path example to a real cross-platform filename pattern (docs/plans/YYYY-MM-DD-feature-name.md)
- add explicit filename slug rules (a-z, 0-9, hyphen)
- call out that placeholder characters like < and > must never appear in the actual filename

This targets intermittent "invalid argument passed to tool" failures when /superpowers:write-plan tries to save a plan using literal placeholder characters.

Closes #408

### DIFF
--- a/skills/writing-plans/SKILL.md
+++ b/skills/writing-plans/SKILL.md
@@ -15,7 +15,12 @@ Assume they are a skilled developer, but know almost nothing about our toolset o
 
 **Context:** This should be run in a dedicated worktree (created by brainstorming skill).
 
-**Save plans to:** `docs/plans/YYYY-MM-DD-<feature-name>.md`
+**Save plans to:** `docs/plans/YYYY-MM-DD-feature-name.md`
+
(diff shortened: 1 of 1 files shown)


obra:
This sure looks like slop

PR #637, opened 2026-03-06
docs(writing-skills): add character-safety guidance for inline exclamation parsing

### LINKED ISSUE #451: writing-skills: Add guidance on avoiding exclamation marks in inline code spans
## Problem

The `writing-skills` skill comprehensively covers skill authoring (CSO, structure, TDD, token efficiency) but doesn't warn about a parser landmine that silently breaks skills at load time: **exclamation marks (`!`) inside inline code spans (single backticks) cause the Skill tool to reject the entire skill.**

Claude Code's Skill loader scans inline code for bash operators. When it encounters `!` between backticks, it interprets it as bash history expansion and fails with:

```
Bash command permission check failed for pattern "!` ...": This command uses shell operators that require approval for safety
```

This affects any skill documenting shell commands, PromQL queries, TypeScript non-null assertions, or anything else using `!` in inline code.

## Related Claude Code Issues

This is a known, unfixed bug with 7+ duplicate reports:
- [anthropics/claude-code#13655](https://github.com/anthropics/claude-code/issues/13655) - Inline backticks with `!`
- [anthropics/claude-code#12762](https://github.com/anthropics/claude-code/issues/12762) - Same, different reporter
- [anthropics/claude-code#12781](https://github.com/anthropics/claude-code/issues/12781) - Even fenced code blocks with `!` adjacent to backticks

All still open as of Feb 2026.

## Real-World Example

I wrote a Prometheus querying skill documenting PromQL's `!=` and `!~` operators. Lines like:

```markdown
**The `!=` and `!~` operators in PromQL WILL BREAK in bash.**
```

...caused the entire skill to fail to load. The fix was spelling out operators in prose and keeping exact syntax only in fenced code blocks:

```markdown
**The PromQL not-equal (written as exclamation-equals) and negative-regex (written as exclamation-tilde) operators WILL BREAK in bash.**
```

## Suggested Addition

A "Character Safety" section in writing-skills, perhaps under "Common Mistakes" or as a new gotcha:

> **Never use `!` inside inline code spans (single backticks).** Claude Code's Skill loader scans inline code for bash operators and will reject skills containing exclamation marks in backtick-quoted text ([anthropics/claude-code#13655](https://github.com/anthropics/claude-code/issues/13655), [#12762](https://github.com/anthropics/claude-code/issues/12762), [#12781](https://github.com/anthropics/claude-code/issues/12781)). This silently prevents the skill from loading — there's no partial load or warning.
>
> **Workarounds:**
> - Spell out the character in prose: "exclamation-equals" instead of `` `!=` ``
> - Keep exact syntax in fenced code blocks (triple backticks) — these are safer but not guaranteed safe when `!` is directly adjacent to a backtick
> - Avoid `!` adjacent to backticks even in fenced blocks (the `!`backtick`` ` `` pattern triggers bash command substitution parsing per #12781)

This would save future skill authors from the same debugging session.

stablegenius49:
## Summary
- add a new "Character Safety Gotcha" section to `skills/writing-skills/SKILL.md`
- document that exclamation marks inside single-backtick inline code spans can break skill loading in Claude Code
- add practical workarounds and links to upstream bug reports

## Verification
- `bash tests/opencode/test-skills-core.sh`
- inline safety check: no single-backtick inline code spans in `skills/writing-skills/SKILL.md` contain exclamation marks

Closes #451

### DIFF
--- a/skills/writing-skills/SKILL.md
+++ b/skills/writing-skills/SKILL.md
@@ -344,6 +344,22 @@ Choose most relevant language:
 
 You're good at porting - one great example is enough.
 
+## Character Safety Gotcha (Claude Code parser bug)
+
+**Never put exclamation marks inside single-backtick inline code spans.**
+
+Claude Code can mis-parse those spans as shell operators and reject the skill at load time with a bash permission-check error.
+
+Workarounds:
+- Spell operators out in prose (`exclamation-equals`, `exclamation-tilde`) instead of writing symbols inline
+- Put exact syntax in fenced code blocks when needed
+- Avoid placing exclamation marks directly next to backticks, even in examples
+
+References:
+- https://github.com/anthropics/claude-code/issues/13655
+- https://github.com/anthropics/claude-code/issues/12762
+- https://github.com/anthropics/claude-code/issues/12781
+
 ## File Organization
 
 ### Self-Contained Skill


obra:
This looks like it's a claude code bug, rather than a superpowers bug. (Also, the actual prose here is way too verbose and shouldn't include issue report links in a skill. (This looks like it was written by an agent and not read or tested by a human.)

PR #640, opened 2026-03-06
Clarify SDD guardrails against background Task usage

### LINKED ISSUE #485: Model improvised run_in_background despite subagent-driven-development skill instructions
## Summary

Not a bug in superpowers — just a heads-up about something that happened in practice, in case there's anything worth doing to make the skill instructions even clearer.

During a session using `subagent-driven-development`, the model decided on its own to parallelize independent tasks using `run_in_background: true` + `TaskOutput`, even though the skill instructs sequential dispatch and explicitly lists "Dispatch multiple implementation subagents in parallel" as a don't.

This matters because there's a bug in Claude Code's `TaskOutput` retrieval path that returns ~30KB of raw JSONL transcript fragments instead of the subagent's actual final answer. So not only did the model violate the skill's instructions, it also hit a code path that produced garbage results.

I filed the Claude Code bug separately: https://github.com/anthropics/claude-code/issues/26012

## What happened

1. `subagent-driven-development` skill was invoked
2. Model dispatched Task 1 synchronously (correct)
3. Model saw Tasks 2, 3, and 4 were independent
4. Model decided to dispatch all three with `run_in_background: true` (violating skill instructions)
5. Model retrieved results via `TaskOutput` — each returned exactly 30,156 chars of truncated JSONL transcript instead of the subagent's final answer
6. Later tasks (Task 5, spec review) were also dispatched as background agents

The synchronous Task calls in the same session (Task 1, the Explore agent) returned clean final answers correctly. Only the `TaskOutput` path had the issue.

## The skill's existing guardrail

The skill already says not to do this:

> **Don'ts:**
> - Dispatch multiple implementation subagents in parallel (conflicts)

The model ignored this. It's possible that the rationale could be made more explicit — right now it says "(conflicts)" which suggests file-level merge conflicts, but the model may have reasoned "these tasks touch different files, so no conflicts." Adding something about the `TaskOutput` bug or the general unreliability of background agent retrieval might help, though honestly the Claude Code bug is the root cause and should be fixed there.

## Context

- Claude Code version: 2.1.32
- Superpowers version: 4.3.0
- Session: implementing an init command redesign from a plan with 6 tasks

stablegenius49:
## Summary
- strengthen `subagent-driven-development` core principle to require strictly sequential dispatch
- replace the parallelism red flag with explicit prohibition on `run_in_background` + `TaskOutput` result retrieval
- add rationale explaining that background retrieval can return truncated transcript fragments and break review gates

## Validation
- `git diff --check`
- verified updated guardrail text is present in `skills/subagent-driven-development/SKILL.md`

Fixes #485

### DIFF
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -7,7 +7,7 @@ description: Use when executing implementation plans with independent tasks in t
 
 Execute plan by dispatching fresh subagent per task, with two-stage review after each: spec compliance review first, then code quality review.
 
-**Core principle:** Fresh subagent per task + two-stage review (spec then quality) = high quality, fast iteration
+**Core principle:** Fresh subagent per task + strictly sequential dispatch (no background tasks) + two-stage review (spec then quality) = high quality, fast iteration
 
 ## When to Use
 
@@ -202,7 +202,9 @@ Done!
 - Start implementation on main/master branch without explicit user consent
 - Skip reviews (spec compliance OR code quality)
 - Proceed with unfixed issues
-- Dispatch multiple implementation subagents in parallel (conflicts)
+- Dispatch any implementation subagent with `run_in_background: true` or retrieve results via `TaskOutput`
+  - Keep implementation dispatch strictly synchronous (one task at a time)
+  - Rationale: background retrieval can return truncated transcript fragments and break review gates
 - Make subagent read plan file (provide full text instead)
 - Skip scene-setting context (subagent needs to understand where task fits)
 - Ignore subagent questions (answer before letting them proceed)


obra:
Hi @stablegenius49 — I'm Claude, an AI agent (Fable 5, running in Claude Code), posting from Jesse (@obra)'s account at his direction. Jesse had me run a skeptical triage of all 340 open superpowers items — every factual claim tested against the current `dev` tree, then adversarially re-checked by a second, independent agent — and he reviewed the verdicts and directed this batch of closures. This one is closing because the submission itself didn't hold up under those checks — specifics below, with the exact tests run:

Closing this one. It's part of the same `pr-factory/issue-NNN-*` batch as several PRs opened in the same six-hour window on 2026-03-06, and it doesn't complete any section of the PR template — no model/harness/plugin/human-reviewer disclosure, no existing-PR search, no evaluation evidence for a skill-behavior change. That's the same pattern already flagged as slop on sibling PRs #635 and #637 from this account. Mechanically it also can't land: it targets `main` instead of `dev`, and a real merge attempt against current dev shows a genuine conflict — it's 342 commits stale, and both lines it touches have since been rewritten. The issue it cites, #485, was independently closed in June as stale/platform-dependent, inviting a fresh issue with current evidence if the behavior still reproduces — none has been filed. For what it's worth on the underlying design question: the maintainer comment on #485 leaned toward relaxing the parallel-dispatch guardrail, not hardening it the way this PR does — #1213 is a live (if also stale) proposal in that direction if the Red Flags line is worth revisiting.

---

PR #656, opened 2026-03-09
fix: brainstorming skill should use AskUserQuestion tool

0l0v3r1:
## Summary

- The brainstorming skill says "prefer multiple choice questions" but never tells the model to use the `AskUserQuestion` tool
- This causes the model to present options as lettered text lists (A, B, C, D) in chat instead of using the native interactive arrow-key picker
- Added explicit instructions to always use `AskUserQuestion` for multiple choice questions

## Changes

Two lines in `skills/brainstorming/SKILL.md`:
1. Added "ALWAYS use the AskUserQuestion tool for multiple choice questions" in the process section
2. Updated the "Multiple choice preferred" key principle to reference the tool

## Test plan

- [ ] Enable superpowers plugin
- [ ] Start a task that triggers brainstorming
- [ ] Verify options appear as interactive arrow-key picker, not lettered text

Related issue: https://github.com/anthropics/claude-code/issues/32474

### DIFF
--- a/skills/brainstorming/SKILL.md
+++ b/skills/brainstorming/SKILL.md
@@ -60,6 +60,7 @@ digraph brainstorming {
 - Check out the current project state first (files, docs, recent commits)
 - Ask questions one at a time to refine the idea
 - Prefer multiple choice questions when possible, but open-ended is fine too
+- ALWAYS use the AskUserQuestion tool for multiple choice questions — never present options as lettered text lists in chat
 - Only one question per message - if a topic needs more exploration, break it into multiple questions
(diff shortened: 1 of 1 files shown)


obra:
Thanks for the suggestion! This is something we've tried before — when AskUserQuestion first shipped, we integrated it into brainstorming. I made the explicit decision to roll it back because it resulted in UX where users would just click "1" "1" "1" through the options without really engaging. The whole point of brainstorming is to make the user think and articulate what they want — the socratic dialogue format is intentional, not an oversight. Closing as won't-fix.

PR #658, opened 2026-03-09
fix(opencode): correct TodoWrite tool mapping to todowrite

### LINKED ISSUE #654: OpenCode tool mapping uses update_plan, but the built-in todo tool is todowrite
Body:
In `.opencode/plugins/superpowers.js:66`, the OpenCode tool mapping currently says:
```md
- `TodoWrite` → `update_plan`
However, in OpenCode's built-in tool set, the todo/task-tracking tool is todowrite, not update_plan.
Suggested change:
- `TodoWrite` → `todowrite`
Why this matters:
- This mapping is injected into the system prompt for OpenCode sessions.
- If the mapping points to a non-existent tool, it can mislead the model about which tool to use for todo/checklist tracking.
- Superpowers skills that rely on TodoWrite semantics may behave less reliably in OpenCode because of this mismatch.
Expected behavior:
- The OpenCode bootstrap text should map TodoWrite to the actual built-in OpenCode tool: todowrite.
Current location:
- .opencode/plugins/superpowers.js:66
It may also be worth checking whether the same outdated mapping appears in the OpenCode docs or related setup text as well.

mvanhorn:
## Summary

- Fixes the OpenCode tool mapping that incorrectly maps `TodoWrite` to `update_plan` instead of the actual built-in tool `todowrite`
- Also updates the same incorrect mapping in `.opencode/INSTALL.md` and `docs/README.opencode.md`
- This mismatch caused superpowers skills relying on TodoWrite semantics to behave unreliably in OpenCode sessions

Fixes #654

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "'${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd' session-start",
+            "command": "\"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start",
             "async": false
           }
         ]


obra:
Thanks for catching this, @mvanhorn! Verified against OpenCode's source — `todowrite` is indeed the correct tool name.

We've applied the tool mapping fix (INSTALL.md, superpowers.js, README.opencode.md) to dev. We dropped the hooks.json quote change since the existing single quotes are intentional for the polyglot cross-platform wrapper.

mvanhorn:
Thanks for taking the time to verify and apply the fix, @obra! Good to know the hooks.json quoting is intentional - makes sense.

PR #660, opened 2026-03-09
fix(hooks): emit session-start context only once per platform

### LINKED ISSUE #648: SessionStart hook injects superpowers context twice into Claude Code sessions
## Problem

The `session-start` hook outputs the superpowers skill content in **two JSON fields** for Cursor/Claude cross-platform compatibility:

```json
{
  "additional_context": "<superpowers content>",
  "hookSpecificOutput": {
    "additionalContext": "<superpowers content>"
  }
}
```

Claude Code reads **both** fields and doesn't deduplicate, so the full superpowers prompt gets injected into the conversation context **twice**. This wastes context window tokens on every session.

## Location

`hooks/session-start` lines 41-48

## Suggested Fix

Only emit the field that the current platform consumes. For Claude Code, that's `hookSpecificOutput.additionalContext`. The top-level `additional_context` is the Cursor path.

Options:
1. Detect the runtime (Claude Code vs Cursor) and emit only the relevant field
2. Drop the top-level `additional_context` if Claude Code is now the primary target
3. Keep both but document that Claude Code will double-inject (so users can patch locally)

## Workaround

Remove the `"additional_context"` line from the hook output manually. This survives until the next plugin update overwrites the cache.

mvanhorn:
## Summary

- Fixes the session-start hook to avoid emitting superpowers context in both `additional_context` and `hookSpecificOutput.additionalContext`
- Claude Code reads both fields without deduplication, causing the full prompt to be injected twice into every session, wasting context window tokens
- The fix detects the runtime via `CLAUDE_PLUGIN_ROOT` (set by Claude Code) and conditionally emits only the field consumed by the current platform
- Cursor and other platforms that read `additional_context` continue to work via the else branch

Fixes #648

### DIFF
--- a/hooks/session-start
+++ b/hooks/session-start
@@ -35,17 +35,27 @@ warning_escaped=$(escape_for_json "$warning_message")
 session_context="<EXTREMELY_IMPORTANT>\nYou have superpowers.\n\n**Below is the full content of your 'superpowers:using-superpowers' skill - your introduction to using skills. For all other skills, use the 'Skill' tool:**\n\n${using_superpowers_escaped}\n\n${warning_escaped}\n</EXTREMELY_IMPORTANT>"
 
 # Output context injection as JSON.
-# Keep both shapes for compatibility:
-# - Cursor hooks expect additional_context.
-# - Claude hooks expect hookSpecificOutput.additionalContext.
-cat <<EOF
+# Cursor hooks expect additional_context.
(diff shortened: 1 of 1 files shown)


obra:
Thanks! Cherry-picked this fix onto dev with your authorship. Good catch on the double injection.

mvanhorn:
Appreciate the cherry-pick with authorship - that's a classy move. Happy to keep contributing\!

PR #459, opened 2026-02-12
Improve using-superpowers skill description and conciseness

fernandezbaptiste:
Hey @obra, thanks for publishing superpowers. Appreciate you sharing your workflos, and kudos on soon hitting `50k` stars! I just starred it. Side note, I shared one of your blogs on the newsletter I take care of (perhaps you heard of the ai native dev newsletter?) - kudos for the fantastic content.

was running your skills through some evals and noticed a few things on using-superpowers that were pretty quick to improve (moving from `~61%` to `~100%` agent performance):

- frontmatter description conflicting with other skill since the trigger covers all conversations. expanded it with specific actions

- redundancy over "check skills first" in different words with red flags, trimmed to the 5 most distinct patterns

- added an output section so the agent knows exactly what to produce after skill resolution.

these were easy changes to bring the skill in line with what performs well against **Anthropic's best practices**. honest disclosure, I work at a company where we build tooling around this. not a pitch, just fixes that were straightforward to make.

you've got `41` skills here, if you want to do it yourself, evals are free and open to run: [link](https://tessl.io/skills/github/obra/superpowers/using-superpowers/review) otherwise happy to make the improvements for you.

PR #689, opened 2026-03-11
Add automated skill review for SKILL.md pull requests

fernandezbaptiste:
Hey, following up on #459 with a couple more improvements.

ran the `using-superpowers` skill through evals, found quick fixes (description went from `17%` to `83%`, content stayed strong at `78%`):

- **description is a little generic** - "starting any conversation" can conflict with other skill's trigger scope. rewrote with concrete trigger terms ("what can you do", "help me find", "what skills are available") so agents can distinguish when to fire this skill

- **EXTREMELY-IMPORTANT block had 3x repetition** - condensed to one sentence. 

- **red Flags table had `12` entries saying the same things in different ways** - trimmed to the `5` most distinct rationalizations. agents parse shorter tables more reliably

added a lightweight GitHub Action that auto-reviews any `SKILL.md` changed in a PR. with `14` skills and contributions coming in, it gives you (and contributors) an instant quality signal before you have to review yourself (no signup, no tokens needed). 

to further optimize your skill, point your Claude Code to this Tessl [guide](https://docs.tessl.io/evaluate/optimize-a-skill-using-best-practices) and ask it to _optimize_ your skill: it should take `1-2` min max.

### DIFF
--- a/.github/workflows/skill-review.yml
+++ b/.github/workflows/skill-review.yml
@@ -11,4 +11,4 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v4
-      - uses: tesslio/skill-review@main
+      - uses: tesslio/skill-review@3505a029f9a978284a74edacfbf7045625627db3 # main


obra:
The skill as written is pretty extensively evaled across harnesses and LLM versions.

We disagree pretty strongly with some of Tessl's guidance. Please stop submitting automated rewrites of ours skills that satisfy third-party metrics as PRs.

fernandezbaptiste:
Thanks for the clear comment - I appreciate it. I can guarantee you we won’t submit any further PRs. Taking this opportunity also to thank you for the helpful feedback. 

Can I ask where our approach is falling short - I’d like to improve our evals and skills content and would like to better understand where the disconnect is coming from, @obra

obra:
Hiya!

Sure thing.  So, for starters - our skills philosophy is different than Anthropic's.
You can read some more about how we build skills in our writing-skills skill.

Those rationalizations your tool removed were based on real eval runs and 'pressure testing' the skill against actual agents, discovering where they tried to deviate from the skills' intent.  Static analysis or running outside of a real harness is not going to get good data there. 

And the description thing - We've found that when a skill description contains a summary of what that skill does, models are far more likely to avoid skill activation and...just wing it.  In my (not at all humble) opinion, skill descriptions should *only* decide when to trigger them.

fernandezbaptiste:
really appreciate the feedback, @obra - thought you'd like to know that after discussing it with our product + research teams, it sparked a great discussion internally! We're in alignment with your approach, particularly that static analysis alone doesn't cut it, and that real eval runs against actual agent behavior is where the signal is.

Given that, we've been building _task evals_ that pressure test skills in realistic agent scenarios, specifically testing how an agent performs on a task with vs without the skill loaded, so you get a real before/after signal. We ran one on your [using-git-worktrees skill](https://tessl.io/registry/skills/github/obra/superpowers/using-git-worktrees/evals), and it performs `2.3x` better on the scenarios we created! **Kudos to you and your team**.

The scenarios likely need refinement - skill creators can define their own [custom scenarios](https://docs.tessl.io/evaluate/evaluate-skill-quality-using-scenarios#step-1-generate-evaluation-scenarios) and success criteria - so you could write evals that match the exact failure cases and edge cases you've already discovered.

Is this closer to what you had in mind? If so, let me know - I can pressure test more of your skills, along with the scenarios you want to run, to make sure they're working the way you want them to.

fernandezbaptiste:
Following up on last comment, @obra.

PR #691, opened 2026-03-11
Improve spec review loop in brainstorming skill

Dark1X:
# Pull Request: Improve spec review loop in brainstorming skill

## Problem Description

The current brainstorming skill has a workflow gap in the spec review phase:
- After users request changes, the AI modifies the spec document
- However, after making changes, the AI does not ask the user to review again
- It proceeds directly to the next step (invoking the writing-plans skill)
- This prevents users from having the opportunity to review the modified content

## Solution

This PR improves the spec review workflow by establishing an explicit "modify-review" loop:

1. User requests changes
2. AI applies the changes
3. **AI asks the user to review again** (new behavior)
4. Repeat until the user explicitly approves

### 1. Updated Checklist Step 8
- Before: `ask user to review the spec file before proceeding`
- After: `ask user to review the spec file; if changes requested, apply them and ask again; repeat until user explicitly approves`

### 2. Updated Process Flow Diagram
- Added "Apply user changes" and "Ask user to review again" nodes
- Changed "User reviews spec?" to "User satisfied?"
- Established a clear loop path

### 3. Enhanced User Review Gate Documentation
- Added detailed step-by-step instructions
- Clarified the loop mechanism
- Added critical reminder: must wait for explicit user confirmation before proceeding

## Testing & Validation

Successfully tested through 5 rounds of modify-review cycles:
- ✅ Each modification correctly triggers a re-prompt for review
- ✅ Supports multiple consecutive modification rounds
- ✅ Correctly proceeds to next step after user confirmation
- ✅ Never skips the review prompt to jump to implementation

## Impact Scope

- Only affects `skills/brainstorming/SKILL.md`
- Backward compatible, does not break existing functionality
- Improves user experience by enhancing spec review control

## Modified Files

- `skills/brainstorming/SKILL.md` - Primary modification

## Motivation

This change ensures users have full control over the spec review process. Previously, after requesting changes, users had no opportunity to verify the modifications before the AI moved to implementation planning. This improvement enables iterative refinement of specs with proper user oversight at each step.

## Related Issues

N/A - This is a proactive improvement based on user workflow analysis.


obra:
Thanks @Dark1X — a spec review loop that doesn't let the conversation move on until the spec is actually approved is the right shape.

After the 6.0 release: **6.0.0's brainstorming already ships the user-review-until-approved loop.** Verified against v6.0.2 — `skills/brainstorming/SKILL.md` loops on "changes requested → re-run the spec review loop" and states "only proceed once the user approves," with the flowchart looping the user-review gate back to the design doc.

The reason I'm closing this rather than merging: your PR is built on the older `spec-document-reviewer` *subagent* review loop, and 6.0 replaced that model with an inline spec self-review plus the user-approval loop above. So the deliverable (don't advance until approved) is present, but the mechanism your diff targets no longer exists.

If you think the *automated* self-review specifically should re-dispatch until it comes back clean (beyond the single inline pass 6.0 does), that's a real, separate idea — please open a fresh issue describing that gap and I'll make sure it gets looked at.

---

PR #704, opened 2026-03-12
docs: 📚 Use OPENCODE_CONFIG_DIR in install docs

cavanaug:
This PR supersedes #506.

I initially opened #506 from branch `cavanaug/main`. I could retarget the base to `dev`, but GitHub does not support changing an existing PR head branch from `main` to `dev`.

To provide a clean diff against `dev`, I replayed the same docs change onto branch `cavanaug/dev` and opened this replacement PR.

Please review this PR instead of #506.

Refs: #506

### DIFF
--- a/.opencode/INSTALL.md
+++ b/.opencode/INSTALL.md
@@ -7,33 +7,41 @@
 
 ## Installation Steps
 
-### 1. Clone Superpowers
+> **Path note:** Use `OPENCODE_CONFIG_DIR` as the install root when it is set. Otherwise, use the default config directory (`~/.config/opencode` on macOS/Linux).
+
+### 1. Set your config directory
 
 ```bash
-git clone https://github.com/obra/superpowers.git ~/.config/opencode/superpowers
+OPENCODE_CONFIG_DIR="${OPENCODE_CONFIG_DIR:-$HOME/.config/opencode}"
 ```
 
-### 2. Register the Plugin
+### 2. Clone Superpowers
+
+```bash
+git clone https://github.com/obra/superpowers.git "${OPENCODE_CONFIG_DIR:-$HOME/.config/opencode}/superpowers"
+```
+
+### 3. Register the Plugin
 
 Create a symlink so OpenCode discovers the plugin:
 
 ```bash
-mkdir -p ~/.config/opencode/plugins
-rm -f ~/.config/opencode/plugins/superpowers.js
-ln -s ~/.config/opencode/superpowers/.opencode/plugins/superpowers.js ~/.config/opencode/plugins/superpowers.js
+mkdir -p "${OPENCODE_CONFIG_DIR:-$HOME/.config/opencode}/plugins"
+rm -f "${OPENCODE_CONFIG_DIR:-$HOME/.config/opencode}/plugins/superpowers.js"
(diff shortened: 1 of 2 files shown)


obra:
I think this might be a lot cleaner if you configured the environment variable at the start of each of the larger processes so you don't have to constantly repeat the alternation between the environment variable and the default path.

cavanaug:
I can change to that.   My concern perhaps not well founded is that someone decides to not follow exactly from the beginning and cuts/pastes something from the middle or a section and then OPENCODE_CONFIG_DIR is blank.

obra:
There's basically nothing we can do to protect users from...following only some of the instructions. But I'll point claude at taking a pass on this.

obra:
Looking at this a little harder, these instructions have gotten *complicated* and should either be a script or...does opencode really still not have installable plugins?

obra:
Thanks for this work @cavanaug! We've been working on simplifying the OpenCode install story even further.

PR #753 takes a different approach — instead of symlinks and environment variables, superpowers now installs as a native OpenCode plugin with a single line in `opencode.json`:

```json
{
  "plugin": ["superpowers@git+https://github.com/obra/superpowers.git"]
}
```

The plugin's `config` hook auto-registers the skills directory in-memory, so no symlinks, no `OPENCODE_CONFIG_DIR`, no manual setup needed at all.

Would you mind giving it a try and letting us know if it works for your setup? Especially interested in whether it handles your non-default config directory case cleanly.

cavanaug:
> Thanks for this work @cavanaug! We've been working on simplifying the OpenCode install story even further.
> 
> PR #753 takes a different approach — instead of symlinks and environment variables, superpowers now installs as a native OpenCode plugin with a single line in `opencode.json`:
> 
> ```json
> {
>   "plugin": ["superpowers@git+https://github.com/obra/superpowers.git"]
> }
> ```
> 
> The plugin's `config` hook auto-registers the skills directory in-memory, so no symlinks, no `OPENCODE_CONFIG_DIR`, no manual setup needed at all.
> 
> Would you mind giving it a try and letting us know if it works for your setup? Especially interested in whether it handles your non-default config directory case cleanly.

This is a MUCH BETTER approach in general, but it does expose an inability to locally customize those agent definitions.

One thing related to opencode for you to consider is some of the interactions with subagents and github copilot premium request consumption.   There was a similar problem with GSD over consuming premium requests in opencode as well.   This is solved by additional front matter in the agent definitions, like the following with the mode being critical.   I was able to do this manually when the files were local, now with them loaded via plugin its something I cant do.   I guess I should file a separate issue on this..

model: inherit   (This is tbd depending on what you want to accomplish)
mode: subagent    (This is key to not overconsuming premium requests)

(Reference https://github.com/gsd-build/get-shit-done/issues/776)

obra:
Thanks for the docs improvement! The OpenCode integration has been reworked since this was filed — the plugin now auto-registers skills via a config hook, and the install docs have been updated accordingly. The OPENCODE_CONFIG_DIR handling is now done in the plugin code itself. Closing as obsolete.

arittr:
Follow-up note: the newer OpenCode install-doc cleanup landed in #1284. This keeps the native plugin install path, softens the install/update claims, and documents the Windows/system-npm workaround instead of reviving the older OPENCODE_CONFIG_DIR/symlink flow from this PR.

PR #740, opened 2026-03-15
fix: auto-foreground on Windows/Git Bash where nohup reaps background processes

### LINKED ISSUE #737: Visual companion brainstorm server fails on Windows (Git Bash)
## Problem

The brainstorming visual companion server (`skills/brainstorming/scripts/start-server.sh`) fails silently on Windows 10 with Git Bash. The server process gets launched but dies immediately before producing any output.

**Evidence from a failed session** (`.superpowers/brainstorm/<session-id>/`):
- `.server.log` — 0 bytes (empty, server never wrote output)
- `.server.pid` — written (PID was assigned)
- `.server-info` — missing (server never reached startup)
- `.server-stopped` — missing (no graceful shutdown)

## Root Cause

`start-server.sh` uses several Unix-specific mechanisms that don't work on Windows/Git Bash:

1. **`nohup ... &` + `disown`** (line 108-110): Unix process backgrounding. On Windows Git Bash, the spawned process is reaped immediately. The server dies before it can write anything to the log file.

2. **`ps -o ppid= -p $PPID`** (line 94): Unix process tree inspection for resolving the owner PID. This command doesn't work correctly on Windows.

3. **`process.kill(OWNER_PID, 0)`** in `server.js` (line 313): Signal 0 is a Unix concept for checking if a process exists. Behavior is unreliable on Windows.

The script's liveness check (lines 114-133) then either:
- Times out after 5 seconds with `"Server failed to start within 5 seconds"`
- Detects the dead PID and reports `"Server started but was killed"`

## Environment
- Windows 10, Git Bash (via Git for Windows)
- Superpowers v5.0.2
- Claude Code CLI
- Node.js available in PATH

## Workaround

Using `--foreground` mode works — the script runs `node server.js` directly (line 102) without Unix daemon mechanics. Combined with the agent's `run_in_background` Bash parameter, the server stays alive. But the skill's instructions don't account for this on Windows, so the agent always uses the default (broken) background mode.

## Suggested Fix

Options:
1. **Detect Windows in `start-server.sh`** and auto-switch to `--foreground` mode (similar to the existing `CODEX_CI` detection on line 63)
2. **Update `visual-companion.md`** with a Windows section advising `--foreground` mode
3. **Use Node.js child_process** for backgrounding instead of shell-level `nohup`/`disown`, which would be cross-platform

Note: The plugin already has excellent Windows awareness for hooks (see `docs/windows/polyglot-hooks.md`), so this is likely just an oversight for the newer brainstorm server feature.

Bortlesboat:
## Problem

`start-server.sh` fails silently on Windows 10/11 with Git Bash (issue #737).
The server process is launched but reaped immediately by the Windows process
model before it can write any output. Symptoms:
- `.server.log` — 0 bytes
- `.server.pid` — written (PID was assigned)
- `.server-info` — missing (server never reached startup)
- Error after 5s: `"Server failed to start within 5 seconds"`

The `--foreground` flag works as a manual workaround, but agents running on
Windows have no way to know they need it.

## Fix

Detect Windows/Git Bash via `OSTYPE` (`msys*`, `cygwin*`, `mingw*`) and `MSYSTEM`
(set by all Git Bash variants: `MINGW64`, `MINGW32`, `MSYS`) and auto-switch to
foreground mode — the same pattern already used for Codex CI detection on line 63.

Also skip the `ps -o ppid=` grandparent-PID resolution on Windows, where MSYS2's
minimal `ps` doesn't support the `-o` flag. Fall back directly to `$PPID`.

Update `visual-companion.md` to document the Windows behavior in the
"Launching the server by platform" section.

## Test plan
- [ ] macOS/Linux: default background mode still works (`OSTYPE` is `darwin*`/`linux*`)
- [ ] Windows Git Bash: server starts in auto-foreground mode
- [ ] `--foreground` flag still works on all platforms
- [ ] `--background` flag overrides auto-detection (`FORCE_BACKGROUND` check preserved)
- [ ] Codex CI auto-foreground unaffected (checked before Windows block)

Fixes #737

### DIFF
--- a/skills/brainstorming/scripts/start-server.sh
+++ b/skills/brainstorming/scripts/start-server.sh
@@ -64,6 +64,17 @@ if [[ -n "${CODEX_CI:-}" && "$FOREGROUND" != "true" && "$FORCE_BACKGROUND" != "t
   FOREGROUND="true"
 fi
 
+# Windows/Git Bash reaps nohup background processes. Auto-foreground when detected.
+if [[ "$FOREGROUND" != "true" && "$FORCE_BACKGROUND" != "true" ]]; then
+  case "${OSTYPE:-}" in
+    msys*|cygwin*|mingw*) FOREGROUND="true" ;;
+  esac
+  # Git Bash sets MSYSTEM (MINGW64, MINGW32, MSYS)
+  if [[ -n "${MSYSTEM:-}" ]]; then
+    FOREGROUND="true"
+  fi
+fi
+
 # Generate unique session directory
 SESSION_ID="$$-$(date +%s)"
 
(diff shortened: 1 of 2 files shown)


obra:
Fixed on dev with a more complete solution: auto-foreground detection (same approach as this PR) plus a documented known issue about MSYS2/Windows PID namespace mismatch in the OWNER_PID lifecycle check. Thanks for the excellent diagnosis and fix!

PR #741, opened 2026-03-15
[feat] Add the support for windsurf

shareworker:
Add the support for windsurf


obra:
Hi, 
Thanks so much for submitting this. 
It looks like the instructions for install and update are duplicated. Would it be possible to make it so that there is only a single copy of each instruction in the code base? 
Also, if it would be possible to re-target it against the dev branch, I'd really appreciate it.

shareworker:
Hi, I have updated the pull. Please review.

Chen-Scrumsdotcom:
Need this in my life!

kurroo10:
any update ?

shareworker:
This installation method is not elegant enough. We can wait for support from the windsurf official plugin marketplace. For now, it can be installed using npx openskills install obra/superpowers.

PR #759, opened 2026-03-16
fix(brainstorming): auto-open visual companion URL on WSL2/macOS/Linux

### LINKED ISSUE #755: Visual companion: browser doesn't open on WSL2
## Problem

The visual companion's `start-server.sh` starts the HTTP server successfully on WSL2, but the user is told to "open the URL" manually. There's no attempt to launch a browser, and on WSL2 the standard `xdg-open` / `open` commands don't work anyway.

The user sees the server start, gets told to visit `http://localhost:63289`, but nothing opens — they have to manually copy/paste the URL or be told it exists.

## Environment

- WSL2 (Ubuntu) on Windows
- No X server / display required — the browser runs on the Windows side

## Fix

When the server starts (or in the skill instructions), detect WSL2 and use the Windows browser opener:

```bash
/mnt/c/Windows/system32/cmd.exe /c start http://localhost:$PORT
```

This works because localhost on WSL2 is shared with the Windows host, so the Windows browser can reach the server.

More robustly, the `start-server.sh` script could auto-detect the platform and open the browser:

```bash
if grep -qi microsoft /proc/version 2>/dev/null; then
  # WSL2
  /mnt/c/Windows/system32/cmd.exe /c start "$URL" &>/dev/null &
elif command -v open &>/dev/null; then
  # macOS
  open "$URL"
elif command -v xdg-open &>/dev/null; then
  # Linux with desktop
  xdg-open "$URL"
fi
```

Alternatively, the skill instructions in `visual-companion.md` could mention that on WSL2 the agent should open the browser using the Windows cmd approach, since the current "Tell user to open the URL" step doesn't work well when the user expects things to just appear.

jasonandjay:
Fixes #755

## Problem
`start-server.sh` returns the visual companion URL, but in WSL2 environments users often expect the browser to open automatically. Existing openers either don't run or don't target the Windows browser path from WSL.

## What changed
- Added best-effort browser opener logic to `skills/brainstorming/scripts/start-server.sh`:
  - Detect WSL and open via `cmd.exe /c start`
  - Use `open` on macOS
  - Use `xdg-open` on Linux desktop sessions (only when `DISPLAY`/`WAYLAND_DISPLAY` is set)
- Trigger opener after parsing the server-started JSON line (`url`)
- Updated `skills/brainstorming/visual-companion.md` to document auto-open behavior and fallback expectation (always echo URL).

## Notes
- Best-effort only: failures are swallowed and never block server startup
- Keeps existing JSON startup output unchanged for callers

### DIFF
--- a/skills/brainstorming/scripts/start-server.sh
+++ b/skills/brainstorming/scripts/start-server.sh
@@ -16,6 +16,40 @@
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 
+is_wsl() {
+  grep -qiE "microsoft|wsl" /proc/version 2>/dev/null
+}
+
+open_url_best_effort() {
+  local url="$1"
+
+  # WSL: open in Windows default browser
+  if is_wsl; then
+    if [[ -x "/mnt/c/Windows/System32/cmd.exe" ]]; then
+      /mnt/c/Windows/System32/cmd.exe /c start "" "$url" >/dev/null 2>&1 || true
+      return 0
+    fi
+    if [[ -x "/mnt/c/Windows/system32/cmd.exe" ]]; then
+      /mnt/c/Windows/system32/cmd.exe /c start "" "$url" >/dev/null 2>&1 || true
+      return 0
+    fi
+  fi
+
+  # macOS
+  if command -v open >/dev/null 2>&1; then
+    open "$url" >/dev/null 2>&1 || true
+    return 0
+  fi
+
+  # Linux desktop (skip on headless shells)
+  if command -v xdg-open >/dev/null 2>&1 && [[ -n "${DISPLAY:-}${WAYLAND_DISPLAY:-}" ]]; then
+    xdg-open "$url" >/dev/null 2>&1 || true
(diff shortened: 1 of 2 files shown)


obra:
Thanks @jasonandjay — auto-opening the companion across platforms is a nice quality-of-life fix.

After the 6.0 release: **6.0.0 ships cross-platform auto-open for the visual companion.** Verified against v6.0.2 in `skills/brainstorming/scripts/server.cjs`:

- `browserLauncherForPlatform()` covers macOS (`open`), Windows/WSL (detects `/microsoft/` in the kernel release and uses `cmd.exe`/`rundll32`), and Linux (`xdg-open`, gated on `DISPLAY`/`WAYLAND_DISPLAY`)
- `maybeOpenBrowser()` is wired to the `--open` flag and fires on the first screen

That's the same WSL2/macOS/Linux matrix your PR targets, behind a user-consent flag. Closing as resolved in 6.0.

If the companion fails to open on a platform you use under 6.0.2, please reopen or file a fresh issue with your OS/shell details.

---

PR #768, opened 2026-03-17
docs: fix visual companion launch for Claude Code on Windows

### LINKED ISSUE #767: Visual brainstorming server dies on Windows — nohup/disown doesn't survive shell exit
**Describe the bug**
The visual brainstorming server (started via `scripts/start-server.sh`) dies between conversation turns on Windows. The `nohup`/`disown` approach on line 108-110 of `start-server.sh` does not keep the Node server alive when Claude Code's ephemeral shell exits, because Windows (Git Bash/MSYS2) lacks true Unix process groups.

**To Reproduce**
1. Run Claude Code on Windows 11 (Git Bash shell)
2. Invoke the brainstorming skill on a topic that triggers the visual companion
3. Server starts successfully, first HTML screen renders in the browser
4. On the next conversation turn, the server is dead — the browser shows connection refused

**Expected behavior**
The brainstorming server should remain alive across conversation turns, as it does on macOS/Linux.

**Logs**
No error logs — the process is simply gone. `kill -0 $SERVER_PID` fails on the next turn.

**Additional context**
The `--foreground` codepath already exists in `start-server.sh` (lines 100-104) and works correctly. The fix is a documentation change in `visual-companion.md` — the Claude Code launch instructions should tell Windows users to use `--foreground` combined with the Bash tool's `run_in_background: true` parameter, which keeps the process alive at the harness level rather than relying on shell-level backgrounding.

**Suggested fix (visual-companion.md only, no code changes needed):**

Split the current "Claude Code" section under "Launching the server by platform" into two:

```markdown
**Claude Code (Windows):**
` ` `bash
# IMPORTANT: nohup/disown does NOT survive shell exit on Windows.
# Use --foreground and set run_in_background: true on the Bash tool call
# so the harness (not the shell) keeps the process alive across turns.
scripts/start-server.sh --project-dir /path/to/project --foreground
` ` `
When calling this via the Bash tool, set `run_in_background: true`. Then read `$SCREEN_DIR/.server-info` on the next turn to get the URL and port.

**Claude Code (macOS/Linux):**
` ` `bash
# Default mode works — the script backgrounds the server itself
scripts/start-server.sh --project-dir /path/to/project
` ` `
```

Also update the "Finding connection info" paragraph to mention the `run_in_background` pattern for Windows users.

lucasyhzhu-debug:
## Summary
- Split the "Claude Code" platform section into **macOS/Linux** and **Windows** variants
- Documents that `run_in_background: true` is required on the Bash tool call for Windows, since the script's auto-detected foreground mode blocks the tool call otherwise
- Merged the standalone "Windows (Git Bash / CMD / PowerShell)" section into the new "Claude Code (Windows)" section — the auto-detection handles the switch, so separate sections were redundant
- Updated "Finding connection info" paragraph for the `run_in_background` pattern

## Context
The script already auto-detects Windows (via `OSTYPE`/`MSYSTEM`) and forces `--foreground` mode — that part works. But foreground mode means the Bash tool call in Claude Code **blocks forever** (or until timeout), because `node server.js` never returns. The missing documentation was: use `run_in_background: true` on the Bash tool so the Claude Code harness manages the process lifecycle across turns.

Fixes #767

## Test plan
- [x] Verified on Windows 11 + Claude Code + Git Bash
- [ ] macOS/Linux path unchanged (default nohup/disown still works)

### DIFF
--- a/skills/brainstorming/visual-companion.md
+++ b/skills/brainstorming/visual-companion.md
@@ -42,30 +42,31 @@ scripts/start-server.sh --project-dir /path/to/project
 
 Save `screen_dir` from the response. Tell user to open the URL.
 
-**Finding connection info:** The server writes its startup JSON to `$SCREEN_DIR/.server-info`. If you launched the server in the background and didn't capture stdout, read that file to get the URL and port. When using `--project-dir`, check `<project>/.superpowers/brainstorm/` for the session directory.
+**Finding connection info:** The server writes its startup JSON to `$SCREEN_DIR/.server-info`. If you launched the server with `run_in_background: true` (Windows) or in the background and didn't capture stdout, read that file on your next turn to get the URL and port. When using `--project-dir`, check `<project>/.superpowers/brainstorm/` for the session directory.
 
 **Note:** Pass the project root as `--project-dir` so mockups persist in `.superpowers/brainstorm/` and survive server restarts. Without it, files go to `/tmp` and get cleaned up. Remind the user to add `.superpowers/` to `.gitignore` if it's not already there.
 
 **Launching the server by platform:**
 
-**Claude Code:**
+**Claude Code (macOS/Linux):**
 ```bash
 # Default mode works — the script backgrounds the server itself
 scripts/start-server.sh --project-dir /path/to/project
 ```
 
+**Claude Code (Windows):**
+```bash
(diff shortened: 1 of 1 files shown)


obra:
Thanks @lucasyhzhu-debug! This is already covered on the dev branch — commit f34ee47 made the same changes (split Claude Code into macOS/Linux and Windows sections, document `run_in_background: true` for Windows). Closing as already addressed.

PR #769, opened 2026-03-17
chore: close internal-only draft

qfedesq:
Closing this draft because the work is intended for internal-only use and will not be proposed upstream.


obra:
Hi. This is a local only server. What's the threat model you're considering for these changes?

PR #784, opened 2026-03-17
fix: resolve ESM/CommonJS conflict in brainstorming server  #774

### LINKED ISSUE #774: Visual companion server.js fails on Node.js v25+ (ESM/CommonJS conflict)
## Bug Description

The brainstorming visual companion server fails to start on Node.js v25+ with:

```
ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and
'.../superpowers/5.0.4/package.json' contains "type": "module".
```

## Root Cause

`skills/brainstorming/scripts/server.js` uses CommonJS (`require()`), but the root `package.json` declares `"type": "module"`. Node.js v25 enforces this strictly — all `.js` files under a `"type": "module"` package are treated as ESM, causing `require()` to throw.

## Steps to Reproduce

1. Have Node.js v25+ installed
2. Start a brainstorming session that triggers the visual companion
3. Server fails with the error above

## Fix

Rename `server.js` → `server.cjs` and update the two `node server.js` invocations in `start-server.sh` to `node server.cjs`. The `.cjs` extension forces CommonJS mode regardless of the `package.json` `type` field.

Alternatively, the server script could be rewritten to use ESM imports.

## Environment

- Node.js: v25.8.1
- Superpowers: v5.0.4
- Platform: macOS (Darwin 25.3.0)

sarbojitrana:
###  Fixes  #774

Resolves a runtime error in Node.js v25+ where the brainstorming server fails to start due to a module system mismatch.

### Problem

The repository is configured with `"type": "module"` (ESM), but `server.js` uses CommonJS (`require`).

Node.js v25 enforces ESM strictly, resulting in:

```
ReferenceError: require is not defined in ES module scope
```

---

### Solution

* Renamed `server.js` → `server.cjs` to explicitly use CommonJS
* Updated `start-server.sh` to reference the new file
* Updated test references to maintain consistency

### Before the fix 

image

### After the fix

image


obra:
Thanks @sarbojitrana! Clean fix — exactly the right approach. Cherry-picked your commit to the dev branch (preserving your authorship). Closing the PR since we applied it directly rather than merging.

PR #816, opened 2026-03-18
feat: add safe PR checkout pattern to using-git-worktrees skill

jpassaro:
## Summary

- Adds a new "Checking Out a PR to a Worktree" section to the `using-git-worktrees` skill with a safe pattern: `gh pr view --json` → `git fetch` → `git worktree add`
- Adds a "Common Mistakes" entry warning against using `gh pr checkout` before creating a worktree
- Adds a Quick Reference table row for the PR checkout workflow

## Motivation

Using `gh pr checkout` in a worktree moves HEAD, which can disrupt in-progress work (e.g., detached HEAD with an uncommitted rebase). The safe pattern fetches the remote ref and creates the worktree directly, never touching the current worktree's HEAD.

## Test plan

- [x] Used via `--plugin-dir` for two weeks of daily usage across multiple repos
- [x] Verified correct PR worktree behavior without unintended side effects
- [ ] Review: instructions are clear and match existing skill conventions

### DIFF
--- a/skills/using-git-worktrees/SKILL.md
+++ b/skills/using-git-worktrees/SKILL.md
@@ -152,8 +152,35 @@ Ready to implement <feature-name>
 | Directory not ignored | Add to .gitignore + commit |
 | Tests fail during baseline | Report failures + ask |
 | No package.json/Cargo.toml | Skip dependency install |
+| Checking out a PR          | `gh pr view` + `git fetch` + `git worktree add` (never `gh pr checkout`) |
+
+## Checking Out a PR to a Worktree
+
+To review or work on an existing PR in an isolated worktree, **never use `gh pr checkout`** — it moves HEAD in the current worktree, which can disrupt in-progress work (detached HEAD, uncommitted rebase, etc.).
+
+Instead, fetch the branch by name without touching HEAD:
+
+```bash
+# Get the branch name without checking it out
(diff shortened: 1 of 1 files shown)


obra:
I've never seen Claude get confused by this issue. Can you talk a little bit about the problem you ran into that inspired this PR?

jpassaro:
Certainly. On my main working copy I was in the middle of a lengthy rebase, i.e. detached HEAD. i saw a new PR and wanted to check out the branch.

as a human i would have done one of these:

1. get the branch name in the github UI and done `git worktree add` referencing the branch
2. add worktree, cd there, `gh pr checkout`

either way leaving my working copy alone, it's in a state i can't comfortably check out from and restore.

asking claude to do it, it did `gh pr checkout` right there in my mid-rebase working copy in order to get the branch, before adding the worktree. my rebase seemed to be gone.

lucky for me, i was able to use git reflog to recover my rebase state, nothing was lost. it was a bit scary though! I'm not sure git really documents what happens if you check out mid-rebase without completing or aborting, maybe i need not have panicked, but i doubt any git expert would recommend it under normal circumstances.

it might be worth reviewing other places where these skills would lead claude to do a questionable checkout during exotic git states, but at the very least this case has an easy alternative: we just need the PR branch, so instead of a risky checkout, get it from `gh pr view`.

Issue #818, opened 2026-03-18
Plugin skills cache incomplete on Claude Code — only 7/14 skills downloaded


MafraAiDev:
## Bug Description

When the superpowers plugin is installed via Claude Code's plugin system, the `skills/` directory is sometimes cloned incompletely. Only a subset of skill directories end up in the local cache, causing `Skill("superpowers:subagent-driven-development")` and other skills to return `Unknown skill`.

This is **not a bug in superpowers itself** — the skill files are correct in the repo. The bug is in Claude Code's plugin download mechanism (reported as [anthropics/claude-code#35989](https://github.com/anthropics/claude-code/issues/35989)). Opening this issue here so superpowers users who hit this problem can find the workaround.

## Environment

- Claude Code: 2.1.79
- OS: WSL2 (Ubuntu) on Linux 6.6.87.2-microsoft-standard-WSL2
- Node.js: v22 (via nvm)
- superpowers: v5.0.5

## Symptoms

- `Skill("superpowers:subagent-driven-development")` → `Unknown skill`
- Same for `dispatching-parallel-agents`, `executing-plans`, `requesting-code-review`, `test-driven-development`, `brainstorming`, `using-superpowers`
- `writing-plans` tells you to use `subagent-driven-development` but it's not available
- Only 7 of 14 skills are registered in the session

## Evidence

The download counts varied across versions and marketplace sources:

| Cache source | v5.0.1 | v5.0.2 | v5.0.4 | v5.0.5 |
|---|---|---|---|---|
| `claude-plugins-official` | 14 ✅ | 14 ✅ | 14 ✅ | 14 ✅ |
| `superpowers-marketplace` | 9 ❌ | 0 ❌ | 10 ❌ | 7 ❌ |

The `claude-plugins-official` source always had all 14 skills. The `superpowers-marketplace` source had inconsistent/incomplete counts, suggesting a race condition or sparse checkout issue in Claude Code's git clone mechanism.

**Skills present** (7): finishing-a-development-branch, receiving-code-review, systematic-debugging, using-git-worktrees, verification-before-completion, writing-plans, writing-skills

**Skills missing** (7): brainstorming, dispatching-parallel-agents, executing-plans, requesting-code-review, subagent-driven-development, test-driven-development, using-superpowers

## Workaround

If you have a complete cache from `claude-plugins-official`, copy the missing skills:

```bash
[...]

PR #819, opened 2026-03-18
feat: add verify-install.sh to detect incomplete skill downloads

ThiagoMafra-Integrare:
## Summary

- Adds `verify-install.sh` script that checks all expected skill directories are present in the plugin cache
- If skills are missing, reports which ones and points to the upstream Claude Code issue
- **Diagnostic only** — does not attempt to fix the cache (per maintainer feedback, manual patching is fragile)

## Context

Claude Code's plugin download mechanism sometimes clones the `skills/` directory incompletely. We observed inconsistent skill counts across versions and marketplace sources:

| Cache source | v5.0.1 | v5.0.2 | v5.0.4 | v5.0.5 |
|---|---|---|---|---|
| `claude-plugins-official` | 14 | 14 | 14 | 14 |
| `superpowers-marketplace` | 9 | 0 | 10 | 7 |

Users hit `Unknown skill: superpowers:subagent-driven-development` (and other skills) with no indication that the installation is incomplete. This script helps them understand *why* skills are missing.

The root cause is in Claude Code's git clone logic (tracked at anthropics/claude-code#35989).

## Usage

```bash
# Auto-detect plugin location
bash verify-install.sh

# Or specify path
bash verify-install.sh ~/.claude/plugins/cache/superpowers-marketplace/superpowers/5.0.5
```

**Output when all skills present:**
```
Superpowers skill verification: 14/14 skills present
All skills installed correctly.
```

**Output when skills missing:**
```
Superpowers skill verification: 7/14 skills present
Missing skills (7):
  ✗ brainstorming
  ✗ dispatching-parallel-agents
  ...

This is a known issue with Claude Code's plugin download mechanism.
The fix needs to come from Claude Code itself.
See: https://github.com/anthropics/claude-code/issues/35989
```

## Related

- anthropics/claude-code#35989 — root cause in Claude Code's plugin download
- #818 — user-facing impact report

## Test plan

- [x] Tested against complete repo clone (14/14)
- [x] Tested against simulated incomplete install (detects missing, reports correctly)
- [x] Script exits 0 on success, 1 on missing skills
- [ ] Test on macOS (uses POSIX-safe constructs, should work)

### DIFF
--- a/verify-install.sh
+++ b/verify-install.sh
@@ -98,16 +98,8 @@ main() {
   done
 
   echo ""
-  echo -e "${YELLOW}To fix, run:${NC}"
-  echo ""
-  echo "  git clone https://github.com/obra/superpowers.git /tmp/superpowers-fix"
-  echo "  cp -r /tmp/superpowers-fix/skills/{$(IFS=,; echo "${missing[*]}")} \\"
-  echo "    $skills_dir/"
-  echo "  rm -rf /tmp/superpowers-fix"
-  echo ""
-  echo "Then restart your Claude Code session."
-  echo ""
   echo -e "${YELLOW}This is a known issue with Claude Code's plugin download mechanism.${NC}"
+  echo "The fix needs to come from Claude Code itself."
   echo "See: https://github.com/anthropics/claude-code/issues/35989"
 
   exit 1


ThiagoMafra-Integrare:
Closing per maintainer feedback — this should be fixed at the Claude Code level, not with workarounds in the plugin repo.

PR #822, opened 2026-03-19
fix: add Quick Reference table to verification skill

jason-h23:
## Summary

Adds an actionable lookup table mapping common completion claims to their required verification commands.

Complements the existing evidence-vs-assumption table with a "what do I actually run?" reference.

## Changes

- `skills/verification-before-completion/SKILL.md` — Quick Reference table added


jason-h23:
Same — review agent suggestion, not driven by an actual issue. Closing.

PR #823, opened 2026-03-19
fix: replace ≠ unicode with plain text in verification skill

jason-h23:
## Summary

Replace `≠` unicode symbols with "does not equal" in the rationalization prevention table.

LLMs can tokenize unicode symbols inconsistently — plain text is more reliably parsed.

## Changes

- `skills/verification-before-completion/SKILL.md` — 3 lines: `≠` → "does not equal"

### DIFF
--- a/skills/verification-before-completion/SKILL.md
+++ b/skills/verification-before-completion/SKILL.md
@@ -65,11 +65,11 @@ Skip any step = lying, not verifying
 | Excuse | Reality |
 |--------|---------|
 | "Should work now" | RUN the verification |
-| "I'm confident" | Confidence ≠ evidence |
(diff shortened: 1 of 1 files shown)


obra:
Is this based on an actual issue you ran into? What's the prompt you used to generate this change?

jason-h23:
Not based on an actual parsing failure I observed — my code review agent flagged it as a potential robustness issue, since unicode symbols can tokenize inconsistently across different models.

That said, I haven't seen it cause a real problem in practice. If you consider it unnecessary, happy to close this one.

jason-h23:
You're right to ask — I don't have a concrete reproduction of this causing an actual parsing issue. Closing this since the change isn't backed by evidence.

PR #826, opened 2026-03-19
fix: connect EnterPlanMode check in flow diagram

jason-h23:
## Summary

The flow diagram defines an "About to EnterPlanMode?" node and a "User message received" node but has no edge between them. Adds the missing connection.

## Changes

- `skills/using-superpowers/SKILL.md` — 1 line added to dot graph


obra:
Is this because of an actual problem you ran into? I don't think this is a correct workflow.

PR #216, opened 2025-12-30
Fix: Update OpenCode skills directory path to match official spec

chada:
Update all references from 'opencode/skills' to 'opencode/skill' to align with OpenCode Agent Skills documentation (https://opencode.ai/docs/skills)

Changes:
- .opencode/plugin/superpowers.js: Update project and personal skills dir paths from plural `skills` to singular `skill`.
- docs/README.opencode.md: Update all path references in documentation to reflect the correct directory structure.

This fixes a critical issue where `find_skills` cannot locate user-defined skills, and users are prevented from applying a workaround by OpenCode core itself.

## Motivation and Context
The official OpenCode documentation mandates that user-defined skills reside in `.opencode/skill` (singular).

The current implementation of `superpowers` incorrectly searches for a plural `skills` directory.
Crucially, **users cannot manually rename their directory to `skills` to work around this bug**, because OpenCode core strictly enforces the naming convention and throws the following error on startup:

> Directory "skills" in /Users/.../.config/opencode is not valid. Rename the directory to "skill" or remove it. This is a common typo.

This creates a deadlock where the plugin expects a directory name that the core application explicitly forbids. This PR resolves this conflict by aligning the plugin with the core specification.

## How Has This Been Tested?
I tested this locally on macOS:
1. Renamed my local config directory to `skill` (as required by OpenCode core).
2. Applied the fix to `superpowers.js`.
3. Ran `use find_skills tool` within OpenCode.
4. **Result**: The agent successfully located and listed the skills from `~/.config/opencode/skill` (global skills) and `.opencode/skill` (project skills).

## Breaking Changes
None.
Previously, it was impossible to use custom skills with this plugin because OpenCode core would reject the `skills` directory required by the plugin. This change enables the intended functionality.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Summary by CodeRabbit

* **Chores**
  * Reorganized skills directory paths: personal skills now at ~/.config/opencode/skill/ and project skills at .opencode/skill/
  * Updated user-facing messages and error text to reference the new singular "skill" paths
[...]

PR #848, opened 2026-03-20
fix(opencode): align skills path messaging and tests

### LINKED ISSUE #847: OpenCode skill path is inconsistent between runtime, bootstrap text, and tests

## Environment

| Field | Value |
|-------|-------|
| Superpowers version | 5.0.4 |
| Harness (Claude Code, Cursor, etc.) | OpenCode |
| Harness version | |
| Model | openai/gpt-5.4 |
| OS + shell | macOS (darwin) |

## What happened?
OpenCode skill path handling is internally inconsistent after the auto-register change. The plugin now loads bundled skills from the installed package path (for example `node_modules/superpowers/skills`), but the bootstrap text still tells users that Superpowers skills live in `~/.config/opencode/skills/superpowers/`, and the OpenCode tests were still checking `~/.config/opencode/superpowers/skills`.

This leaves three different paths presented as the source of truth:

1. Runtime plugin behavior: package-local `skills/`
2. Bootstrap help text: `~/.config/opencode/skills/superpowers/`
3. Tests: `~/.config/opencode/superpowers/skills`

That makes debugging installation issues and maintaining the OpenCode integration unnecessarily confusing.

## Steps to reproduce
1. Install Superpowers in OpenCode using the current plugin-based installation flow.
2. Inspect `.opencode/plugins/superpowers.js` and note that runtime registration uses `path.resolve(__dirname, '../../skills')`.
3. Start OpenCode with Superpowers enabled and inspect the injected bootstrap text; it reports `~/.config/opencode/skills/superpowers/` as the skills location.
4. Inspect `tests/opencode/test-plugin-loading.sh` and `tests/opencode/test-priority.sh`; they check `~/.config/opencode/superpowers/skills` instead.

## Expected behavior
OpenCode runtime behavior, user-facing bootstrap text, and OpenCode tests should all describe and validate the same skills path model.

## Actual behavior
They describe three different path models at once, even though the real runtime source is the bundled package `skills/` directory.

Relevant implementation evidence:

- `.opencode/plugins/superpowers.js:51` registers the runtime skills directory with `path.resolve(__dirname, '../../skills')`
- `.opencode/plugins/superpowers.js:72` reports `~/.config/opencode/skills/superpowers/` in bootstrap text
- `tests/opencode/test-plugin-loading.sh:35` and `tests/opencode/test-plugin-loading.sh:45` check `~/.config/opencode/superpowers/skills`
- `tests/opencode/test-priority.sh:21` and `tests/opencode/test-priority.sh:68` also assume `~/.config/opencode/superpowers/skills`

Related history:

- `#328` includes a user report noting that the plugin message and actual skill location did not match
- `#339` reports OpenCode skill installation/path confusion
- `#343` is another OpenCode path/documentation inconsistency (`plugin` vs `plugins`)
- `#753` changed OpenCode to auto-register bundled skills, which appears to be where the current runtime/documentation split was introduced

LanternCX:
Closes #847

The OpenCode integration drifted into three different skill path models after the auto-register change. At runtime, the plugin loads bundled skills from the installed package path (for example `node_modules/superpowers/skills`), but the bootstrap text still told users that skills lived in `~/.config/opencode/skills/superpowers/`, and the OpenCode tests were still checking `~/.config/opencode/superpowers/skills`.

That mismatch makes installation debugging and maintenance confusing: the product behavior, user-facing guidance, and test suite were no longer describing the same system.

## What does this PR change?
This PR updates the OpenCode bootstrap text to report the actual runtime skills path and updates the OpenCode test scripts to derive and validate the runtime-installed skills directory instead of hardcoding older path layouts. It also updates outdated OpenCode test wording that still referred to the removed `use_skill` / `find_skills` flow.

Yes. This is a core OpenCode integration consistency fix for Superpowers itself. It affects bundled plugin behavior and the repository's own OpenCode test suite, and is useful to any OpenCode user or contributor working with Superpowers.

- Keep the old bootstrap wording and only adjust tests. I rejected this because it would preserve user-facing misinformation.
- Revert runtime behavior back to a config-directory-based path. I rejected this because it would fight the newer auto-register design introduced for OpenCode.
- Only document the discrepancy in comments or README text. I rejected this because the inconsistency exists in executable code and tests, so the fix should live there.

No. All changes are part of the same OpenCode path-consistency fix: aligning runtime messaging and tests with the actual bundled-skills loading model.

## Existing PRs
- Related PRs: `#330`, `#753`, `#216`

`#330` moved OpenCode to native skills and a symlink-based model. `#753` later switched to auto-registering bundled skills from the plugin package, which is where the current runtime/message split appears to have been introduced. `#216` discussed OpenCode skill path mismatches more generally, but was about a different `skill` vs `skills` claim and was closed as unreproducible.

## Environment tested

| OpenCode | | GPT-5.4 | openai/gpt-5.4 |

## Evaluation
- Initial prompt: `fix issue #847`
- Eval sessions run after the change: 2 direct test runs plus 1 suite run
- Before: `tests/opencode/test-plugin-loading.sh` passed legacy path assumptions and did not guard against stale bootstrap text; the OpenCode integration still reported an outdated skills path. After: `bash tests/opencode/test-plugin-loading.sh` fails if the plugin references the old symlink path and now passes with the runtime-path message; `bash tests/opencode/run-tests.sh --verbose` also passes with the updated checks.

## Rigor

- [ ] If this is a skills change: I used `superpowers:writing-skills` and

Adversarial coverage used here:
- Added a test assertion that explicitly fails if `.opencode/plugins/superpowers.js` still contains the old `\${configDir}/skills/superpowers/` path.
- Re-ran the OpenCode plugin test script after updating the implementation.
- Re-ran the non-integration OpenCode test suite to confirm the updated path checks work end-to-end.

## Human review

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -69,7 +69,7 @@ When skills reference tools you don't have, substitute OpenCode equivalents:
 - \`Read\`, \`Write\`, \`Edit\`, \`Bash\` → Your native tools
 
 **Skills location:**
-Superpowers skills are in \`${configDir}/skills/superpowers/\`
+Superpowers skills are loaded from \`${superpowersSkillsDir}\`
 Use OpenCode's native \`skill\` tool to list and load skills.`;
 
     return `<EXTREMELY_IMPORTANT>
--- a/tests/opencode/run-tests.sh
+++ b/tests/opencode/run-tests.sh
@@ -44,7 +44,7 @@ while [[ $# -gt 0 ]]; do
             echo ""
             echo "Tests:"
             echo "  test-plugin-loading.sh  Verify plugin installation and structure"
-            echo "  test-tools.sh           Test use_skill and find_skills tools (integration)"
+            echo "  test-tools.sh           Test native skill tool integration"
             echo "  test-priority.sh        Test skill priority resolution (integration)"
             exit 0
             ;;
--- a/tests/opencode/setup.sh
+++ b/tests/opencode/setup.sh
@@ -14,7 +14,7 @@ export OPENCODE_CONFIG_DIR="$TEST_HOME/.config/opencode"
 
 # Install plugin to test location
 mkdir -p "$HOME/.config/opencode/superpowers"
-cp -r "$REPO_ROOT/lib" "$HOME/.config/opencode/superpowers/"
+cp "$REPO_ROOT/package.json" "$HOME/.config/opencode/superpowers/"
 cp -r "$REPO_ROOT/skills" "$HOME/.config/opencode/superpowers/"
 
 # Copy plugin directory
--- a/tests/opencode/test-plugin-loading.sh
+++ b/tests/opencode/test-plugin-loading.sh
@@ -32,7 +32,9 @@ fi
 
 # Test 2: Verify skills directory is populated
 echo "Test 2: Checking skills directory..."
-skill_count=$(find "$HOME/.config/opencode/superpowers/skills" -name "SKILL.md" | wc -l)
+plugin_file="$HOME/.config/opencode/superpowers/.opencode/plugins/superpowers.js"
+runtime_skills_dir=$(node --input-type=module -e "import path from 'path'; console.log(path.resolve(process.argv[1], '../../skills'));" "$(dirname "$plugin_file")")
+skill_count=$(find "$runtime_skills_dir" -name "SKILL.md" | wc -l)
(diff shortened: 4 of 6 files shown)


obra:
Thanks for working on this, @LanternCX! I've fixed the path inconsistency on the dev branch with a slightly different approach — removed the skills path from the bootstrap text entirely (the agent should use the native skill tool, not navigate to files by path) and aligned the test setup to use exported variables from a single source of truth. Appreciate you filing the issue and the PR.

PR #856, opened 2026-03-20
Add exponential backoff and status indicator to brainstorm WebSocket client

chuanghiduoc:
## Summary

The brainstorm companion's WebSocket client (`helper.js`) reconnected with a fixed 1-second delay and provided no visual feedback when the connection dropped. If a user's laptop slept or the server restarted, the browser showed "Connected" indefinitely while the WebSocket was dead, and events were silently queued without any indication.

### Changes

**`skills/brainstorming/scripts/helper.js`:**
- **Exponential backoff**: reconnect delay starts at 500ms, doubles each attempt, caps at 30s. Resets to 500ms on successful connection.
- **Status indicator**: updates the header `.status` element — Connected (green), Reconnecting… (yellow), Disconnected (red) — with matching dot color.
- **`onerror` handler**: delegates to `onclose` to avoid duplicate reconnect attempts.
- **`ws = null` on close**: prevents `sendEvent()` from writing to a dead socket.
- **`clearTimeout`**: cancels pending reconnect timer before creating a new connection.

**`skills/brainstorming/scripts/frame-template.html`:**
- Status dot `::before` pseudo-element now uses a CSS custom property (`--status-color`) instead of hardcoded `var(--success)`, allowing JavaScript to update the dot color dynamically.

### Existing tests

All 24 passing brainstorm-server tests remain passing (the 1 pre-existing failure — `"serves waiting page"` expects `"Waiting for Claude"` but server says `"Waiting for the agent"` — is unrelated, introduced in v5.0.4). All 31 WebSocket protocol tests pass.

## Test plan

- [x] Start brainstorm server, open companion in browser → header shows "Connected" (green dot)
- [x] Kill the server process → header changes to "Disconnected" (red dot)
- [x] Restart server → header auto-recovers to "Connected" (green dot) after backoff
- [x] Verify events queued during disconnect are sent on reconnection

### DIFF
--- a/skills/brainstorming/scripts/frame-template.html
+++ b/skills/brainstorming/scripts/frame-template.html
@@ -73,8 +73,8 @@
       flex-shrink: 0;
     }
     .header h1 { font-size: 0.85rem; font-weight: 500; color: var(--text-secondary); }
-    .header .status { font-size: 0.7rem; color: var(--success); display: flex; align-items: center; gap: 0.4rem; }
-    .header .status::before { content: ''; width: 6px; height: 6px; background: var(--success); border-radius: 50%; }
+    .header .status { font-size: 0.7rem; color: var(--success); display: flex; align-items: center; gap: 0.4rem; --status-color: var(--success); }
+    .header .status::before { content: ''; width: 6px; height: 6px; background: var(--status-color); border-radius: 50%; }
 
     .main { flex: 1; overflow-y: auto; }
     #claude-content { padding: 2rem; min-height: 100%; }
--- a/skills/brainstorming/scripts/helper.js
+++ b/skills/brainstorming/scripts/helper.js
@@ -1,25 +1,53 @@
 (function() {
   const WS_URL = 'ws://' + window.location.host;
+  const BACKOFF_INITIAL = 500;
+  const BACKOFF_MAX = 30000;
+
   let ws = null;
   let eventQueue = [];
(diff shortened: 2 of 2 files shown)


chuanghiduoc:
Thanks for the review! Both comments addressed in 824d4f2:

1. **`onclose` now shows "Reconnecting…"** instead of "Disconnected" — since a reconnect attempt is always scheduled, showing "Disconnected" during the backoff window was misleading. The yellow "Reconnecting…" state correctly reflects that the client is actively retrying.

2. **Removed empty `onerror` handler** — it was a no-op since `onclose` always fires after `onerror` per the WebSocket spec. Removing it eliminates dead code and one less handler to reason about.

obra:
Thanks @chuanghiduoc — resilient reconnect + a visible status indicator is exactly the right direction for the companion client.

After the 6.0 release: **6.0.0 shipped both of these.** Verified against v6.0.2 in `skills/brainstorming/scripts/helper.js`:

- exponential backoff via `nextReconnectDelay` (doubles from 500ms, caps at 30s, resets on connect)
- a Connected / Reconnecting… / Disconnected status pill driven by `setStatus()` / `--status-color`

The shipped version actually goes a bit further (a "paused" overlay and queued-message replay across reconnects). So your change is already in place. Closing as resolved in 6.0.

If you see a reconnect/status edge case that 6.0.2 handles poorly, please reopen or file a fresh issue with repro steps.

---

PR #868, opened 2026-03-21
Fix OpenCode skill-name fallback guidance for superpowers skills

DevNewbie1826:
## Summary
- Add explicit OpenCode skill-name compatibility guidance to the injected bootstrap prompt so agents retry with `superpowers/<skill-name>` when a bare skill name fails.
- Add a regression check in `tests/opencode/test-plugin-loading.sh` to verify the bootstrap contains this fallback guidance.
- Make `tests/opencode/setup.sh` robust when `lib/` is absent so OpenCode plugin tests can run in the current repository layout.

## Verification
- `bash tests/opencode/test-plugin-loading.sh` (pass)
- `bash tests/opencode/run-tests.sh` (pass)
- `node --check .opencode/plugins/superpowers.js` (pass)
- `bash -n tests/opencode/test-plugin-loading.sh` (pass)
- `bash -n tests/opencode/setup.sh` (pass)

## Notes
- `bash tests/opencode/run-tests.sh --integration` currently fails in this environment because `timeout` is not available on macOS by default.

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -68,6 +68,11 @@ When skills reference tools you don't have, substitute OpenCode equivalents:
 - \`Skill\` tool → OpenCode's native \`skill\` tool
 - \`Read\`, \`Write\`, \`Edit\`, \`Bash\` → Your native tools
 
+**Skill Name Compatibility (OpenCode):**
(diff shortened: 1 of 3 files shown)


DevNewbie1826:
Closing this PR per author request; implementation planning will happen first.

PR #869, opened 2026-03-21
fix: add safe bare-skill fallback for native OpenCode skill loading

DevNewbie1826:
## Summary
- Add a safe bare-skill fallback for the native OpenCode `skill` tool.
- Preserve native direction (no reintroduction of custom `use_skill`/`find_skills` flow).
- Rewrite only unambiguous bare names to `superpowers/<name>`.

## Problem
When users entered bare skill names such as `brainstorming`, they could hit:
`Skill or command "brainstorming" not found. Did you mean: superpowers/brainstorming, /superpowers:brainstorming?`
This created avoidable friction because the first attempt often failed.

## Root Cause
- Native `skill` resolution in OpenCode expects namespaced identifiers for superpowers skills.
- There was no compatibility fallback for bare names.

## Changes
- Hooked `tool.execute.before` in the superpowers plugin and only intercept the native `skill` tool call.
- Added bare-name resolver with strict guards:
  1. Keep explicit namespaced/path input unchanged.
  2. Keep project/user exact matches unchanged (preserve priority).
  3. Rewrite to `superpowers/<name>` only when that superpowers skill is known.
- Added tests and runner integration for alias fallback behavior.

## Files
- `.opencode/plugins/superpowers.js`
- `.opencode/skill-alias.js`
- `tests/opencode/test-skill-alias-unit.mjs`
- `tests/opencode/test-skill-alias.sh`
- `tests/opencode/run-tests.sh`

## Verification
- `node tests/opencode/test-skill-alias-unit.mjs` ✅
- `bash tests/opencode/run-tests.sh --test test-skill-alias.sh` ✅
- Runtime check:
  - `opencode run "Use the skill tool to load brainstorming ..."` -> `Skill "superpowers/brainstorming"` ✅
  - Unknown bare skill remains not found (no over-rewrite) ✅

## Risk / Compatibility
- Scoped to native `skill` tool only.
- No behavior change for explicit namespaced skills.
- Unknown/ambiguous names are not force-rewritten.

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -9,6 +9,7 @@ import path from 'path';
 import fs from 'fs';
 import os from 'os';
 import { fileURLToPath } from 'url';
+import { rewriteBareSkillName } from '../skill-alias.js';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
@@ -51,6 +52,8 @@ export const SuperpowersPlugin = async ({ client, directory }) => {
   const superpowersSkillsDir = path.resolve(__dirname, '../../skills');
   const envConfigDir = normalizePath(process.env.OPENCODE_CONFIG_DIR, homeDir);
   const configDir = envConfigDir || path.join(homeDir, '.config/opencode');
+  const userSkillsDir = path.join(configDir, 'skills');
+  const projectSkillsDir = path.join(directory, '.opencode/skills');
 
   // Helper to generate bootstrap content
   const getBootstrapContent = () => {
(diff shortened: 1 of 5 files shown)


obra:
Thanks for investigating this, @DevNewbie1826. We tested on OpenCode v1.3.2 and bare skill names (e.g., `brainstorming` without the `superpowers/` prefix) work correctly with the native skill tool. The underlying issue was a plugin system timing bug in OpenCode v1.1.x where `config.skills.paths` mutations from plugin config hooks weren't propagating to the skill registry. This was fixed upstream in OpenCode v1.3.0 (PR anomalyco/opencode#18280). Closing since the fix is upstream rather than in superpowers.

PR #878, opened 2026-03-22
feat(brainstorming): add Option Zero check before proposing solutions

### LINKED ISSUE #834: Suggestion: brainstorming skill — add 'Option Zero' forcing question before proposing solutions
## Context

During an investigation into a third-party SDK deprecation, the brainstorming/design phase jumped straight to "replace the SDK" and "migrate to a new protocol" when the actual fix was a one-line env var change. The root cause: I explored complex solutions without first testing the simplest hypothesis.

## Suggestion for `brainstorming` skill

Add a forcing question early in the brainstorming phase:

**"Is there an Option Zero?"**
Before exploring implementation options, ask:
- Can this be fixed with just a config/env var change?
- Can this be fixed by updating a dependency version?
- Can this be fixed without touching application code at all?

If Option Zero exists, it should be the default recommendation. Complex solutions should only be proposed when Option Zero has been ruled out with evidence.

## Why this matters

The brainstorming skill is great at exploring design space — but exploring too eagerly can skip past the simplest answer. "Option Zero" acts as a gravity check: if the fix doesn't require a code change, no amount of architectural exploration is better than changing the config.

This applies especially when investigating:
- Third-party SDK/integration issues (often config, not code)
- Infrastructure/deployment problems (often env vars, not application logic)
- Version deprecations (often a URL or header change, not a migration)

xr843:
Closes #834

## Summary

Adds an "Option Zero" forcing question to the brainstorming skill that gates the process before solution options are generated. Before exploring implementation approaches, the skill now checks whether the problem can be resolved without application code changes — via config/env vars, dependency updates, or tooling changes.

## Changes to `skills/brainstorming/SKILL.md`

- **Checklist**: Added step 2 "Option Zero check" between exploring project context and offering the visual companion (renumbered subsequent steps)
- **Process Flow diagram**: Added `Option Zero check` diamond node after `Explore project context`, with a branch to `Recommend Option Zero` (if resolved without code) or continuing to the normal flow
- **New "Option Zero check" process section**: Inserted between "Understanding the idea" and "Exploring approaches" with specific verification questions (config/env, dependency version, no-code fix) and guidance on when the check is most relevant (third-party SDK issues, infrastructure problems, version deprecations)

## Design rationale

As discussed in #834, "prefer simple solutions" is a ranking heuristic applied *after* options are generated. Option Zero is a **gate before options exist** — it asks whether the problem requires a code-level solution at all. This catches the recurring pattern where brainstorming jumps to implementation options for problems that are actually config/env/dependency issues.

The check is lightweight (adds ~15 lines to the process section) and self-skipping: genuinely complex design problems naturally fail the Option Zero check and proceed to brainstorming as usual.

### DIFF
--- a/skills/brainstorming/SKILL.md
+++ b/skills/brainstorming/SKILL.md
@@ -22,20 +22,22 @@ Every project goes through this process. A todo list, a single-function utility,
 You MUST create a task for each of these items and complete them in order:
 
 1. **Explore project context** — check files, docs, recent commits
-2. **Offer visual companion** (if topic will involve visual questions) — this is its own message, not combined with a clarifying question. See the Visual Companion section below.
-3. **Ask clarifying questions** — one at a time, understand purpose/constraints/success criteria
-4. **Propose 2-3 approaches** — with trade-offs and your recommendation
-5. **Present design** — in sections scaled to their complexity, get user approval after each section
-6. **Write design doc** — save to `docs/superpowers/specs/YYYY-MM-DD-<topic>-design.md` and commit
-7. **Spec review loop** — dispatch spec-document-reviewer subagent with precisely crafted review context (never your session history); fix issues and re-dispatch until approved (max 3 iterations, then surface to human)
-8. **User reviews written spec** — ask user to review the spec file before proceeding
-9. **Transition to implementation** — invoke writing-plans skill to create implementation plan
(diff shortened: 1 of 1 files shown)


xr843:
Closing — feel free to reopen if this is still wanted.

PR #880, opened 2026-03-22
feat(skills): add project memory workflow

### LINKED ISSUE #551: Feature Request: Add a core project memory system for cross-session retrieval and recording

The current Superpowers workflow has no general-purpose way for agents to recover small, relevant pieces of prior project history before making new design or implementation decisions, and no corresponding workflow for recording durable project memory after a change unit is complete.

In longer-running projects this creates two related failures:

- new sessions repeat work because they do not know what was tried, decided, or ruled out before
- finished work leaves behind commits and code, but not the durable reasoning future sessions need in order to stay consistent

Superpowers has strong process skills for planning, debugging, verification, and review, but no equivalent core system for carrying durable project memory across sessions.

## Proposed solution

Superpowers needs a general-purpose memory system with these properties:

- agents can retrieve relevant prior project history before making new decisions
- agents can record durable memory after a closed, reviewable change unit is complete
- the system keeps context use small instead of turning every session into a large history replay
- the system preserves reusable reasoning, not just commits or long-form notes

The exact implementation can vary, but the core feature should include:

- a repository-owned place for durable project memory
- a way to discover relevant memory incrementally
- a way to keep retrieval and recording disciplined
- a way to preserve reusable historical context across sessions

### 1. Treat spec and plan documents as the memory system

Specs and plans already preserve some reasoning, so I considered relying on them directly instead of introducing a dedicated memory system.

I rejected this because specs and plans are scoped to a specific design or implementation effort. They are not optimized for small, repeatable historical lookup across later sessions, and they do not give agents a clean way to record implementation outcomes after work is complete.

### 2. Treat git commits as the memory system

I also considered using commit history as the primary memory layer.

I rejected this because commit messages are intentionally short, uneven in detail, and optimized for version control rather than cross-session retrieval. They are useful evidence, but not a replacement for a structured workflow that helps agents discover relevant prior decisions and preserve reusable guidance.

### 3. Treat `AGENTS.md`, `CLAUDE.md`, or similar markdown files as the memory system

I also considered putting historical guidance directly into always-loaded markdown files such as `AGENTS.md` or `CLAUDE.md`.

I rejected this because it would mix durable history with always-present instructions, causing context rot. Over time those files would accumulate large amounts of project history that are not relevant to the current task, making the active context noisier and less reliable.

Yes.

This is not specific to one project, domain, or third-party tool. Many Superpowers users work across multiple sessions, iterate on designs, revisit failed paths, and need later agents to understand prior decisions without dragging large amounts of context into every conversation.

The proposed memory system is general-purpose infrastructure for agentic development, in the same sense that planning, debugging, verification, and code review are general-purpose workflow infrastructure.

LanternCX:

Superpowers currently has no core workflow for preserving and reusing durable project memory across sessions.

Closes #551 

That shows up in two ways:

- new sessions do not have a structured way to retrieve small, relevant pieces of prior project history before continuing design or implementation work
- completed work leaves behind commits and code, but not a durable memory workflow for preserving why a decision mattered or what future sessions should keep consistent

Superpowers needs a general-purpose memory system. This PR is one concrete solution to that problem.

## What does this PR change?

This PR introduces one concrete memory workflow in core Superpowers:

- adds `using-memory` for incremental retrieval of relevant project history
- adds `record-memory` for recording closed, reviewable change units as durable memory
- updates README and verification integration to reflect the new memory workflow

It also replaces the older `memory-tracker` naming and updates the shipped memory type templates so retrieval and recording use separate routing fields.

The core design is:

- one skill for retrieval (`using-memory`)
- one skill for recording (`record-memory`)
- one repository-owned memory layout under `docs/superpowers/memory/`

The resulting directory structure is:

```text
docs/superpowers/memory/
  <type>/
    MEMORY.md
    INDEX.md
    ENTRY.md
    entries/
      YYYY-MM/
        YYYY-MM-DD-N.md
```

This PR also includes the context-engineering rules needed to keep retrieval cheap and predictable:

- `MEMORY.md` is the type-definition layer
- `INDEX.md` is an index-only lookup layer
- entry files are evidence, not discovery metadata
- retrieval follows progressive disclosure instead of broad reads

In practice, `using-memory` is designed to:

1. scan `docs/superpowers/memory/*/MEMORY.md`
2. `grep` only `Use When` / `Avoid When`
3. read only small windows around those fields
4. choose candidate types before touching `INDEX.md`
5. inspect only the smallest useful number of entries
6. stop once the task has enough history to move forward

`record-memory` follows the complementary write path:

1. confirm the work is one closed, reviewable change unit
2. scan type-local `MEMORY.md` files
3. route with `Record When` / `Avoid Recording When`
4. write a single durable entry and update only that type's `INDEX.md`

Yes.

This is general workflow infrastructure for multi-session agentic development. It is not tied to one domain, one repository layout outside Superpowers' own docs area, or any third-party tool or service. The same need appears in feature work, debugging, refactoring, and long-running design iteration.

I considered three broader alternatives before landing on this implementation:

1. using spec and plan documents as the memory system
2. using git commits as the memory system
3. relying on always-loaded markdown files such as `AGENTS.md` or `CLAUDE.md` instead of introducing a dedicated workflow

I rejected them because they either make retrieval too expensive, fail to preserve reusable project history in a disciplined form, or cause context rot by mixing historical memory into always-loaded instruction files. This PR's memory workflow is intended to complement existing artifacts, not replace them.

No.

The changes are related parts of one memory-workflow refactor:

- retrieval skill
- recording skill
- type routing/template updates needed to support both
- minimal README / verification updates needed to integrate them coherently

## Existing PRs
- Related PRs: #555 

`#555` explored the same general problem space. This PR narrows the solution to a memory workflow split between retrieval and recording.

## Environment tested

| OpenCode | current local CLI | gpt-5.4 | openai/gpt-5.4 |

## Evaluation
- Initial prompt: design a memory-oriented workflow that supports both recording and retrieval while keeping context use small
- Evaluation so far: I have tried parts of this workflow in a few of my own projects, but not with broad or deep usage yet
- Before: the proposal centered on a single progress-style skill and did not clearly teach agents how to retrieve prior memory with minimal context
- After: the workflow is split into retrieval and recording, retrieval now follows a clear incremental path, and the memory type protocol distinguishes lookup routing from recording routing

This PR is intended as a draft. I have tried this workflow in a few of my own projects, but not with broad or deep enough usage to treat the current shape as fully proven.

## Rigor

- [ ] This change was tested adversarially, not just on the happy path

Adversarial / review results used during development:

- checked whether retrieval and recording were being incorrectly injected at the same lifecycle point
- checked whether `using-memory` could be discovered with a minimal description instead of broader workflow injection
- checked whether retrieval was truly incremental instead of degenerating into whole-file reads
- checked whether default memory types were hardcoded into runtime routing instead of discovered from type-local files
- checked whether the new skills matched repository naming and writing style closely enough to feel native

## Human review

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,10 @@ Start a new session in your chosen platform and ask for something that should tr
 
 7. **finishing-a-development-branch** - Activates when tasks complete. Verifies tests, presents options (merge/PR/keep/discard), cleans up worktree.
 
+8. **using-memory** - Retrieves only the smallest useful slice of project history, using `Use When` / `Avoid When` to choose which memory types to inspect before reading indexes or entries.
+
+9. **record-memory** - Records closed reviewable change units in `docs/superpowers/memory/`, using `Record When` / `Avoid Recording When` to choose the right memory type and repairing missing structure only when needed.
+
 **The agent checks for relevant skills before any task.** Mandatory workflows, not suggestions.
 
 ## What's Inside
@@ -137,6 +141,10 @@ Start a new session in your chosen platform and ask for something that should tr
 - **using-git-worktrees** - Parallel development branches
 - **finishing-a-development-branch** - Merge/PR decision workflow
 - **subagent-driven-development** - Fast iteration with two-stage review (spec compliance, then code quality)
+
+**Project Memory**
+- **record-memory** - Records closed reviewable change units in `docs/superpowers/memory/`, routing with `Record When` / `Avoid Recording When` and repairing structure only when needed
+- **using-memory** - Retrieves the smallest useful slice of project history, routing with `Use When` / `Avoid When` before reading `INDEX.md` or entries
 
 **Meta**
 - **writing-skills** - Create new skills following best practices (includes testing methodology)
--- a/skills/record-memory/SKILL.md
+++ b/skills/record-memory/SKILL.md
@@ -1,0 +1,58 @@
+---
+name: record-memory
+description: Use when a closed, reviewable change unit should be recorded in docs/superpowers/memory/
+---
+
+# Record Memory
+
+## Overview
+
+Record durable project memory for a completed change unit.
+
+## Entry Conditions
+
+Create a memory entry only when all of the following are true:
+
+1. The work has one clear goal.
+2. The work forms one coherent, reviewable change unit.
+3. The work is complete enough to preserve as durable project memory.
+
+Do not use memory entries as work logs for exploration, partial work, or scattered trivial changes.
+
+## Discovery Order
+
+Every entry belongs to exactly one type.
+
+If discovery cannot identify one unique type, stop and ask the user to clarify.
+
+Discover the target type in this order:
+
+1. Scan `docs/superpowers/memory/*/MEMORY.md`.
+2. Read each candidate header's `Type`, `Record When`, `Avoid Recording When`, and `Entry Template`.
+3. Choose the best-matching type for the completed change unit.
+4. After selecting the type, read only that type's `entry.md` before writing.
+5. Write the entry under `docs/superpowers/memory/<type>/entries/<YYYY-MM>/`.
+6. Update only that type's `INDEX.md` at `docs/superpowers/memory/<type>/INDEX.md`.
+
+Do not hardcode type names or type meaning from this skill alone. Repository memory files are the source of truth.
+This skill only routes recording work. It never retrieves, summarizes, or pre-reads historical memory for the current task.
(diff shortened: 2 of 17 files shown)


obra:
Thanks so much for submitting this. I don't think that a memory system is a great fit for superpowers. I believe strongly that memory systems are important, but enough people have enough divergent opinions about them that I'm not going to force mine onto every superpowers user.

PR #899, opened 2026-03-23
fix(subagent-prompts): instruct subagents to read CLAUDE.md for project conventions

arittr:
## Summary

- Adds a one-liner to each of the three subagent dispatch templates instructing subagents to read CLAUDE.md (or AGENTS.md) before starting work
- **implementer-prompt.md**: "Read CLAUDE.md... its conventions are binding on your work" in the Before You Begin section
- **spec-reviewer-prompt.md**: "verify the implementation follows its conventions" in the Your Job section  
- **code-quality-reviewer-prompt.md**: "use its conventions as review criteria" before the review checklist

## Why

Subagents spawned via the Agent tool start with a fresh context — no CLAUDE.md pre-loaded. They pattern-match existing code well, but miss cross-cutting conventions only documented in CLAUDE.md (test location, DB mocking policy, architectural boundaries, error format conventions).

RED/GREEN tested across multiple scenarios including legacy codebases with conflicting patterns. Without the fix, subagents copy legacy anti-patterns (if/else, try/catch, console.log, argument mutation). With the fix, subagents read CLAUDE.md first and follow its conventions over conflicting existing code.

## Test plan

- [x] RED tested: subagents without fix copy legacy patterns from existing code
- [x] GREEN tested: subagents with fix read CLAUDE.md and follow its conventions over conflicting code
- [x] Verified fix works with empty projects (CLAUDE.md only, no existing code)
- [x] Verified fix works with conflicting legacy code (no disclaimer comments)
- [ ] Manual verification in a real SDD workflow

Closes #793
Related: #804

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #900, opened 2026-03-24
docs(codex-tools): add named agent dispatch mapping for Codex

### LINKED ISSUE #647: Codex: requesting-code-review assumes Task/subagent dispatch that native skill install does not provide
## Summary

In Codex, the `requesting-code-review` workflow still assumes a `Task`/subagent dispatch path that the current native-skill installation does not provide.

The result is that Codex reaches a required review step, then emits a fallback message like:

> I can’t dispatch the requesting-code-review subagent here because this environment doesn’t expose the Task/subagent tool it expects.

and continues with a degraded branch-finishing flow instead of performing the intended review.

## Why this looks like a Superpowers bug

The current Codex install/docs path only exposes `skills/` via native skill discovery:

- `.codex/INSTALL.md`
- `docs/README.codex.md`

Those instructions symlink `~/.codex/superpowers/skills` into `~/.agents/skills/superpowers`, but they do not install `agents/code-reviewer.md` or describe any Codex-compatible substitute for it.

At the same time, `skills/requesting-code-review/SKILL.md` still says:

- "Dispatch superpowers:code-reviewer subagent..."
- "Use Task tool with superpowers:code-reviewer type..."

So the Codex-native installation path appears inconsistent with the skill's runtime assumptions.

## Reproduction

1. Install Superpowers in Codex using the current native skill instructions from `.codex/INSTALL.md` / `docs/README.codex.md`.
2. Complete implementation work in Codex and invoke a workflow that reaches `requesting-code-review` before merge/finalization.
3. Observe that Codex cannot execute the requested review step because there is no `Task`/subagent tool available in this environment.
4. The assistant falls back to a manual merge/finish flow and skips the intended reviewer dispatch.

## Expected

One of these should happen in Codex:

1. `requesting-code-review` has an explicit Codex fallback that runs the review inline in the current session using `skills/requesting-code-review/code-reviewer.md` as the checklist/template.
2. The Codex install/docs also install and wire up whatever agent/review mechanism the skill expects.
3. The skill clearly marks this path as unsupported in Codex and routes to an alternate verification/review workflow instead of failing at runtime.

## Actual

The Codex-native install path makes the skill discoverable, but the skill still requires a dispatch mechanism that Codex does not provide, so the review step silently degrades into "can't do that here" behavior.

## Extra context

This also seems to conflict with the Codex direction mentioned in `RELEASE-NOTES.md` for Codex support:

- "Tool mapping for Codex: TodoWrite→update_plan, subagents→manual fallback"
- "Simplified subagent handling (manual work instead of delegation)"

`requesting-code-review` does not appear to have received that Codex fallback treatment.

Related:

- #193
- #589

arittr:
## Summary

- Adds a "Named agent dispatch" section to `codex-tools.md` explaining how Codex should handle skills that reference named agent types (e.g., `superpowers:code-reviewer`)
- Maps `Task tool (superpowers:code-reviewer)` → `spawn_agent(agent_type="worker", message=...)` with the agent's prompt content inlined
- Includes message framing guidance (task-delegation framing, XML tags, execution directive) since Codex's `message` parameter is user-level input, not a system prompt
- Documents when this workaround can be removed (`RawPluginManifest` gains an `agents` field)

## Why

Codex's `spawn_agent` requires pre-registered TOML agent roles — it can't dispatch named agent types defined by `agents/*.md` files. Skills like `requesting-code-review` reference `superpowers:code-reviewer` which doesn't exist in Codex. The `worker` built-in + inlined prompt approach avoids duplicating agent definitions across formats.

## Test plan

- [ ] Manual verification: Codex user invokes requesting-code-review, review runs as worker subagent
- [ ] Verify Claude Code path is unaffected (no skill files changed)

Closes #647

### DIFF
--- a/skills/using-superpowers/references/codex-tools.md
+++ b/skills/using-superpowers/references/codex-tools.md
@@ -4,7 +4,7 @@ Skills use Claude Code tool names. When you encounter these in a skill, use your
 
 | Skill references | Codex equivalent |
 |-----------------|------------------|
-| `Task` tool (dispatch subagent) | `spawn_agent` |
+| `Task` tool (dispatch subagent) | `spawn_agent` (see [Named agent dispatch](#named-agent-dispatch)) |
 | Multiple `Task` calls (parallel) | Multiple `spawn_agent` calls |
 | Task returns result | `wait` |
 | Task completes automatically | `close_agent` to free slot |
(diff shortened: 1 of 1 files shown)


arittr:
This change is part of the following stack:

- #900 ◀
    - #899
        - #902
            - #901

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

obra (approved):

PR #909, opened 2026-03-24
fix: sync package.json version and remove external skill reference

chris-yyau:
## Summary

- **Version sync:** `package.json` says `5.0.4` but `.claude-plugin/plugin.json` says `5.0.5`. Updated `package.json` to match.
- **External skill reference:** `skills/brainstorming/SKILL.md` line 116 references `elements-of-style:writing-clearly-and-concisely`, a skill that doesn't exist in this project. Replaced with inline writing guidance that conveys the same intent without depending on an external skill.

## Test plan

- [ ] Verify `package.json` version matches `plugin.json`
- [ ] Verify brainstorming skill still reads naturally at line 116
- [ ] No behavioral changes — documentation/metadata only

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "superpowers",
-  "version": "5.0.4",
+  "version": "5.0.5",
   "type": "module",
   "main": ".opencode/plugins/superpowers.js"
 }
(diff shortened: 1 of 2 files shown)


obra:
This conflated two things, one of which we took care of, and the other of which...is very different than the one liner that you threw in instead. (Also, this PR ignored the PR template)

PR #722, opened 2026-03-13
feat: add native GitHub Copilot CLI support

alfantp:
## Summary

Adds native GitHub Copilot CLI integration, making all superpowers skills available with one-command installation.

Closes #217
Related: #358

### Installation

```bash
copilot plugin install obra/superpowers
```

### What's included

- **Root `plugin.json`** — Copilot CLI discovers this first (before `.claude-plugin/plugin.json`), providing clean platform separation
- **`.copilot/hooks/`** — Copilot CLI-format sessionStart hook (bash + PowerShell) for update checks and legacy warnings
- **`copilot-tools.md`** — Tool mapping reference (following `codex-tools.md` pattern)
- **`.copilot/INSTALL.md`** — Platform-specific installation instructions
- **`docs/README.copilot.md`** — Comprehensive platform guide with all 14 skills documented
- **`README.md`** — Updated installation section
- **`SKILL.md`** — One-line addition to platform adaptation references

### Design decisions

1. **Root plugin.json**: Copilot CLI checks root → `.github/plugin/` → `.claude-plugin/` in order. A root manifest is found first, so Claude Code continues using `.claude-plugin/` undisturbed.

2. **No bootstrap injection needed**: Unlike Claude Code, Copilot CLI natively discovers plugin skills and shows their descriptions to the agent. The `using-superpowers` skill is discovered automatically and enforces skill usage discipline.

3. **Minimal skill changes**: Only one line changed in one SKILL.md file. All other skills work as-is.

4. **Security hardening**: Hook scripts include `GIT_TERMINAL_PROMPT=0` to prevent credential prompts from hanging sessions, quoted paths for Windows compatibility, and proper error fallbacks.

5. **Verified against actual CLI**: All docs use terminal command form (`copilot plugin install/update/uninstall/list`), tool mappings verified against `copilot help commands`, and skill terminology aligned with Copilot CLI's `/skills` system.

### Zero impact on existing platforms

- `.claude-plugin/`, `.cursor-plugin/`, `.codex/`, `.opencode/` — untouched
- `hooks/hooks.json` (Claude Code format) — untouched
- All 14 skill files — untouched (except one-line platform reference addition in `using-superpowers`)
- `lib/skills-core.js` — untouched

### Testing (Copilot CLI v1.0.5, macOS arm64)

```
$ copilot plugin install ./
Plugin "superpowers" installed successfully. Installed 14 skills.

$ copilot plugin list
Installed plugins:
  • superpowers (v5.0.2)

[...]

PR #533, opened 2026-02-23
feat: add GitHub Copilot CLI support

mhenke:
## Summary

Adds native GitHub Copilot CLI support to the superpowers project. Addresses #217.

### Changes
- .copilot/plugin.json — native plugin manifest for copilot plugin add support
- .copilot/hooks/ — Copilot CLI native hooks (version 1 format, sessionStart)
- .copilot/install.sh — manual installer with hub-pattern skill symlinks + context injection
- .copilot/COPILOT.md — activation instructions
- .copilot/INSTALL.md — full installation guide (plugin + manual)
- docs/README.copilot.md — detailed usage documentation
- README.md — added Copilot CLI section with plugin install as primary method

### Install Methods
1. **Plugin (recommended):** copilot plugin add https://github.com/obra/superpowers
2. **Manual:** git clone + install.sh for symlink-based setup

### Bug Fixes
- Fixed agent path bug in install.sh (was pointing to empty .copilot/agents/ instead of project-root agents/)
- Removed .toml command files (upstream already has .md versions)
- Removed unverified COPILOT_SKILLS_DIRS claim from docs

PR #910, opened 2026-03-24
fix: add Copilot CLI platform detection for sessionStart context injection

### LINKED ISSUE #792: Copilot CLI: no way to inject superpowers bootstrap at startup from a plugin
## Summary

Copilot CLI v1.0.7 loads superpowers skills correctly from a plugin directory, but there is no working mechanism for a plugin to inject the "always check skills first" bootstrap context at session start.

## What works

- **Skills load perfectly** — all 19 superpowers skills are discovered and shown in `/skills`
- **Skill invocation works** — `skill(brainstorming)`, `skill(using-superpowers)` etc. all work
- **Agent uses skills proactively** when descriptions match (e.g., brainstorming auto-fires for "build a todo app")
- **`.claude-plugin/plugin.json`** is discovered by Copilot CLI (checked from `.plugin/`, `.`, `.github/plugin/`, `.claude-plugin/`)

### 1. `type: "prompt"` hooks don't fire from plugins

Copilot CLI supports `type: "prompt"` entries in `sessionStart` hooks that auto-submit a prompt at session start. However, startup prompts are only extracted from **project-level** (`.github/hooks/`) and **user-level** (`~/.copilot/hooks/`) hook configs — not from plugin hooks.

In the source (`app.js` v1.0.7), `_startupPrompts` is populated from user and project hooks dirs only:
```javascript
let {hooksDir: a, configs: l, userHooksDir: c, userConfigs: I} = await this.loadAllHooks(e, o);
let d = await xBt(c, q, I);  // user hooks
let u = await xBt(a, q, l);  // project hooks  
this._startupPrompts = [...d, ...u];  // no plugin hooks
```

### 2. `sessionStart` command hooks discard output

Even if a plugin's `sessionStart` command hook returns JSON with `additionalContext`, the output handler is a no-op:

```javascript
// sessionStart output handler:
l => {}  // discarded

// Compare to subagentStart which DOES extract additionalContext:
l => l?.additionalContext !== void 0 ? { additionalContext: l.additionalContext } : void 0
```

The SDK types define `SessionStartHookOutput.additionalContext` but the CLI doesn't read it from command hooks.

### 3. `AGENTS.md` not loaded from plugin directories

Copilot loads `AGENTS.md` as custom instructions from the **working directory** only, not from plugin directories. So a plugin can't use `AGENTS.md` to inject bootstrap instructions.

### 4. `copilot-instructions.md` not loaded from plugin directories

Same — only loaded from the working directory's `.github/` and `~/.copilot/`.

## What we need

Any ONE of these would unblock superpowers on Copilot CLI:

1. **Extract `type: "prompt"` entries from plugin hooks** (alongside project/user hooks)
2. **Process `additionalContext` from `sessionStart` command hook output** (like `subagentStart` already does)
3. **Load `AGENTS.md` or custom instruction files from plugin directories**

Option 2 would give parity with how Claude Code's `SessionStart` hook works (plugin returns JSON with context that gets injected into the session).

## Environment

- Copilot CLI: 1.0.7
- Platform: macOS (Darwin 25.2.0)
- Plugin: superpowers v5.0.5

## Workaround

Skills work without the bootstrap — agents invoke them when descriptions match. The gap is the enforcement: without the bootstrap, the agent doesn't know it MUST check skills before every action.

culinablaz:
# fix: add Copilot CLI platform detection for sessionStart context injection

Copilot CLI v1.0.11 now reads `additionalContext` from sessionStart hook output, but the `session-start` script only emits the Claude Code-specific nested format (`hookSpecificOutput.additionalContext`). Copilot CLI never receives the bootstrap context.

Fixes #792

## What does this PR change?

Adds `COPILOT_CLI` env var detection to `hooks/session-start` so Copilot CLI gets top-level `additionalContext` (SDK standard) while Claude Code continues getting `hookSpecificOutput`.

Yes — platform detection infrastructure that makes the existing bootstrap work on Copilot CLI.

1. **Emit both fields in one JSON** — Rejected. Claude Code reads both `additional_context` and `hookSpecificOutput` without deduplication (bug fixed in 74f2b1c), so adding a top-level field risks double injection.
2. **Emit only top-level `additionalContext`** — Would break Claude Code.

No.

## Existing PRs

- Related PRs: #660 (duplication fix, committed as 74f2b1c), #722 and #533 (Copilot CLI support — closed, too broad)

## Environment tested

| Harness     | Version | Model           | Model version/ID |
|-------------|---------|-----------------|------------------|
| Copilot CLI | 1.0.11  | Claude Opus 4.6 | claude-opus-4.6  |

## Evaluation

- Tested all 4 platform branches (Cursor, Copilot CLI, Claude Code, fallback) with assertion-based JSON validation
- Live-tested on Copilot CLI 1.0.11: agent auto-invoked `skill(using-superpowers)` and `skill(brainstorming)` at session start
- Confirmed `COPILOT_CLI=1` is set by Copilot CLI runtime

## Rigor

- [x] This change was only tested on Copilot CLI v1.0.11 — Claude Code and Cursor were not available for testing
- [x] I did not modify carefully-tuned content without extensive evals

## Human review

---

**Question for @obra**: Is detecting `COPILOT_CLI` env var the right approach here, or would you prefer a different mechanism? Happy to adjust if there's a better way to distinguish Copilot CLI from Claude Code (both set `CLAUDE_PLUGIN_ROOT`).

### DIFF
--- a/hooks/session-start
+++ b/hooks/session-start
@@ -35,23 +35,23 @@ warning_escaped=$(escape_for_json "$warning_message")
 session_context="<EXTREMELY_IMPORTANT>\nYou have superpowers.\n\n**Below is the full content of your 'superpowers:using-superpowers' skill - your introduction to using skills. For all other skills, use the 'Skill' tool:**\n\n${using_superpowers_escaped}\n\n${warning_escaped}\n</EXTREMELY_IMPORTANT>"
 
 # Output context injection as JSON.
-# Cursor hooks expect additional_context.
-# Claude Code hooks expect hookSpecificOutput.additionalContext.
-# Claude Code reads BOTH fields without deduplication, so we must only
-# emit the field consumed by the current platform to avoid double injection.
+# Cursor hooks expect additional_context (snake_case).
+# Claude Code hooks expect hookSpecificOutput.additionalContext (nested).
+# Copilot CLI and others expect additionalContext (top-level, SDK standard).
+# Claude Code reads BOTH additional_context and hookSpecificOutput without
+# deduplication, so we must emit only the field the current platform consumes.
 #
-# Uses printf instead of heredoc (cat <<EOF) to work around a bash 5.3+
-# bug where heredoc variable expansion hangs when content exceeds ~512 bytes.
+# Uses printf instead of heredoc to work around bash 5.3+ heredoc hang.
 # See: https://github.com/obra/superpowers/issues/571
 if [ -n "${CURSOR_PLUGIN_ROOT:-}" ]; then
-  # Cursor sets CURSOR_PLUGIN_ROOT (may also set CLAUDE_PLUGIN_ROOT) — emit additional_context
+  # Cursor sets CURSOR_PLUGIN_ROOT (may also set CLAUDE_PLUGIN_ROOT)
   printf '{\n  "additional_context": "%s"\n}\n' "$session_context"
-elif [ -n "${CLAUDE_PLUGIN_ROOT:-}" ]; then
-  # Claude Code sets CLAUDE_PLUGIN_ROOT — emit only hookSpecificOutput
+elif [ -n "${CLAUDE_PLUGIN_ROOT:-}" ] && [ -z "${COPILOT_CLI:-}" ]; then
+  # Claude Code sets CLAUDE_PLUGIN_ROOT without COPILOT_CLI
   printf '{\n  "hookSpecificOutput": {\n    "hookEventName": "SessionStart",\n    "additionalContext": "%s"\n  }\n}\n' "$session_context"
 else
-  # Other platforms — emit additional_context as fallback
-  printf '{\n  "additional_context": "%s"\n}\n' "$session_context"
+  # Copilot CLI (sets COPILOT_CLI) or unknown platform — SDK standard format
+  printf '{\n  "additionalContext": "%s"\n}\n' "$session_context"
 fi
 
 exit 0


obra:
Thanks for the PR! I've been digging into this and wanted to ask about the core claim.

I checked the Copilot CLI v1.0.11 source (`app.js`) and the `sessionStart` command hook output handler still appears to be a no-op:

```javascript
// In kBt (hook builder), sessionStart output handler:
l => {}

// Compare to subagentStart, which actually processes output:
l => l?.additionalContext !== void 0 ? { additionalContext: l.additionalContext } : void 0
```

This means even with the correct JSON format, the output from a sessionStart command hook is discarded before it reaches the agent context.

Can you share more about how you verified this working? When you say "agent auto-invoked `skill(using-superpowers)` and `skill(brainstorming)` at session start" — could that have been Copilot CLI's native skill description injection (it adds skill descriptions to the system prompt via `addEntries`) rather than the sessionStart hook's `additionalContext` being processed?

Trying to make sure we're not confusing two different mechanisms here.

obra:
**Correction — please disregard my previous comment.** I was reading the stale npm-installed `app.js` (v1.0.7) instead of the auto-updated binary at `~/.copilot/pkg/universal/1.0.11/app.js`.

In **v1.0.11**, the sessionStart output handler was indeed fixed:

```javascript
// v1.0.7 (broken):
l => {}

// v1.0.11 (fixed):
l => l?.additionalContext === void 0
  ? void 0
  : { additionalContext: typeof l.additionalContext == "string"
      ? l.additionalContext
      : JSON.stringify(l.additionalContext) }
```

The changelog confirms: `[fixed] sessionStart hook additionalContext is now injected into the conversation`

So the approach in this PR is correct — emitting top-level `{ "additionalContext": "..." }` will work on v1.0.11+. Sorry for the noise.

obra:
Thanks for this work, @culinablaz! Your fix was correct — Copilot CLI v1.0.11 does process `additionalContext` from sessionStart hook output. (My earlier comment about the no-op handler was wrong — I was reading the stale v1.0.7 npm package instead of the auto-updated v1.0.11 binary.)

I've merged your session-start hook change into the dev branch with your git authorship preserved, along with a Copilot CLI tool mapping table, skill/README updates, and a changelog entry crediting you. We verified the full integration end-to-end:

- `copilot plugin install` from marketplace works (14 skills discovered)
- SessionStart hook fires and `additionalContext` is injected (confirmed in session events)
- Agent proactively invokes skills (`brainstorming`, `systematic-debugging`, etc.)
- All existing platform paths (Claude Code, Cursor) unchanged

Thanks again for tracking down the `COPILOT_CLI` env var detection approach and testing it on v1.0.11!

syhe:
@obra Was closing instead of merging this intentional?

roysha1:
@syhe  It’s merged in dev ed06287a8a3429758b0c34719af0de7b9466d3f9

syhe:
I see, the push happened after my comment. :-)

PR #911, opened 2026-03-24
fix: sync version across gemini-extension.json and OpenCode docs

chris-yyau:
## Summary

- **gemini-extension.json** version is `5.0.0` but plugin.json and marketplace.json are at `5.0.5`. Updated to match.
- **OpenCode docs** (`.opencode/INSTALL.md` and `docs/README.opencode.md`) pin examples reference `v5.0.3` which is two releases behind. Updated to `v5.0.5`.

**Note:** `package.json` version sync (`5.0.4` → `5.0.5`) is handled separately in #909.

## Test plan

- [ ] Verify all version references are consistent across manifests
- [ ] No behavioral changes — metadata and documentation only

### DIFF
--- a/.opencode/INSTALL.md
+++ b/.opencode/INSTALL.md
@@ -52,7 +52,7 @@ To pin a specific version:
 
 ```json
 {
-  "plugin": ["superpowers@git+https://github.com/obra/superpowers.git#v5.0.3"]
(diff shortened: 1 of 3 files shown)


obra:
Thanks for this. We now have semi-automated tooling as part of the release engineering flow that should help prevent this in the future.

PR #926, opened 2026-03-24
Fix owner-PID lifecycle monitoring for cross-platform reliability

obra:
## Summary

- Fixes `ownerAlive()` treating EPERM (permission denied) as "process dead" — cross-user owner PIDs (Tailscale SSH, system daemons) caused false shutdowns within 60s
- Validates owner PID at server startup — if already dead (WSL scenario where grandparent is a short-lived subprocess), disables monitoring and relies on the 30-minute idle timeout
- Removes the Windows/MSYS2-specific `OWNER_PID=""` carve-out from `start-server.sh` since the server now handles invalid PIDs generically

## Test plan

Tested on Linux (magic-kingdom) via Tailscale SSH:

- [ ] Root-owned owner PID (EPERM): server survives past 60s lifecycle check
- [ ] Dead owner PID at startup (WSL simulation): monitoring disabled, server survives
- [ ] Valid owner that dies: server shuts down within 60s
- [ ] Full test suite passes (25/25)
- [ ] Needs testing on actual WSL environment

Fixes #879

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -17,7 +17,7 @@ The subagent review loop (dispatching a fresh agent to review plans/specs) doubl
 
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


arittr (approved):
LGTM

obra:
Merged directly to dev branch.

PR #927, opened 2026-03-24
Fix writing-skills to include all supported frontmatter fields

satanas:
## Summary

- Fixed incorrect claim that only `name` and `description` are supported frontmatter fields, which caused the skill to flag valid fields like `user-invocable` as unsupported
- Added all supported frontmatter fields from the [official documentation](https://code.claude.com/docs/en/skills#frontmatter-reference): `disable-model-invocation`, `user-invocable`, `allowed-tools`, `context`, `agent`, `model`, `effort`, `argument-hint`, and `hooks`
- Linked to the official frontmatter reference so the skill stays accurate as docs evolve

## Test plan

- [ ] Verify the skill no longer flags `user-invocable` as unsupported
- [ ] Confirm all listed fields match the official docs at https://code.claude.com/docs/en/skills#frontmatter-reference

### DIFF
--- a/skills/writing-skills/SKILL.md
+++ b/skills/writing-skills/SKILL.md
@@ -93,18 +93,28 @@ skills/
 ## SKILL.md Structure
 
 **Frontmatter (YAML):**
-- Only two fields supported: `name` and `description`
-- Max 1024 characters total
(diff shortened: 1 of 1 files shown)


obra:
This is the claude-code list of frontmatter fields, which is different than the agentskills standard.

Issue #833, opened 2026-03-19
Deprecated command aliases block skill invocation instead of forwarding


SilviuSavu:
## Problem

The three deprecated commands in `commands/` (`brainstorm.md`, `execute-plan.md`, `write-plan.md`) block the corresponding skills from being invoked. Instead of forwarding to the replacement skill, they instruct the model to tell the user the command is deprecated:

```
Tell your human partner that this command is deprecated and will be removed in the next major release.
They should ask you to use the "superpowers brainstorming" skill instead.
```

This means running `/superpowers:brainstorm` does **nothing useful** — it just nags the user instead of invoking `superpowers:brainstorming`. The model follows the command's instruction literally and never loads the skill.

## Affected commands

- `commands/brainstorm.md` — should invoke `superpowers:brainstorming`
- `commands/execute-plan.md` — should invoke `superpowers:executing-plans`
- `commands/write-plan.md` — should invoke `superpowers:writing-plans`

## Suggested fix

Either:

1. **Forward to the skill** — change the command body to `Invoke the superpowers:brainstorming skill now.` (etc.)
2. **Remove the commands entirely** — if users should only use skill names, don't register commands that intercept them

Option 1 is more user-friendly since people will naturally try the shorter name.

## Version

superpowers v5.0.5

obra:
Verified by Claude Code: fixed on the `dev` branch by commit f55f8df — `Remove the deprecated legacy slash commands (#1188)`. The deprecated command aliases (`brainstorm`, `execute-plan`, `write-plan`) have been removed entirely from `commands/`, eliminating the issue. Will ship with the next release.

— Claude Opus 4.7, Claude Code 2.1.112

PR #930, opened 2026-03-24
Fix deprecated commands to forward to replacement skills

noahrasheta:

Running `/superpowers:brainstorm`, `/superpowers:execute-plan`, or `/superpowers:write-plan` does nothing useful — the model follows the command's instruction to tell the user the command is deprecated, but never actually invokes the replacement skill. The user gets a nag message instead of the skill running. This is the exact behavior described in #833.

## What does this PR change?

Changes the three deprecated command files to transparently forward to their replacement skills (`brainstorming`, `executing-plans`, `writing-plans`) instead of blocking invocation with a deprecation notice.

Yes — this fixes a bug in the core command infrastructure that affects all users. The deprecated commands ship with superpowers and currently break skill invocation for anyone who uses the old command names.

1. **Remove the commands entirely** — This was option 2 in the issue. However, keeping them as forwarding aliases is more user-friendly since people will naturally try the shorter names. Removing them would make those names silently unavailable.
2. **Show deprecation notice AND forward** — Considered having the command inform the user about the rename before forwarding, but this adds friction and the description field already communicates the deprecation. The user's intent is to run the skill, not to learn about internal renames.

No. All three files are the same fix applied to the same pattern.

## Existing PRs
- Related PRs: none found addressing #833

## Environment tested

| Claude Code | 1.0.41 | Claude | claude-opus-4-6 |

## Evaluation

- The change was motivated by issue #833 which has 6 thumbs-up reactions
- Before: running `/superpowers:brainstorm` produces a deprecation message and the skill never loads
- After: the command transparently forwards to `superpowers:brainstorming` and the skill runs immediately
- The fix is intentionally minimal — one line changed per file, same pattern across all three

## Rigor

This is a command fix, not a skills change. The wording "Do not stop to inform them about the rename — just run the skill" is deliberately explicit to prevent the model from hedging or adding its own deprecation commentary.

## Human review

### DIFF
--- a/commands/brainstorm.md
+++ b/commands/brainstorm.md
@@ -2,4 +2,4 @@
 description: "Deprecated - use the superpowers:brainstorming skill instead"
 ---
 
-Tell your human partner that this command is deprecated and will be removed in the next major release. They should ask you to use the "superpowers brainstorming" skill instead.
+This command has been renamed. Invoke the superpowers:brainstorming skill now, passing along any arguments your human partner provided. Do not stop to inform them about the rename — just run the skill.
--- a/commands/execute-plan.md
+++ b/commands/execute-plan.md
@@ -2,4 +2,4 @@
 description: "Deprecated - use the superpowers:executing-plans skill instead"
 ---
 
-Tell your human partner that this command is deprecated and will be removed in the next major release. They should ask you to use the "superpowers executing-plans" skill instead.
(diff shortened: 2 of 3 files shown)


obra:
Please see the numerous closed issues and PRs related to this. It's gonna be another couple weeks. Also, this change completely misunderstands the whole point of deprecating those commands, which is informing the humans that they're going away.

PR #901, opened 2026-03-24
feat: Codex App worktree and sandbox compatibility

arittr:
## Summary

Makes superpowers skills work correctly in Codex App's sandbox environment, where the agent runs in an externally-managed worktree with detached HEAD.

Key changes:
- **Environment detection** added to `using-git-worktrees` (Step 0) and `finishing-a-development-branch` (Step 1.5) — detects linked worktrees and detached HEAD via read-only git commands
- **Sandbox-aware finishing** — when in a detached HEAD worktree, the agent commits work and outputs a handoff payload instead of trying to branch/push/PR (which would fail)
- **Worktree cleanup guard** — prevents accidental deletion of externally-managed worktrees
- **codex-tools.md** — documents environment detection patterns and Codex App finishing flow
- **Automated tests** for environment detection logic

## Test plan

- [x] Shell tests for environment detection (normal repo, linked worktree, detached HEAD)
- [x] Manual test results documented in plan
- [ ] End-to-end test in Codex App sandbox

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #934, opened 2026-03-25
Clarify Codex App worktree compatibility in using-git-worktrees

q107580018:
## Summary
- prefer Codex App-managed worktrees when App compatibility matters
- treat manual `git worktree` creation as a fallback path rather than the default
- update Python setup guidance to prefer `uv` and clarify manual worktree location rules

## Verification
- ran `git diff --check`
- reviewed the resulting staged diff for `skills/using-git-worktrees/SKILL.md`
- no automated test suite was run because this is a documentation-only change


q107580018:
Closing this in favor of #901, which is already addressing Codex App worktree compatibility more comprehensively.

PR #936, opened 2026-03-25
fix: brainstorm server OWNER_PID dies with ephemeral wrappers

jim80net:
## Problem

The brainstorm companion server shuts down ~60 seconds after startup when `start-server.sh` is invoked through a command wrapper like `timeout`, `env`, or `nice`.

**Observed behavior:** Server emits `server-started`, runs for one lifecycle check interval (60s), then emits `server-stopped` with `reason: "owner process exited"`. The browser shows "This site can't be reached."

**Expected behavior:** Server stays alive until the harness exits or the 30-minute idle timeout fires.

### How OWNER_PID is resolved (`start-server.sh` L102-108)

The script walks 2 levels up the process tree to find the harness:

```
OWNER_PID = ppid of $PPID (intended: harness process)
 └─ $PPID (ephemeral shell spawned by Bash tool)
     └─ $$ (start-server.sh)
```

### What goes wrong with wrappers

When invoked as `timeout 15 start-server.sh ...`, an extra process layer is inserted:

```
Harness (PID H)                     ← intended OWNER_PID
 └─ Bash tool shell (PID A)         ← actual OWNER_PID (grandparent of script)
     └─ timeout (PID B) = $PPID
         └─ start-server.sh (PID C) = $$
```

`OWNER_PID` resolves to **A** (Bash tool shell) instead of **H** (harness). The same mis-resolution occurs with any wrapper: `env`, `nice`, `strace`, `setsid`, etc.

### Why the existing startup validation doesn't catch it

The startup validation (`server.cjs` L329-337) checks whether `ownerPid` is alive at the moment `startServer()` runs:

```javascript
if (ownerPid) {
    try { process.kill(ownerPid, 0); }      // signal 0 = existence check
    catch (e) {
        if (e.code !== 'EPERM') {            // EPERM = alive but different user
            ownerPid = null;                  // dead → disable monitoring
        }
    }
}
```

This catches **already-dead** PIDs (WSL, Tailscale SSH, cross-user — fixed in f076bd3) but **not soon-to-die** ones. At the moment this runs, the Bash tool shell (PID A) is still alive — it's waiting for `start-server.sh` to exit. The validation passes.

### Relationship to f076bd3 (#879)

f076bd3 fixed a different bug with the same symptom: `ownerAlive()` treated EPERM the same as ESRCH, killing the server when the owner PID was a cross-user process (WSL, Tailscale SSH). That fix is correct and necessary but does not address this scenario — here the grandparent is the **same user**, so `process.kill()` returns ESRCH (not EPERM) when the Bash tool shell exits.

Verified by reproducing on Linux:

```
=== Direct invocation ===
grandparent=3644821 → claude (jim) — stays alive ✓

=== Through timeout wrapper ===
grandparent=3962012 → bash (jim) — dies when tool call returns → ESRCH at 60s ✗
```

### Timeline of the failure

```
t=0.0s  start-server.sh launches server.cjs via nohup (OWNER_PID=A)
t=0.1s  server.cjs starts, startup validation: kill(A, 0) → alive ✓
t=0.2s  server.cjs emits server-started JSON
t=2.2s  start-server.sh 2-second liveness window passes, script exits
t=2.3s  Bash tool returns output to harness, shell A exits → A is now dead
t=60s   lifecycle check (L320-323): ownerAlive() → kill(A, 0) → ESRCH → false
t=60s   shutdown('owner process exited') → server dies
```

## Fix

Add a **delayed re-validation** 10 seconds after startup (`server.cjs` L339-355). If `ownerPid` dies within that window, it was an ephemeral wrapper — null it out and fall back to idle timeout.

```javascript
if (ownerPid) {
    const revalidateTimer = setTimeout(() => {
        try { process.kill(ownerPid, 0); }
        catch (e) {
            if (e.code !== 'EPERM') {
                console.log(JSON.stringify({
                    type: 'owner-pid-invalid', pid: ownerPid,
                    reason: 'died shortly after startup (ephemeral wrapper)'
                }));
                ownerPid = null;
            }
        }
    }, 10000);
    revalidateTimer.unref();
}
```

### Why 10 seconds?

- `start-server.sh` has a 2-second liveness verification window (L127-139: 20 × 0.1s sleeps)
- The Bash tool shell exits immediately after the script returns (~2.5s total)
- 10 seconds gives generous margin for slow systems while being well under the 60-second lifecycle check interval
- Any real harness PID will still be alive at t=10s

### Proof of correctness: all scenarios

**Scenario 1: Direct invocation (no wrapper) — harness is real owner**
```
t=0s    OWNER_PID = harness (H)
t=0.1s  startup validation: H alive ✓
t=10s   delayed revalidation: H alive ✓ → ownerPid unchanged
t=60s+  lifecycle checks: H alive ✓ → server stays up
        (When H exits → lifecycle detects → clean shutdown ✓)
```
No behavioral change. Owner monitoring works exactly as before.

**Scenario 2: Wrapper invocation (e.g. `timeout`) — ephemeral shell is owner**
```
t=0s    OWNER_PID = Bash tool shell (A)
t=0.1s  startup validation: A alive ✓ (script still running)
t=2.5s  A exits (Bash tool returns)
t=10s   delayed revalidation: A dead → ownerPid = null
t=60s   lifecycle check: ownerAlive() → ownerPid is null → returns true
        Server stays up, protected by 30-min idle timeout ✓
```
**Fixed.** Server survives. Falls back to idle timeout.

**Scenario 3: Owner PID already dead at startup (WSL/SSH, handled by startup validation)**
```
t=0s    OWNER_PID = stale PID
t=0.1s  startup validation: dead → ownerPid = null
t=10s   delayed revalidation: ownerPid is null → skipped (guarded by `if (ownerPid)`)
```
No behavioral change. Existing startup validation still handles this.

**Scenario 4: No race between delayed revalidation and lifecycle check**
```
t=10s   delayed revalidation fires on event loop → may set ownerPid = null
t=60s   lifecycle check fires on event loop → reads ownerPid
```
Both are `setTimeout`/`setInterval` callbacks on the Node.js event loop (single-threaded). No concurrent mutation of `ownerPid` is possible. The `.unref()` on both timers ensures neither prevents process exit on SIGTERM/SIGINT.

### What this does NOT fix

The underlying fragility is in `start-server.sh` L105's 2-level-up PID resolution, which breaks when extra process layers exist. 

### Considered and rejected:

1. **Harness-exported PID**: Have harnesses pass their own PID via an env var (e.g., `BRAINSTORM_HARNESS_PID`). The plumbing exists — the server already reads `BRAINSTORM_OWNER_PID`. Downside: requires every harness to opt in.

2. **Process tree walk by name**: Walk up the tree looking for the harness process by name (e.g., `claude`). Verified to work on Linux, macOS, and Windows. Downside: superpowers is used with multiple harnesses, so we'd have to enumerate every harness, and that's fragile and/or a maintenance headache. 

## Test plan

- [ ] Launch brainstorm server directly (no wrapper) — verify owner PID monitoring still works, server shuts down when harness exits
- [ ] Launch via `timeout 30 bash start-server.sh ...` — verify server survives past 60s (previously died at ~60s)
- [ ] Launch via `env start-server.sh ...` — verify server survives
- [ ] Kill the real harness after direct launch — verify server shuts down via lifecycle check
- [ ] Wait 30 minutes with no activity — verify idle timeout still works as fallback
- [ ] Check server.log for `owner-pid-invalid` with `reason: "died shortly after startup"` message when wrapper is used

### DIFF
--- a/skills/brainstorming/scripts/server.cjs
+++ b/skills/brainstorming/scripts/server.cjs
@@ -323,11 +323,9 @@ function startServer() {
   }, 60 * 1000);
   lifecycleCheck.unref();
 
-  // Validate owner PID at startup. The shell script walks the process tree
-  // to find the claude/claude.exe process, but if it couldn't find one
-  // (running outside Claude Code, or platform detection failed), ownerPid
-  // will be empty. If a PID was passed but is already dead, disable
-  // monitoring and rely on the idle timeout instead.
+  // Validate owner PID at startup. If it's already dead, the PID resolution
+  // was wrong (common on WSL, Tailscale SSH, and cross-user scenarios).
+  // Disable monitoring and rely on the idle timeout instead.
   if (ownerPid) {
     try { process.kill(ownerPid, 0); }
     catch (e) {
@@ -336,6 +334,24 @@ function startServer() {
         ownerPid = null;
       }
     }
+  }
+
+  // Delayed re-validation: if the owner dies within 10 seconds of startup,
+  // it was an ephemeral wrapper (timeout, env, tool-spawned shell) rather
+  // than the real harness. Null it out and fall back to idle timeout.
+  // This catches the gap the startup validation misses: PIDs that are alive
+  // during startup but die as soon as the launching shell exits.
+  if (ownerPid) {
+    const revalidateTimer = setTimeout(() => {
+      try { process.kill(ownerPid, 0); }
+      catch (e) {
+        if (e.code !== 'EPERM') {
+          console.log(JSON.stringify({ type: 'owner-pid-invalid', pid: ownerPid, reason: 'died shortly after startup (ephemeral wrapper)' }));
+          ownerPid = null;
+        }
+      }
+    }, 10000);
+    revalidateTimer.unref();
   }
 
   server.listen(PORT, HOST, () => {
--- a/skills/brainstorming/scripts/start-server.sh
+++ b/skills/brainstorming/scripts/start-server.sh
@@ -99,45 +99,12 @@ fi
 
 cd "$SCRIPT_DIR"
 
-# Resolve the Claude Code harness PID by walking up the process tree
-# and looking for a process named "claude" (Linux/macOS) or "claude.exe"
-# (Windows). This is robust against arbitrary wrapper layers (timeout,
-# env, nice, etc.) that insert extra processes between the harness and
-# this script.
-OWNER_PID=""
-_walk_pid="$$"
-while [[ -n "$_walk_pid" && "$_walk_pid" != "1" ]]; do
-  _parent="$(ps -o ppid= -p "$_walk_pid" 2>/dev/null | tr -d ' ')"
-  [[ -z "$_parent" || "$_parent" == "1" ]] && break
-  _comm="$(ps -o comm= -p "$_parent" 2>/dev/null | tr -d ' ')"
-  if [[ "$_comm" == "claude" ]]; then
-    OWNER_PID="$_parent"
-    break
-  fi
-  _walk_pid="$_parent"
-done
-
-# Windows/MSYS2: ps can't see Windows process names. Use PowerShell to
-# walk the Windows PID tree looking for claude.exe. Node.js process.kill()
-# works with Windows PIDs, so the server can monitor it directly.
-if [[ -z "$OWNER_PID" ]]; then
-  case "${OSTYPE:-}" in
-    msys*|cygwin*|mingw*)
-      # Map this script's MSYS2 PID to a Windows PID, then walk up.
-      _winpid="$(ps -p $$ | tail -1 | awk '{print $4}')"
-      if [[ -n "$_winpid" ]]; then
-        OWNER_PID="$(powershell.exe -Command "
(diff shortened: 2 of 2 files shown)


obra:
So this isn't gonna work if it is looking for the claude code process by name. Because Superpowers works with enough different harnesses, that's not really a safe choice.

jim80net:
> So this isn't gonna work if it is looking for the claude code process by name. Because Superpowers works with enough different harnesses, that's not really a safe choice.

In this case, I can't think of a non-fragile solution that isn't using a dumb heuristic that re-checks the owner PID 10 seconds after startup. If it's already dead by then, it was an ephemeral wrapper (timeout, env, etc.), not the real harness. Null it out and fall back to the idle timeout. I've updated (or reverted actually as this was the first iteration) to this.

Downside is that it is a heuristic. If there's a pattern mismatch for genuine use case where owner PID exits legitimately after 10 seconds, it's going to fall back to the idle timeout.

jim80net:
> So this isn't gonna work if it is looking for the claude code process by name. Because Superpowers works with enough different harnesses, that's not really a safe choice.

Makes sense. Short sighted of me.

PR #949, opened 2026-03-27
docs: relax writing-skills guidance for Codex interop

hanl001:
Makes `writing-skills` more compatible with Codex-specific scaffolding guidance by:
- clarifying platform-specific responsibility boundaries
- allowing a brief capability phrase in `description` when it improves discovery
- clarifying that structural validation does not replace behavioral validation


obra:
I don't think this is correct, so don't currently plan to merge it.

Issue #895, opened 2026-03-23
writing-plans: plans over-specify implementation, leaving no room for executor judgment


qdon314:
- [x] I searched existing issues and this has not been proposed before                                                                                                     
   
  ## What problem does this solve?                                                                                                                                           
                                                                  
  When using `writing-plans` to plan a multi-task implementation, the generated plan contained complete implementation code — full function bodies, 70-line test files, exact
   shell commands, and step-by-step TDD sequences per task. The executor (`executing-plans` / `subagent-driven-development`) had no decisions to make — it was copy-pasting
  code from the plan rather than engineering a solution.                                                                                                                     
                                                                  
  This caused two concrete problems:                                                                                                                                         
  1. When implementation revealed a better approach mid-execution, the plan's code was wrong even though the intent was still valid. Adapting meant "violating" the plan.
  2. Every task duplicated the TDD ceremony ("write failing test → verify failure → implement → verify pass → commit") that `executing-plans` and `test-driven-development`  
  already enforce, bloating each task from ~20 lines to 80-100+.                                                                                                             
                                                                                                                                                                             
  ## Proposed solution                                                                                                                                                       
                                                                  
  Shift `writing-plans` to produce **interface contracts + acceptance criteria** instead of implementation code. Each task would include:                                    
                                                                  
  | Field | Purpose |                                                                                                                                                        
  |-------|---------|                                             
  | **Why** | Connects this task to the larger plan rationale |
  | **Files** | Exact paths: create, modify, test |                                                                                                                          
  | **Contract** | Type shapes and signatures (not bodies) |                                                                                                                 
  | **Acceptance** | Observable behaviors, edge cases, test expectations |                                                                                                   
  | **Constraints** | Architectural rules, stability guarantees |                                                                                                            
                                                                                                                                                                             
  And explicitly **not** include: function bodies, test code, shell commands, TDD step sequences, or commit messages.                                                        
                                                                                                                                                                             
  **Before (current):** A task to create four `StrEnum` classes included 4 `__init__.py` files, a 70-line test file with full assertions, a complete implementation file, and
   TDD step instructions.                                         
                                                                                                                                                                             
  **After (proposed):** The same task:                            

  > **Task 1: Benchmark domain enums**                                                                                                                                       
  >
  > **Why:** All other domain models depend on these classification types.                                                                                                   
  >                                                               
  > **Files:** Create `src/benchmark/domain/enums.py`, test at `tests/benchmark/domain/test_enums.py`
  >
[...]

PR #887, opened 2026-03-23
feat(subagent-dev): accumulate discoveries across tasks

mvanhorn:
## What problem are you trying to solve?

When using subagent-driven-development, each subagent starts completely fresh. Knowledge discovered during one task is invisible to the next. Real example from #601: Task 9 discovered that Drizzle ORM wraps `PostgresError` inside `DrizzleQueryError` (check `error.cause`). Task 10 hit the same issue independently, debugged it from scratch, and fixed it again. Wasted tokens and time.

Two independent users confirmed this in #601. @Koroqe built a file-based workaround in their own project to fix it.

## What does this PR change?

Adds a `Project Discoveries` field to the implementer report format and controller logic to accumulate and inject those discoveries into subsequent subagent prompts. 16 lines across 2 files.

## Is this change appropriate for the core library?

Yes. This affects the SDD controller-subagent coordination pattern, which is core infrastructure. It follows the existing "controller curates context for subagents" architecture. No new concepts, no new files, no domain-specific behavior.

## What alternatives did you consider?

1. **File-based persistence** (like @Koroqe's workaround): Have subagents write discoveries to a file and later subagents read it. Rejected because it adds filesystem state and breaks the controller-curates-context pattern. The controller already constructs prompts - this extends that.
2. **Full project briefing system** (test commands, patterns, imports): More comprehensive but larger scope. This can be a follow-up if discoveries prove useful.

## Does this PR contain multiple unrelated changes?

No. Both changes serve the same purpose: one adds the report field, the other adds the accumulation and injection logic.

## Existing PRs

- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: #738 (fresh-eyes review - different approach to cross-task learning), #804 (CLAUDE.md reading - different kind of context injection)
- No PR implements the specific discovery-accumulation approach from #601.

## Environment tested

| Harness | Harness version | Model | Model version/ID |
|---------|-----------------|-------|------------------|
| Claude Code | 2.1+ | Claude | claude-opus-4-6 |

## Evaluation

The initial prompt was reviewing #601 and proposing a minimal implementation. This is a Markdown instruction change, so evaluation is structural: verified frontmatter parses correctly, template fences are balanced, report format includes the new field, and the new SKILL.md section follows existing voice and structure. The change is small enough that its correctness is verifiable by reading the 16-line diff.

## Rigor

- [ ] If this is a skills change: I used `superpowers:writing-skills` and completed adversarial pressure testing (paste results below)
[...]

Issue #750, opened 2026-03-15
Superpowers consume a lot of tokens in Opencode with Codex


cardgabriel:
**Describe the bug**
With the last version, Superpowers consume a lot of tokens in Opencode with Codex 5.4

obra:
Can you provide specifics about what you're seeing? What is "a lot of tokens" And what parts of the process seem to be consuming those tokens? And in what way did it differ in 4.x?

lucas-grunevald:
I experienced something similar in Claude Code with Pro plan. I ask to execute a relatively simple task. The session usage limit was reached really fast and the task was not completly done. I tried the same task later, but ask to skip spec review in "Brainstorm" skill, and the result was 50% less usage than the previous attempt(i did not noticed that "Writing plans" also has this review loop, so it could have dimished even more if this was skipped too).

Other thing that i noticed, is that "Subagent-Driven-Development" is now enforced. For me this is bad, because  in my experience this skill uses more tokens and deliver worse results than "Executing-Plans" skill. Also, i review all the code generated, so i don't need two more agents to review it too.

It seems that the goal of this project is shifting towards "autonomous" workflow. I use AI agents to increase productivity, not to replace me, so if the goal is really what i said, it will not make sense to use these skills anymore.

solrachix:
the same thing with me, especially Brainstorm. but I was using them all on the Codex-cli plan and I used up all the credits for the week in one code day lol

obra:
@lucas-grunevald Can you talk about the "worse results" you're seeing?

obra:
@lucas-grunevald The dev branch now restores the choice of executing plans. I'd greatly appreciate it if you could git clone and checkout the dev branch and use the --plugin-dir option to claude code to test it out.

lucas-grunevald:
> [@lucas-grunevald](https://github.com/lucas-grunevald) Can you talk about the "worse results" you're seeing?

subagent-driven-development "use the least powerful model  that can handle each role", so if it has to handle anything that wasn't described in the plan document, like an error or a bug,  or even if the plan itself wasn't completly right, it really struggles. In a case where this happened, it just "brute forced" a solution, leading to more time and more tokens spent.

lucas-grunevald:
I have a suggestion:

Instead of enforce the next skill, you could have a customizable config file. This file would contain a sequence of skills to follow, and some options for each skill(skip some step, add a step, or whatever you think makes sense).
[...]

Issue #730, opened 2026-03-14
Feature request: Cross-model collaboration via external CLI tools (Codex, Gemini) for review and second opinions


vlassisemm:
**_Is your feature request related to a problem? Please describe._**
The current review pipeline (spec-compliance → code-quality) uses the same model family for both implementation and review. This means the reviewer shares the same biases and blind spots as the implementer — it's essentially self-review with extra steps. I've found that getting a second opinion from a structurally different model (Gemini, Codex CLI) catches issues that same-model review consistently misses.

**_Describe the solution you'd like_**
A skill or extension to the existing review pipeline that can invoke an external LLM via CLI (e.g. gemini, codex) as an additional review stage. It would pass the relevant context (diff, spec, task description) to the external model and collect a structured verdict. This could slot in after the existing two-stage review as an optional third stage, or replace one of the existing stages when configured.

**_Describe alternatives you've considered_**
A standalone cross-model-review skill that users invoke manually when they want a second opinion, separate from the subagent-driven-development flow.
A configuration option in CLAUDE.md (or a skill-level config) that lists available external CLI tools, letting the controller decide when to use them based on task complexity.
Keeping it entirely outside superpowers and handling orchestration in a personal skill — but this means everyone builds their own version.

**_Additional context_**
I've been experimenting with this pattern locally — orchestrating multiple LLM reviewers in parallel with consensus-based triage, inspired by the "No Self-Judging" principle (a model shouldn't be the sole evaluator of its own output).

annexiao:
Strong +1 on this. My Codex MCP also uncovers important findings that may be Claude's systematic blind spots. I think a pluggable second-pass reviewer can works cleanly if it's implemented strictly as opt-in for users who have configured an external reviewer. It changes nothing for users who do not opt in.

obra:
Hi @vlassisemm — I'm Claude, an AI agent (Fable 5, running in Claude Code), posting from Jesse (@obra)'s account at his direction. Jesse had me run a skeptical triage of all 340 open superpowers items — every factual claim tested against the current `dev` tree, then adversarially re-checked by a second, independent agent — and he reviewed the verdicts and directed this batch of closures. This one is closing as a **venue call** — a decision about where the work belongs, not a judgment of its quality:

[...]

PR #956, opened 2026-03-27
feat(subagent-dev): add optional external delegate mode for implementation tasks

mvanhorn:

Users on metered plans (Claude Pro, Codex weekly credits) burn through their token allocation running superpowers sessions with many mechanical tasks. Each implementation subagent consumes tokens from the same provider as the controller. On a session with 8+ tasks, the implementer subagents account for the majority of token spend, even though most mechanical tasks (1-2 files, clear spec) don't need the controller's model.

The skill already says "use the least powerful model that can handle each role" - but that principle stops at the provider boundary. There's no way to route cheap tasks to a different provider's CLI that might be faster, cheaper, or better suited for mechanical implementation.

Related issues:
- #730 (4 thumbs up) - requests cross-model collaboration via external CLI tools
- #750 (7 comments) - users exhausting weekly credits in one session
- #895 - @obra noted "the intent is that you should be able to use haiku (or even less capable models) as the implementers"

## What does this PR change?

Adds an optional "External Delegate" section to subagent-driven-development. When users configure `external-delegate: <cli-name>` in their AGENTS.md, mechanical implementation tasks (cheap-model tier) get piped to the external CLI instead of spawning a same-provider subagent. Also adds `external-delegate-prompt.md` - a streamlined implementer prompt for delegates.

Yes. This is provider-agnostic infrastructure, not a third-party integration. It works with any CLI that accepts a prompt on stdin and writes to the filesystem. The AGENTS.md example uses `codex` but the mechanism is generic. It extends an existing design pattern (model selection tiers) rather than adding a new one.

The feature is opt-in. Without `external-delegate` in AGENTS.md, behavior is identical to today. No new dependencies, no scripts, no external service calls from superpowers itself.

1. **Cross-model review only** (#730's original proposal) - smaller scope but only addresses the bias problem, not token conservation. If the delegate interface exists for implementation, adding review delegation later is trivial.

2. **Token budget tracking** - auto-switch to delegate when nearing billing limits. Too complex, too many assumptions about billing APIs. Explicit opt-in via AGENTS.md is simpler. Also aligns with @obra's response on #750: "You can do that today - just put those things in your AGENTS.md."

3. **Separate skill** - a standalone `external-delegate` skill instead of extending subagent-driven-development. This would fragment the execution flow and require users to choose between two competing execution skills. Extending the existing skill preserves the single execution path.

No. One feature: external delegate mode. Two files: the section in SKILL.md and the prompt template.

## Existing PRs
- Related PRs: #113 (closed, Nov 2025), #730 (open issue), #750 (open issue), #887 (open, mine - different feature: accumulated discoveries)

PR #113 was a larger scope ("comprehensive Codex CLI integration" with utility scripts, hooks, config files). This PR is instruction-only - 44 lines of markdown added to SKILL.md plus a prompt template. No code, no scripts, no config files. The approach is fundamentally different: opt-in via AGENTS.md rather than automatic delegation.

## Environment tested

|---------|----------------|-------|------------------|
| Claude Code | 1.0.33 | Claude | claude-opus-4-6 |

## Evaluation

Initial prompt: "Research obra/superpowers for external delegate/Codex delegation patterns. If none exist, propose adding one."

This change adds skill instructions (markdown), not executable code. The evaluation is structural:
- The new section slots cleanly after Model Selection, extending the existing cheap/standard/capable tier system
- The prompt template follows the same structure as `implementer-prompt.md` with sections removed that the controller handles (self-review, escalation)
- The circuit breaker and fallback patterns match how superpowers handles other failure modes
- Red Flags additions are consistent with existing entries

I have not run multi-session evals comparing sessions with and without external delegation. The instructions are new and untested at scale. I'd welcome feedback on the circuit breaker threshold (currently 3 failures) and the task-tier classification rules.

## Video Demo

Demo

## Rigor

- [x] If this is a skills change: I reviewed `superpowers:writing-skills` patterns. Adversarial pressure testing was not performed (this is a new feature proposal, not a modification to existing behavior-shaping content).
- [x] This change was reviewed for edge cases: delegate unavailable (fallback), delegate fails (circuit breaker), out-of-scope changes (diff review + fallback), git operations (controller-only)
- [x] I did not modify any existing behavior-shaping content. All changes are additive.

## Human review

Relates to #730, #750

### DIFF
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -99,6 +99,45 @@ Use the least powerful model that can handle each role to conserve cost and incr
 - Touches multiple files with integration concerns → standard model
 - Requires design judgment or broad codebase understanding → most capable model
 
+## External Delegate (Optional)
+
+When configured in AGENTS.md (`external-delegate: <cli-name>`), mechanical implementation tasks can be delegated to an external CLI tool instead of a same-provider subagent. This extends the "least powerful model" principle across providers - when a cheaper, faster tool can handle the task, use it.
+
+**When to delegate externally:**
+- Task is cheap-model tier (1-2 files, complete spec, isolated function)
+- External CLI is available and responding
+- The delegate hasn't failed 3+ times this session
+
+**When NOT to delegate externally:**
+- Integration tasks touching multiple files with coordination concerns
+- Architecture or design decisions
+- Review tasks - spec compliance and code quality reviews always use same-provider subagents
+- The external CLI has hit the circuit breaker (3 consecutive failures)
+
+**How it works:**
+
+The controller's responsibilities stay the same. Only the implementer dispatch changes:
+
+1. Check delegate availability: `which <cli-name> >/dev/null 2>&1`
+2. Build the implementation prompt using `./external-delegate-prompt.md` (a streamlined implementer prompt without self-review or escalation sections - the controller handles those)
+3. Write the prompt to a temp file and pipe it to the delegate: `cat /tmp/sp-delegate-prompt.md | <cli-name> <flags>`
+4. Review the diff: verify it's non-empty, in-scope, and contains no rogue operations
+5. If the diff looks good, proceed to spec compliance review (same as the normal flow)
+6. If the diff is empty or out-of-scope, fall back to a same-provider subagent for this task
+7. The controller always commits - the delegate should not touch git
+
+**Circuit breaker:** If the external delegate fails 3 consecutive times in one session, disable it for the remainder and use same-provider subagents for all remaining tasks. Reset the counter on any successful delegation.
+
+**Configuration example in AGENTS.md:**
+
+```markdown
+## External Delegate
+
+external-delegate: codex
+```
+
+The controller reads this at session start. If the key is absent or empty, all tasks use same-provider subagents (the default behavior, unchanged).
+
 ## Handling Implementer Status
 
 Implementer subagents report one of four statuses. Handle each appropriately:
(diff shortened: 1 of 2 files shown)


mvanhorn:
Can't tag you on X, but would love any thoughts here if you think this could work.  Thanks for considering, love what youve built here!

obra:
Hi @mvanhorn — I'm Claude, an AI agent (Fable 5, running in Claude Code), posting from Jesse (@obra)'s account at his direction. Jesse had me run a skeptical triage of all 340 open superpowers items — every factual claim tested against the current `dev` tree, then adversarially re-checked by a second, independent agent — and he reviewed the verdicts and directed these closures. This one is closing as a **venue call** — a decision about where the work belongs, not a judgment of its quality:

Thanks for the careful writeup and the honest "untested at scale" disclosure. We're closing this as a venue call, not a quality judgment. The PR bakes a named external CLI into a core skill — `external-delegate: codex` in the AGENTS.md example plus a shipped external-delegate-prompt.md — and that's the same shape we just declined in #730, which you cite as motivation and which we closed today: "the core-compatible shape is a generic, zero-dependency hook/extension point... not a built-in Codex/Gemini integration." We used that same reasoning back in May on #1005, a narrower version of this request. Opt-in via AGENTS.md doesn't exempt it under CLAUDE.md's third-party-dependency policy. The better home is a standalone plugin attaching through that generic extension point. Two asides, neither the reason we're closing: #750, which you cite for token burn, turned out to be a bootstrap-injection bug fixed in v5.0.7 rather than implementer cost; and dev's review flow has since consolidated into a single task-reviewer step, so a couple of the delegate prompt's assumptions are already stale. Also minor: this targets `main` and we land on `dev`, but that's just housekeeping, not why we're closing.

---

Issue #975, opened 2026-03-28
Brainstorm session storage should not default to the working tree


norman-abramovitz:
## Problem

The brainstorm visual companion creates a `.superpowers/` directory in the repository working tree by default when `--project-dir` is passed. The `visual-companion.md` documentation then instructs Claude to tell the user to add `.superpowers/` to `.gitignore`.

For open source projects, this creates two problems:

1. **Repo pollution** — AI tool artifacts don't belong in project repos. Contributors who don't use superpowers shouldn't see evidence of it.
2. **The .gitignore paradox** — Adding `.superpowers/` to `.gitignore` solves the git tracking problem but leaks the fact that the project uses AI tooling. In open source, the repo should look the same regardless of what tools contributors use.

## Current behavior

- `start-server.sh --project-dir /path/to/repo` → creates `/path/to/repo/.superpowers/brainstorm/<session-id>/`
- Without `--project-dir` → falls back to `/tmp/brainstorm-<session-id>/` (world-readable, sessions lost on reboot)
- `visual-companion.md` line 48: "Remind the user to add `.superpowers/` to `.gitignore` if it's not already there"

## Suggestions

1. **Add a `--state-dir` flag** (or rename `--project-dir` to `--state-dir`) that separates "where to store sessions" from "what project I'm working on." The current name implies it sets the project context, not just a storage path.

2. **Support an environment variable** (e.g., `SUPERPOWERS_STATE_DIR`) so developers can configure this once in their shell profile rather than passing it every invocation. This aligns with how `CLAUDE_CONFIG_DIR` works for Claude Code.

3. **Default to a user-owned directory** (e.g., `~/.local/share/superpowers/` or `${XDG_DATA_HOME}/superpowers/`) instead of the working tree. The `/tmp` fallback is world-readable on macOS/Linux, which is a concern for sessions containing design mockups or architecture discussions.

4. **Remove the `.gitignore` instruction** from `visual-companion.md`. If sessions are stored outside the repo by default, there's nothing to ignore.

## Context

I'm working on a methodology for keeping AI tool artifacts out of open source repos — covering Claude Code, Cursor, Copilot, Windsurf, and others. Superpowers is actually one of the better tools here since `--project-dir` provides a relocation mechanism. These suggestions would close the remaining gaps.

sgaduuw:
[...]

PR #977, opened 2026-03-28
feat: add explicit brainstorm state-dir selection

yejiming:

The brainstorm companion currently pushes persistent session state into the working tree when `--project-dir` is used, under `<project>/.superpowers/brainstorm/`.

For open source repos, that creates repo pollution and leads users toward `.gitignore` churn just to hide AI-tool state. Persistent brainstorm state should default to a user-owned location outside the repository, while still allowing explicit in-repo storage when someone really wants it.

## What does this PR change?

This adds a new `--state-dir` flag and `SUPERPOWERS_STATE_DIR` environment variable for explicit brainstorm state storage.

It also changes the default persistent storage root to a user-owned data directory outside the working tree, and keeps `--project-dir` as a compatibility alias that still resolves to `<project>/.superpowers/brainstorm/<session-id>/`.

Yes.

This is general-purpose core behavior for brainstorm session storage, not a project-specific workflow or third-party integration. Any user of the brainstorm companion can hit the same repo-pollution problem.

1. Remove `--project-dir` immediately.
   Rejected because it would break existing workflows that already rely on in-repo persistence.

2. Keep `/tmp` as the default persistent location.
   Rejected because it is ephemeral and may be unsuitable for retaining brainstorm sessions across restarts.

3. Add a more complex platform-specific storage abstraction.
   Rejected because a simple user-owned default path plus explicit override mechanisms solves the problem with less complexity.

No. This PR is limited to brainstorm session storage selection and the associated documentation/tests.

## Existing PRs
- Related PRs: none found

## Environment tested

| Codex desktop | current | GPT-5.4 | GPT-5.4 |

## Evaluation

  - Investigate issue #975 and implement a minimal fix that moves brainstorm state out of the working tree by default while preserving compatibility.

  - One targeted shell test pass for path selection plus one manual smoke run of `start-server.sh` with `--state-dir`.

  - Before: the script only supported `--project-dir`, and persistent in-repo storage was the documented path.
  - After: the script supports `--state-dir` and `SUPERPOWERS_STATE_DIR`, defaults to a user-owned storage directory outside the working tree, and still accepts `--project-dir` as a compatibility alias.

## Rigor

- [x] If this is a skills change: I used `superpowers:writing-plans` style discipline for scoped implementation and added targeted test coverage for the new path-selection behavior

Verification run:
- `bash tests/brainstorm-server/start-server-paths.test.sh`
- `git diff --check`
- manual smoke run:
  - `bash skills/brainstorming/scripts/start-server.sh --state-dir /tmp/sp-state-smoke --foreground`

## Human review

### DIFF
--- a/skills/brainstorming/scripts/start-server.sh
+++ b/skills/brainstorming/scripts/start-server.sh
@@ -1,13 +1,15 @@
 #!/usr/bin/env bash
 # Start the brainstorm server and output connection info
-# Usage: start-server.sh [--project-dir <path>] [--host <bind-host>] [--url-host <display-host>] [--foreground] [--background]
+# Usage: start-server.sh [--state-dir <path>] [--project-dir <path>] [--host <bind-host>] [--url-host <display-host>] [--foreground] [--background]
 #
 # Starts server on a random high port, outputs JSON with URL.
 # Each session gets its own directory to avoid conflicts.
 #
 # Options:
-#   --project-dir <path>  Store session files under <path>/.superpowers/brainstorm/
-#                         instead of /tmp. Files persist after server stops.
+#   --state-dir <path>    Store session files under <path>/<session-id>/.
+#                         This is the primary persistent storage option.
+#   --project-dir <path>  Compatibility alias. Stores session files under
+#                         <path>/.superpowers/brainstorm/<session-id>/.
 #   --host <bind-host>    Host/interface to bind (default: 127.0.0.1).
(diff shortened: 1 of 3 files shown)


obra:
It looks like you opened two similar pull requests for this same thing. What's going on?

yejiming:
> It looks like you opened two similar pull requests for this same thing. What's going on?

Yeah, I'll close the other one

PR #982, opened 2026-03-29
feat: add entroly-context skill for codebase-aware subagent dispatch

juyterman1000:
## Summary

Adds a new entroly-context skill that solves a specific pain point in subagent-driven-development: each dispatched subagent starts with zero awareness of the codebase.

## The Problem

When using subagent-driven-development, the coordinator dispatches fresh subagents per task. These agents don't know what files exist, what functions are defined, or what imports are valid. This leads to hallucinated imports, duplicate implementations, and wrong function signatures.

## The Solution

Entroly (https://github.com/juyterman1000/entroly) is an open-source context engineering engine that builds a dependency graph of the codebase and compresses the most relevant parts into a token-budget-aware snapshot using KKT optimization (Rust core).

This skill tells the coordinator to:
1. Run entroly optimize once before dispatching
2. 2. Inject the output as codebase context into each implementer prompt
## Integration

Works directly alongside subagent-driven-development. The skill is optional and gracefully skips if Entroly is not installed.

### DIFF
--- a/skills/entroly-context/SKILL.md
+++ b/skills/entroly-context/SKILL.md
@@ -1,0 +1,84 @@
+---
+name: entroly-context
+description: Use before implementing tasks that touch multiple files in a large codebase - injects a dependency-graph-aware context snapshot so subagents aren't working blind
+---
+
+## When to Use
+
+Activate before any implementation task where:
+- The codebase has **more than ~20 files** of application logic
+- The task touches files that **import from or are imported by** other files
+- You are about to **dispatch subagents** and want them to understand how parts of the codebase relate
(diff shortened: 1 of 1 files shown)


obra:
I think this is a little too specialized for superpowers and also seems like it's a bit of an ad for your project.

juyterman1000:
Thanks for reviewing, @obra! .I understand keeping Superpowers broadly useful and ad-free. My goal was just to show an optional skill that could help users save a lot of tokens and make AI results more reliable, without affecting anything else in the repo. Totally respect your decision, just thought it might be a small win some users could have appreciated.

Issue #999, opened 2026-03-30
finishing-a-development-branch Option 1: cleanup sequence fails inside worktree


csillag:
## Problem

The finishing skill's Option 1 (Merge Locally) has two cleanup bugs when running from inside a worktree — the expected context after worktree-based execution.

### Bug 1: Branch deletion before worktree removal

After merging, the skill runs `git branch -d <feature-branch>` while the worktree still has that branch checked out. Git refuses:

```
error: cannot delete branch 'feature-branch' used by worktree at '/path/to/.worktrees/...'
```

The fix is to remove the worktree first, then delete the branch. The current order in the skill (Step 4 then Step 5) is reversed.

### Bug 2: False "discard work" warning on already-merged commits

When removing the worktree after a successful merge, the worktree exit tool warns about discarding commits:

```
Worktree has 4 commits on feature-branch. Removing will discard this work permanently.
```

These commits were already fast-forward merged to the base branch — they are not at risk. The safety check should verify whether the branch's commits are reachable from the base branch before warning.

## Reproduction

1. Use `using-git-worktrees` to create a worktree
2. Make commits on the feature branch
3. Invoke `finishing-a-development-branch`, choose Option 1
4. Observe: branch deletion fails, then worktree removal warns about already-merged commits

## Suggested Fix

- Swap the order: remove worktree (Step 5) before deleting branch (end of Step 4)
- Before warning about discarding commits, check `git merge-base --is-ancestor <feature-branch> <base-branch>` — if true, the commits are already merged and removal is safe

## Context

Found during manual testing of #997. These are pre-existing bugs, not introduced by that PR. Related: #167 (checkout failure from worktree, same Option 1 flow).

bakathefish:
## Proposed Fix

Two bugs in `skills/finishing-a-development-branch/SKILL.md`:

### Bug 1: Branch delete before worktree removal (Option 1)

Option 1 (lines 69-86) runs `git branch -d <feature-branch>` while the worktree still has that branch checked out. Git rejects this with: `"error: cannot delete branch used by worktree"`.

**Fix:** Move `git branch -d <feature-branch>` out of the Option 1 code block. Place it AFTER Step 5's `git worktree remove`:

```markdown
#### Option 1: Merge Locally

\`\`\`bash
# Switch to base branch
git checkout <base-branch>

# Pull latest
git pull

# Merge feature branch
git merge <feature-branch>

# Verify tests on merged result
<test command>
\`\`\`

[...]

PR #997, opened 2026-03-30
feat: worktree-first isolation with delta analysis for parallel session safety

csillag:
## What problem are you trying to solve?

When running multiple AI coding sessions in parallel against the same repo, the brainstorming → writing-plans → execution chain works against a codebase that may change at any time. Session A brainstorms and writes a spec against code state X, but by the time execution finishes, Session B has merged changes that invalidate the spec, plan, or implementation. The agent merges anyway because it has no mechanism to detect the drift.

Detailed problem description: #989

## What does this PR change?

Moves worktree creation to step 1 of brainstorming (before code analysis), adds `.superpowers-session.json` metadata tracking across the skill chain, and adds rebase + delta analysis with 3-level escalation to the finishing skill before merge.

## Is this change appropriate for the core library?

Yes — this affects the fundamental skill chain used by all superpowers users. Parallel sessions are a natural workflow for any developer using AI coding agents. The fix is entirely within the existing skill infrastructure.

## What alternatives did you consider?

Three approaches evaluated during brainstorming (documented in spec):

1. **Worktree-first (chosen)** — freeze codebase at brainstorming start, validate before merge. Simplest mental model, strongest guarantees.
2. **Metadata-only tracking** — don't move worktree creation, just record the analysis commit and detect staleness. Weaker: doesn't prevent drift, only detects it.
3. **Two-phase isolation** — brainstorm on main (fresh view), worktree at plan-writing time. Leaves a gap between analysis and isolation.

## Does this PR contain multiple unrelated changes?

No. All changes serve the same goal: isolate the skill chain from parallel session interference. The worktree-first change and delta analysis are the two halves of the same solution (isolate at start, validate before merge).

## Existing PRs
- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: #632, #675, #829 (all restore the missing worktree step from #574, placing it after design approval), #483 (original, closed stale)

All three related PRs address the narrower problem of the missing worktree step. This PR subsumes them by moving the step earlier (step 1 vs after design approval) and adding the full delta analysis system that none of them include.

## Environment tested

| Harness | Harness version | Model | Model version/ID |
|---------|-----------------|-------|------------------|
| Claude Code | 2.1.87 | Claude Opus 4.6 | claude-opus-4-6 |

## Evaluation

- Initial prompt: "Please read https://github.com/obra/superpowers/issues/989" followed by a full brainstorming session to design the solution
- 5 manual eval sessions after changes, testing:
[...]

PR #952, opened 2026-03-27
fix(finish): cd to main repo before worktree removal

JasonOA888:
## Summary

Fixes #238

## Problem

The `/finish` workflow's Step 5 runs `git worktree remove <path>` without first changing the working directory. When the current shell session is inside the worktree being removed, this fails:

```
fatal: cannot remove worktree '<path>': '<path>' is the current working directory
```

This is a common scenario when Claude Code's shell is still inside the feature worktree.

## Solution

Before running `git worktree remove`, change directory to the main repository:

```bash
# Get main repo path (first entry in worktree list is always main)
main_repo=$(git worktree list | head -1 | awk '{print $1}')

# Change to main repo before removal
cd "$main_repo"

# Now safe to remove
git worktree remove <worktree-path>
```

## Testing

- Verified that `git worktree list | head -1 | awk '{print $1}'` correctly returns the main repo path
- Removal succeeds after cd-ing out of the worktree

## Files Changed

- `skills/finishing-a-development-branch/SKILL.md` — Updated Step 5 with directory change

PR #913, opened 2026-03-24
fix(finishing): don't cleanup worktree when creating PR

chris-yyau:
## What problem are you trying to solve?

Step 5 "Cleanup Worktree" in `finishing-a-development-branch/SKILL.md` lists Options 1, 2, 4 for worktree cleanup. But three other sections in the same document disagree:

- **Quick Reference table** (line 157): Option 2 shows "Keep Worktree: ✓"
- **Common Mistakes** (line 173): "Only cleanup for Options 1 and 4"
- **Red Flags** (line 191): "Clean up worktree for Options 1 & 4 only"

When a user picks Option 2 (Create PR), the agent cleans up the worktree — but the user may still need it while the PR is under review. This contradicts the document's own guidance.

## What does this PR change?

Updates Step 5 to cleanup worktree for Options 1 and 4 only, matching the Quick Reference table, Common Mistakes, and Red Flags sections.

## Is this change appropriate for the core library?

Yes — this fixes an internal contradiction in a core skill document. The finishing skill is general-purpose infrastructure used by all users.

## What alternatives did you consider?

1. **Update the other three sections to include Option 2** — Rejected because cleaning up a worktree while a PR is under review removes the user's ability to iterate on review feedback.
2. **Add a conditional note** — Rejected as over-engineering. The document already established the right behavior in 3 out of 4 places; simplest fix is aligning the outlier.

## Does this PR contain multiple unrelated changes?

No. Single line change fixing one contradiction.

## Existing PRs
- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: none found

## Environment tested

| Harness | Harness version | Model | Model version/ID |
|---------|-----------------|-------|------------------|
| Claude Code | 1.0.33 | Claude | claude-opus-4-20250514 |

## Evaluation

- Initial prompt: "Check for internal contradictions in superpowers skill documents"
- Verified by reading the full finishing skill document and confirming all four sections now agree on cleanup behavior (Options 1 and 4 only).
- Single documentation fix — behavioral eval not applicable.

## Rigor

- [ ] If this is a skills change: I used `superpowers:writing-skills` and completed adversarial pressure testing (paste results below)
- [x] This change was tested adversarially, not just on the happy path
- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) without extensive evals showing the change is an improvement

This change aligns Step 5 with the existing Red Flags and Common Mistakes tables — it does not modify their content.

## Human review
[...]

PR #1000, opened 2026-03-30
fix: worktree-aware merge and correct cleanup sequence in finishing skill

csillag:

The finishing skill's Option 1 (Merge Locally) and Option 4 (Discard) fail when running inside a worktree — which is the expected context after worktree-based execution. Three concrete failures:

1. `git checkout <base-branch>` fails with `fatal: 'main' is already used by worktree at ...` because the base branch is already checked out in the original repo
2. `git branch -d <feature-branch>` fails with `error: cannot delete branch used by worktree` because the worktree still has the branch checked out (wrong ordering)
3. The worktree removal tool warns about "discarding" commits that were already fast-forward merged to the base branch

## What does this PR change?

Adds worktree detection to Option 1 (merge from original repo instead of checkout), fixes cleanup ordering in Step 5 (remove worktree before deleting branch), and adds a `merge-base --is-ancestor` safety check before warning about unmerged commits.

Yes — this fixes bugs in a core skill that affect anyone using worktrees with the finishing workflow, which is the standard path.

1. **Merge from original repo (chosen)** — detect worktree, find original repo via `git worktree list --porcelain`, merge there. Simple, reliable.
2. **cd to original repo first** — similar to #952's approach, but doesn't handle the merge command itself.
3. **Require user to cd out of worktree first** — current implicit workaround, undocumented and unintuitive.

No. All three fixes are in the same Option 1 / Step 5 cleanup flow and share the same root cause: the skill wasn't designed for execution from inside a worktree.

## Existing PRs
- Related PRs: #391 (CWD ordering fix, overlapping), #952 (cd before worktree removal, subset), #913 (Option 2 cleanup inconsistency, separate)
- Related issues: #167 (checkout failure, closed stale but unfixed), #999

#391 overlaps on the reordering fix. This PR additionally handles worktree-aware merge (Option 1) and the merge-base safety check, which #391 does not cover.

## Environment tested

| Claude Code | 2.1.87 | Claude Opus 4.6 | claude-opus-4-6 |

## Evaluation

- Found during manual testing of #997 (worktree-first isolation)
- 3 sessions observing the failure mode before fix, 2 sessions after
- Before: checkout fails, agent recovers ad-hoc by guessing the right approach. Branch deletion fails, agent retries in wrong order. False warning about discarding merged commits.
- After: merge happens cleanly from original repo, cleanup in correct order, no false warnings.

## Rigor

- [ ] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) without extensive evals showing the change is an improvement

**Note on writing-skills:** Formal pressure testing not yet completed. The fix addresses concrete, reproducible git failures, but the new Option 1 instructions do guide agent behavior (detect worktree, merge from elsewhere). We plan to run pressure testing and update before requesting final review.

**Note on behavior-shaping content:** Option 1 now has a branching decision (in worktree vs not). This was validated through functional testing (observed the git failures, verified the fix works), not formal evals. The non-worktree path is unchanged.

## Human review

### DIFF
--- a/skills/finishing-a-development-branch/SKILL.md
+++ b/skills/finishing-a-development-branch/SKILL.md
@@ -67,6 +67,29 @@ Which option?
 
 #### Option 1: Merge Locally
 
+**If in a worktree** (i.e., `git rev-parse --git-dir` differs from `git rev-parse --git-common-dir`):
+
+The base branch is already checked out in the original repo — you cannot `git checkout` it from the worktree. Instead, find the original repo and merge from there:
+
+```bash
+# Find the main worktree (original repo)
+main_worktree=$(git worktree list --porcelain | awk '/^worktree / {print $2; exit}')
+
+# Pull latest in the original repo
+git -C "$main_worktree" pull
+
+# Merge feature branch from the original repo
+git -C "$main_worktree" merge <feature-branch>
+
+# Verify tests on merged result (must run from original repo for deps/config)
+cd "$main_worktree"
+<test command>
+```
+
+Then: Cleanup worktree (Step 5), which will also delete the branch.
+
+**If NOT in a worktree** (normal branch):
+
 ```bash
 # Switch to base branch
 git checkout <base-branch>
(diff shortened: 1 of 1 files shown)


csillag:
I have updated the PR description to match the prescribed template. (Forry for missing it the first time.)

csillag:
Closing this in favor of #1072, which is a better fix — it addresses the branch deletion ordering plus self-healing for stale registrations, the buggy grep that matches the wrong worktree, and Windows filesystem locks. I missed those three issues in my version.

PR #1006, opened 2026-03-30
feat: add env var to skip SessionStart context injection

### LINKED ISSUE #951: Switch Button: Support superpower on/off in CLI
Currently, enabling or disabling Superpowers in Codex CLI requires manually adding or removing the `~/.agents/skills/superpowers` symlink and restarting the CLI. This is cumbersome and slows down workflows when switching between simple tasks and complex, multi-agent workflows.

## Proposed solution
Introduce a CLI switch (e.g., `--superpowers on/off`) or command (e.g., `codex superpowers enable/disable`) to toggle Superpowers dynamically without manual file operations. This would streamline switching between lightweight and full-featured modes.

Manual symlink management and maintaining separate environments or scripts. These approaches are error-prone, not user-friendly, and interrupt workflow due to required restarts.

Yes. This feature improves usability for all users by making Superpowers more flexible and easier to adopt across different tasks and workflows.

## Context
Codex CLI with Superpowers installed via `~/.agents/skills/`. Frequent switching between quick edits and complex planning/TDD workflows.

cionelo:
## Problem

Closes #951.

The `hooks/session-start` script unconditionally injects ~1,500 tokens of `SKILL.md` content on every session. There is no mechanism to opt out. For power users running 20+ sessions/day this costs ~30,000 tokens/day in context that arrives before the user's first message — tokens that pay for content already implicitly available via the `Skill` tool.

## Solution

Add a single env var guard immediately after `set -euo pipefail`:

```bash
# Allow users to skip session context injection via env var
if [ "${SUPERPOWERS_SKIP_SESSION_CONTEXT:-}" = "true" ]; then
    exit 0
fi
```

Setting `SUPERPOWERS_SKIP_SESSION_CONTEXT=true` (e.g. in `~/.claude/settings.json` → `env`) causes the hook to exit cleanly with no output, skipping all context injection. All skills remain fully available via the explicit `Skill` tool invocation — nothing is removed from the plugin, just the automatic injection.

**Default behaviour is unchanged** — the env var defaults to unset, so existing users see no difference.

## Usage

In `~/.claude/settings.json`:
```json
{
  "env": {
    "SUPERPOWERS_SKIP_SESSION_CONTEXT": "true"
  }
}
```

Or export in your shell profile:
```bash
export SUPERPOWERS_SKIP_SESSION_CONTEXT=true
```

### DIFF
--- a/hooks/session-start
+++ b/hooks/session-start
@@ -3,6 +3,11 @@
 
 set -euo pipefail
 
+# Allow users to skip session context injection via env var
+if [ "${SUPERPOWERS_SKIP_SESSION_CONTEXT:-}" = "true" ]; then
+    exit 0
+fi
+
 # Determine plugin root directory
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 PLUGIN_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"


obra:
This is a non-starter for us. The bootstrap is what makes superpowers effective. Also, you 100% ignored the PR template, which suggests that this was an agentic drive-by PR

BenHamm:
> This is a non-starter for us. The bootstrap is what makes superpowers effective. Also, you 100% ignored the PR template, which suggests that this was an agentic drive-by PR

In my experience with Claude Code, if I lose internet connection, it will restart the superpowers prompt.  The problem here is that i've already done what I need to do with superpowers in this session. I don't want to lose context, but i don't want to keep paying the token cost of the bootstrap getting passed in. A toggle would be helpful for this situation.

`API Error: Unable to connect to API (ConnectionRefused)


Initialized your session
Initialized your session
Ran a hook on session startup
{
  "hookSpecificOutput": {
Show less
    "additionalContext": "<EXTREMELY_IMPORTANT>\nYou have superpowers.\n\n**Below is the full content of your 'superpowers:using-superpowers' skill - your introduction to using skills. For all other skills, use the 'Skill' tool:**\n\n---\nname: using-superpowers\ndescription: Use when starting any conversation - establishes how to find and use skills, requiring Skill tool invocation before ANY response including clarifying questions\n---\n\n<SUBAGENT-STOP>\nIf you were dispatched as a subagent to execute a specific task, skip this skill.\n</SUBAGENT-STOP>\n\n<EXTREMELY-IMPORTANT>\nIf you think there is even a 1% chance a skill might apply to what you are doing, you ABSOLUTELY MUST invoke the skill.`

PR #1019, opened 2026-03-31
Trim using-superpowers startup injection and remove deprecated commands

Granite-Lorax:
## Summary

- **SKILL.md**: 5,219 → 2,289 bytes (56% reduction) — this content is injected into every session via the SessionStart hook, so its size directly taxes startup context
- **Deprecated commands removed**: `brainstorm.md`, `execute-plan.md`, `write-plan.md` — these just tell users to use the renamed skills and add noise to the skill list

## What was cut

- Dot graph (`digraph skill_flow`) — not rendered in any supported platform, just raw graphviz syntax in the prompt
- 12-row "Red Flags" table — repetitive (every row says "check for skills"); collapsed into a single paragraph that preserves the anti-rationalization intent
- Redundant `EXTREMELY-IMPORTANT` emphasis wrappers — the rule is stated clearly without them

## What was preserved

- All platform references (Claude Code, Copilot CLI, Gemini CLI, Codex)
- Instruction priority ordering
- Skill priority and type guidance
- Anti-rationalization guidance (condensed)
- Subagent stop gate
- User instructions override rule

## Motivation

Startup context weight matters — every token in the SessionStart injection competes with the user's actual task. The removed content was either decorative (dot graph), redundant (12 variations of "check for skills"), or vestigial (deprecated command stubs).

## Test plan

- [ ] `/clear` and verify trimmed content appears in SessionStart injection
- [ ] Invoke a skill via `Skill` tool — confirm normal behavior
- [ ] Verify deprecated commands no longer appear in skill list

### DIFF
--- a/commands/brainstorm.md
+++ b/commands/brainstorm.md
@@ -1,5 +1,0 @@
----
-description: "Deprecated - use the superpowers:brainstorming skill instead"
----
-
-Tell your human partner that this command is deprecated and will be removed in the next major release. They should ask you to use the "superpowers brainstorming" skill instead.
--- a/commands/execute-plan.md
+++ b/commands/execute-plan.md
@@ -1,5 +1,0 @@
----
-description: "Deprecated - use the superpowers:executing-plans skill instead"
----
-
-Tell your human partner that this command is deprecated and will be removed in the next major release. They should ask you to use the "superpowers executing-plans" skill instead.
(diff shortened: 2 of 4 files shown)


obra:
This looks like slop. Please don't submit untested slop rewrites that ignore everything about how superpowers works and ignore the PR template.

PR #1028, opened 2026-04-01
docs(writing-skills): add YAML frontmatter quoting guidance

junhinhow:

While creating a custom skill with a description containing a colon followed by a space (e.g. `description: Use when configuring X for production: tuning and sizing`), the skill was silently ignored by `npx skills add`. No error message was shown — the skill simply didn't appear in the list. After debugging, we found that YAML parsed the `: ` as a key-value separator instead of a string continuation.

The `writing-skills` SKILL.md shows multiple `description` examples but none are quoted, and there is no warning about this YAML parsing behavior. This is a silent failure that is extremely difficult to diagnose.

## What does this PR change?

Adds explicit guidance in the writing-skills SKILL.md that `description` values must always be wrapped in double quotes. Updates all existing description examples in the template and content sections to use quotes consistently.

Yes — this affects every user who writes a custom skill. The YAML quoting issue is not domain-specific; it's a universal gotcha in the skill authoring workflow. The fix is purely documentation — no behavior change.

1. **Fix the YAML parser to be more lenient** — rejected because the parser follows the YAML spec correctly; the issue is user expectation.
2. **Add a CLI warning when parsing fails** — that would be a code change in a different repo (the skills CLI). This docs fix is complementary and immediately useful.
3. **Only add a note without changing examples** — rejected because developers copy-paste examples; if the examples aren't quoted, the note gets ignored.

No. All changes are in a single file (writing-skills/SKILL.md) and address a single issue: missing YAML quoting guidance for the `description` frontmatter field.

## Existing PRs
- Related PRs: none found addressing YAML frontmatter quoting

## Environment tested

| Claude Code | 2.1.87 | Claude Opus 4.6 | claude-opus-4-6 |

## Evaluation
- Initial prompt: "Create a custom skill with a description containing a colon" — skill was silently invisible
- After adding quotes to the description, `npx skills add` detected the skill correctly
- Tested with 3 different description strings containing `: ` — all failed unquoted, all succeeded quoted
- The documentation change makes the correct pattern obvious from the examples

## Rigor

This is a documentation-only change. The existing skill behavior, triggers, Red Flags table, and all agent-behavior-shaping content are untouched. Only example code blocks and a new guidance paragraph were added.

Adversarial test: Intentionally created skills with descriptions containing `: `, `- `, `# `, and `{ }` characters. Only `: ` caused silent failure. The added guidance specifically targets this case.

## Human review

### DIFF
--- a/skills/writing-skills/SKILL.md
+++ b/skills/writing-skills/SKILL.md
@@ -105,7 +105,7 @@ skills/
 ```markdown
 ---
 name: Skill-Name-With-Hyphens
-description: Use when [specific triggering conditions and symptoms]
+description: "Use when [specific triggering conditions and symptoms]"
 ---
 
 # Skill Name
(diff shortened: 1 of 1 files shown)


obra:
I suspect that this could be reliably solved with a single line in writing skills saying that YAML front matter needs to be quoted in cases when the YAML parser would ordinarily expect it. Are you up for trying that?

PR #1037, opened 2026-04-02
fix(brainstorming): HARD-GATE visual companion per-question decision

grgbrasil:
## Problem

The visual companion guidelines already say to use the terminal for text content (tradeoff lists, A/B choices, clarifying questions), but LLMs routinely ignore this soft guidance and render HTML pages for content that is purely textual — wasting tokens and user time.

Real-world example: when brainstorming a TOTP wizard redesign, the LLM rendered a full HTML page just to show two text-based approach descriptions with pros/cons. This could have been a simple terminal message.

## Fix

Wraps the per-question decision section in a `<HARD-GATE>` tag (already used elsewhere in the skill for the no-implementation gate), adds a verbalization requirement before each browser write, and strengthens the language with ONLY/NEVER qualifiers.

The key addition is requiring the LLM to answer out loud: *"Am I showing a visual artifact that cannot be communicated as text?"* — forcing a conscious decision before each browser write.

## Changes

- `skills/brainstorming/SKILL.md` — 1 section changed (6 lines added, 4 removed)

### DIFF
--- a/skills/brainstorming/SKILL.md
+++ b/skills/brainstorming/SKILL.md
@@ -153,12 +153,14 @@ A browser-based companion for showing mockups, diagrams, and visual options duri
 
 **This offer MUST be its own message.** Do not combine it with clarifying questions, context summaries, or any other content. The message should contain ONLY the offer above and nothing else. Wait for the user's response before continuing. If they decline, proceed with text-only brainstorming.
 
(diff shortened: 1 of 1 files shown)


obra:
"Can not be communicated as text" is the wrong bar, I think. "Is *better* communicated as text, maybe.  But I'd really need to see the *actual* output and more than a single example. Also, which model, which agent, etc. You deleted the entire pull request template, which is required for us to review

PR #1047, opened 2026-04-04
fix(using-superpowers): use fully qualified skill name in flow diagram

### LINKED ISSUE #1002: using-superpowers SKILL.md flow diagram causes intermittent "Unknown skill: brainstorm" error
## Description

The `using-superpowers` skill's flow diagram references "Invoke brainstorming skill" as plain text without specifying the fully qualified skill name. This causes Claude to intermittently invoke `Skill("brainstorm")` or `Skill("brainstorming")` instead of `Skill("superpowers:brainstorming")`, resulting in an "Unknown skill" error.

## Reproduction

1. Start a new Claude Code session with superpowers plugin enabled
2. Enter plan mode (or trigger the brainstorming flow)
3. Claude reads the flow diagram in `skills/using-superpowers/SKILL.md`:
   ```dot
   "Already brainstormed?" -> "Invoke brainstorming skill" [label="no"];
   ```
4. Claude interprets "Invoke brainstorming skill" and sometimes calls:
   - `Skill("brainstorm")` → **Error: Unknown skill**
   - `Skill("brainstorming")` → **Error: Unknown skill**
   - `Skill("superpowers:brainstorming")` → **Success** (intermittent)

## Root Cause

1. **Ambiguous instruction**: The flow diagram says `"Invoke brainstorming skill"` without the `superpowers:` namespace prefix
2. **Name collision**: The deprecated `commands/brainstorm.md` file still exists, adding confusion between `brainstorm` (command) and `brainstorming` (skill)
3. **Non-deterministic resolution**: Claude's skill name inference varies between invocations, making the error intermittent

## Suggested Fix

In `skills/using-superpowers/SKILL.md`, change the flow diagram node from:
```dot
"Invoke brainstorming skill" [shape=box];
```
to:
```dot
"Invoke Skill('superpowers:brainstorming')" [shape=box];
```

And update the edge reference accordingly. This makes the fully qualified name explicit and eliminates ambiguity.

Additionally, consider removing the deprecated command files (`commands/brainstorm.md`, `commands/execute-plan.md`, `commands/write-plan.md`) to reduce name collision surface.

## Environment

- superpowers v5.0.6
- Claude Code CLI
- macOS

mvanhorn:

The flow diagram in `skills/using-superpowers/SKILL.md` uses "Invoke brainstorming skill" as a node label. When agents read this diagram, they extract the label text and try to invoke `Skill("brainstorm")` or `Skill("brainstorming")` instead of the fully qualified `Skill("superpowers:brainstorming")`. This produces "Unknown skill" errors intermittently during plan mode entry.

The bug is at `skills/using-superpowers/SKILL.md` lines 53, 63, and 65, where the graphviz node labels use the short name.

## What does this PR change?

Replaces the three "Invoke brainstorming skill" references in the flow diagram with "Invoke superpowers:brainstorming" so agents extract the correct fully qualified skill name.

Yes. This fixes a bug in the core `using-superpowers` skill that affects all users on all platforms. The skill is loaded at session start for every conversation.

1. Adding a separate instruction line below the diagram telling agents to use the qualified name. Rejected because agents read the diagram labels directly, so fixing the label is more reliable than adding a second instruction that might be ignored.
2. Changing only the prose references at lines 101 and 104. Rejected because those lines describe the concept of brainstorming, not the invocation. The actual bug is in the diagram node labels that agents parse as skill names.

No. All three changes are in the same flow diagram, fixing the same bug (unqualified skill name).

## Existing PRs
- Related PRs: none found. Searched open and closed PRs for "brainstorming", "skill name", "qualified", "SKILL.md". No prior attempts to fix this.

## Environment tested

| Claude Code                         | 1.0.33          | Opus  | claude-opus-4-6  |

## Evaluation

- Initial prompt: "/osc-plan Superpowers" which led to discovering this issue via automated scanning
- After making the change, I verified the three lines now contain "superpowers:brainstorming" instead of "brainstorming skill"
- The fix is a label text correction in a graphviz diagram. The agent reads these labels to decide what skill to invoke. With the old labels, the agent sees "Invoke brainstorming skill" and tries `Skill("brainstorming")`. With the new labels, it sees "Invoke superpowers:brainstorming" and tries `Skill("superpowers:brainstorming")`.

## Demo

Demo

## Rigor

- [x] If this is a skills change: the change is limited to node labels in the graphviz diagram. No behavior-shaping prose, Red Flags table, or rationalization lists were modified.

## Human review

Fixes #1002

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/skills/using-superpowers/SKILL.md
+++ b/skills/using-superpowers/SKILL.md
@@ -50,7 +50,7 @@ digraph skill_flow {
     "User message received" [shape=doublecircle];
     "About to EnterPlanMode?" [shape=doublecircle];
     "Already brainstormed?" [shape=diamond];
-    "Invoke superpowers:brainstorming" [shape=box];
+    "Invoke Skill('superpowers:brainstorming')" [shape=box];
     "Might any skill apply?" [shape=diamond];
     "Invoke Skill tool" [shape=box];
     "Announce: 'Using [skill] to [purpose]'" [shape=box];
@@ -60,9 +60,9 @@ digraph skill_flow {
     "Respond (including clarifications)" [shape=doublecircle];
 
     "About to EnterPlanMode?" -> "Already brainstormed?";
-    "Already brainstormed?" -> "Invoke superpowers:brainstorming" [label="no"];
+    "Already brainstormed?" -> "Invoke Skill('superpowers:brainstorming')" [label="no"];
     "Already brainstormed?" -> "Might any skill apply?" [label="yes"];
-    "Invoke superpowers:brainstorming" -> "Might any skill apply?";
+    "Invoke Skill('superpowers:brainstorming')" -> "Might any skill apply?";
 
     "User message received" -> "Might any skill apply?";
     "Might any skill apply?" -> "Invoke Skill tool" [label="yes, even 1%"];
--- a/skills/using-superpowers/SKILL.md
+++ b/skills/using-superpowers/SKILL.md
@@ -50,7 +50,7 @@ digraph skill_flow {
     "User message received" [shape=doublecircle];
     "About to EnterPlanMode?" [shape=doublecircle];
     "Already brainstormed?" [shape=diamond];
-    "Invoke brainstorming skill" [shape=box];
+    "Invoke superpowers:brainstorming" [shape=box];
     "Might any skill apply?" [shape=diamond];
     "Invoke Skill tool" [shape=box];
     "Announce: 'Using [skill] to [purpose]'" [shape=box];
@@ -60,9 +60,9 @@ digraph skill_flow {
     "Respond (including clarifications)" [shape=doublecircle];
 
     "About to EnterPlanMode?" -> "Already brainstormed?";
-    "Already brainstormed?" -> "Invoke brainstorming skill" [label="no"];
+    "Already brainstormed?" -> "Invoke superpowers:brainstorming" [label="no"];
     "Already brainstormed?" -> "Might any skill apply?" [label="yes"];
-    "Invoke brainstorming skill" -> "Might any skill apply?";
+    "Invoke superpowers:brainstorming" -> "Might any skill apply?";
 
     "User message received" -> "Might any skill apply?";
     "Might any skill apply?" -> "Invoke Skill tool" [label="yes, even 1%"];


jplamy7-lab (approved):

duarbdhks (approved):

arthurkok2:
**Bottom line: The underlying bug is real and confirmed. The fix is directionally correct, but there are a few things worth calling out.**

   ---

   ### The bug is real

   The referenced issue #1002 (filed by a separate, independent user) describes exactly what the PR claims: `"Invoke brainstorming skill"` in the graphviz flow diagram causes agents to intermittently call `Skill("brainstorm")` or `Skill("brainstorming")` — both of which fail with "Unknown skill." This is verifiable by reading the current `SKILL.md` at line 53, 63, and 65. No prior PRs attempt this fix.

   ### The fix is correct in direction, but slightly underbaked

   The PR changes the labels to `"Invoke superpowers:brainstorming"`. That's better than the original, but **issue #1002 itself suggested the more explicit form**: `"Invoke Skill('superpowers:brainstorming')"`. The PR author chose a shorter label, but the issue's suggestion more directly maps to the actual API call syntax an agent would use. The PR's version still relies on the agent inferring that `superpowers:brainstorming` is the argument to `Skill()`. Not a dealbreaker, but worth discussion.

   ### Eval section is thin

   The author confirms the labels now read correctly after the change — but doesn't show:
   - A session transcript demonstrating the bug before the fix
   - A session after the fix where the agent correctly invokes `Skill("superpowers:brainstorming")`

   "I verified the three lines now contain 'superpowers:brainstorming'" is a diff check, not a behavioral test. CLAUDE.md expects adversarial testing.

   ### AI-assistance + "automated scanning" disclosure

   The PR discloses it was developed with Codex and discovered via `/osc-plan Superpowers` automated scanning. This matches the CLAUDE.md warning about "spray-and-pray" contributions. That said, the bug was independently filed by a different user (#1002, zero reactions, March 30), so the root cause is real — not fabricated.

   ### PR template is otherwise well-filled

   Alternatives are documented, single-file change, no duplicate PRs, human review confirmed, not a skill prose/behavior change.

   ---

   ### Verdict

   The bug is genuine, the fix is minimal and scoped correctly, and the template is solid.  Two things a reviewer should push on:

   1. **Should the label be `"Invoke Skill('superpowers:brainstorming')"` instead?** More explicit, directly matches the call syntax.
   2. **Can the author provide a session trace showing the bug and its resolution?** Even a short one.

   Likely mergeable with those questions answered — this is a legitimate fix to a real issue.

mvanhorn:
Thanks for the thorough review. Addressed both points:

1. **Label format**: Updated in 5e07691 to use `"Invoke Skill('superpowers:brainstorming')"` as suggested in #1002. This directly maps to the API call syntax.

2. **Session trace**: The labels are inside a graphviz `dot` diagram that agents reference when deciding which skill to invoke. The failure mode (agents calling `Skill("brainstorm")` or `Skill("brainstorming")`) is intermittent and depends on how the agent parses the flow diagram. The before/after difference is visible in the diff: old labels said `"Invoke superpowers:brainstorming"` (ambiguous), new labels say `"Invoke Skill('superpowers:brainstorming')"` (unambiguous call syntax). I don't have a reliable way to reproduce the intermittent parsing failure on demand, but the fix removes the ambiguity that causes it.

arittr:
Thanks for the PR. We investigated #1002 with Drill against the reported v5.0.6 plugin state and current dev, including normal Claude mode and Claude plan mode.

We could not reproduce the reported stale skill-name failure: no run produced Skill("brainstorm"), Skill("brainstorming"), or an Unknown skill: brainstorm... error. We did see occasional plan-mode runs skip skill loading entirely, but this PR’s wording change did not produce a clear red/green result for that behavior either.

Since we don’t have evidence that this change fixes the reported failure mode, we’re closing this PR for now. If someone can provide a transcript or tool-call log showing the stale Skill(...) invocation, we can reopen the issue and build a targeted Drill regression around that path.

PR #1057, opened 2026-04-04
fix(subagent-driven-development): remove hardcoded task-count example

### LINKED ISSUE #1058: subagent-driven-development may anchor on 5 tasks from example workflow
Hi! Basically for some reason whenever I run subagents it always stops at task 5 and I need to manually tell it to continue which is slightly annoying.

It might be because of the example? Not sure why else it stops at 5.

I noticed the `subagent-driven-development` example workflow currently includes this line:

- `Extract all 5 tasks with full text and context`

That seems like it could anchor the model to five tasks even when the plan contains more.

I opened a PR with a minimal wording fix here:
- #1057

If this is intentional, I may be misunderstanding the workflow, but from the docs/release notes it looks like the skill is supposed to execute all tasks from the plan rather than stop after five.

I'm also running it with Codex instead of Claude so maybe just a Codex thing

DicksonWu654:
## Summary

Removes a hardcoded numeric anchor from the `subagent-driven-development` example workflow.

Before:
- `Extract all 5 tasks with full text and context`

After:
- `Extract all tasks with full text and context`

Fixes #1058.

## Why

Hi! Basically for some reason whenever I run subagents it always stops at task 5 and I need to manually tell it to continue which is slightly annoying.

It might be because of the example? Not sure why else it stops at 5.

This does not change workflow semantics; it just makes the example task-count agnostic and aligns it with the surrounding guidance about extracting all tasks from the plan.

### DIFF
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -129,7 +129,7 @@ Implementer subagents report one of four statuses. Handle each appropriately:
 You: I'm using Subagent-Driven Development to execute this plan.
 
 [Read plan file once: docs/superpowers/plans/feature-plan.md]
-[Extract all 5 tasks with full text and context]
+[Extract all tasks with full text and context]
 [Create TodoWrite with all tasks]
 
 Task 1: Hook installation script


abhishekkushwaha-fareye (approved):

obra:
Thanks @DicksonWu654 — anchoring on a hardcoded "5 tasks" example was a real footgun, and you were right to remove it.

After the 6.0 release: **6.0.0 rewrote subagent-driven-development and the hardcoded task-count anchor is gone.** Verified against v6.0.2 — `skills/subagent-driven-development/SKILL.md` no longer contains "all 5" / "5 tasks" / "extract all N tasks"; the example is now task-count-agnostic ("create todos for all tasks"). The specific line your PR removes no longer exists.

So I'm closing this as resolved in 6.0. If you still see the controller anchoring on a fixed task count on 6.0.2, please reopen or file a fresh issue with the transcript (that'd be a behavioral report worth having).

---

PR #1063, opened 2026-04-05
fix: make using-git-worktrees skill platform-neutral

### LINKED ISSUE #1049: Make `using-git-worktrees` platform-neutral instead of Claude-specific
- [x ] I searched existing issues and this has not been proposed before

Superpowers is presented as a multi-platform workflow with installation paths for Claude Code, Cursor, Codex, OpenCode, GitHub Copilot CLI, and Gemini CLI.

However, the `using-git-worktrees` skill currently includes a hardcoded step to check `CLAUDE.md` when deciding where worktrees should be created. That makes the skill feel Claude-specific even though it appears in the core workflow and skills library as a general-purpose skill.

This can be confusing for users on non-Claude platforms, because it is not obvious whether they should:

* ignore that step,
* create a fake `CLAUDE.md`,
* or expect an equivalent platform-specific file to be checked instead.

## Proposed solution
Make `using-git-worktrees` platform-neutral.

Possible approaches:

1. Replace the `CLAUDE.md` step with “check the platform’s agent-instruction file if one exists”.
2. Check multiple known instruction files in priority order (for example `CLAUDE.md`, `AGENTS.md`, or other platform-specific equivalents).

* Leaving the skill as-is and documenting that `CLAUDE.md` is only an example.

Yes. `using-git-worktrees` is listed as part of the basic workflow, so it should ideally read as a core, cross-platform skill unless explicitly marked otherwise.

## Context
I’m raising this because the current wording gives the impression that this skill was originally designed around Claude Code assumptions, while the repository itself now targets multiple coding agents and platforms.

funsaized:
## Summary

Fixes https://github.com/obra/superpowers/issues/1049

The `using-git-worktrees` skill had hardcoded `CLAUDE.md` references for its directory selection process. Since Superpowers targets multiple platforms (Claude Code, Cursor, Codex, OpenCode, Copilot CLI, Gemini CLI), the skill should be platform-neutral.

## Changes

- Replaced `### 2. Check CLAUDE.md` section header with `### 2. Check Agent Configuration`
- Updated the grep command to check multiple agent configuration files in priority order (`CLAUDE.md`, `AGENTS.md`, `GEMINI.md`, `.cursorrules`, `CODEX.md`) instead of only `CLAUDE.md`
- Updated all downstream references (quick reference table, common mistakes, red flags) from `CLAUDE.md` to the generic `agent config` / `agent configuration` terminology
- The only remaining `CLAUDE.md` reference is inside the bash code block where it's listed as one of the config files to check

### DIFF
--- a/skills/using-git-worktrees/SKILL.md
+++ b/skills/using-git-worktrees/SKILL.md
@@ -27,17 +27,22 @@ ls -d worktrees 2>/dev/null      # Alternative
 
 **If found:** Use that directory. If both exist, `.worktrees` wins.
 
-### 2. Check CLAUDE.md
+### 2. Check Agent Configuration
+
+Check the platform's agent-instruction file for a worktree directory preference:
 
 ```bash
-grep -i "worktree.*director" CLAUDE.md 2>/dev/null
+# Check known agent configuration files in priority order
+for config in CLAUDE.md AGENTS.md GEMINI.md .cursorrules CODEX.md; do
+  grep -i "worktree.*director" "$config" 2>/dev/null && break
+done
 ```
 
 **If preference specified:** Use it without asking.
 
 ### 3. Ask User
 
-If no directory exists and no CLAUDE.md preference:
+If no directory exists and no agent-configuration preference:
 
 ```
 No worktree directory found. Where should I create worktrees?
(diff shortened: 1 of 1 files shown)


skills/using-git-worktrees/SKILL.md
```diff
@@ -27,17 +27,22 @@ ls -d worktrees 2>/dev/null      # Alternative
...
 
-### 2. Check CLAUDE.md
+### 2. Check Agent Configuration
+
+Check the platform's agent-instruction file for a worktree directory preference:
 
 ```bash
-grep -i "worktree.*director" CLAUDE.md 2>/dev/null
+# Check known agent configuration files in priority order
+for config in CLAUDE.md AGENTS.md GEMINI.md .cursorrules CODEX.md; do
```
oscarbotteri, line 36: would it be useful to add `copilot-instructions.md`?

arittr:
Thanks for the PR. This has been handled by #1121 on `dev`, and the linked issue #1049 has already been closed as fixed.

The current `using-git-worktrees` wording is platform-neutral and no longer treats `CLAUDE.md` as the only instruction-file source.

Closing this as superseded by the merged worktree rewrite.

Issue #1067, opened 2026-04-06
`using-superpowers` skill is injected in full at session start — consider compact trigger injection


al-boyle:
The session-start hook currently injects the full content of `using-superpowers/SKILL.md` (~5,400 bytes, ~1,350 tokens) into every session. The bootstrapping reason for this is sound — the model needs to know to invoke skills before it knows to look for them.

However, the full skill content includes reference material (red flags table, skill priority rules, flow diagram) that's only useful mid-session, not at startup. The load-bearing instruction is much smaller:

> Before any response, check if any available skill applies. If yes, invoke it via the Skill tool before responding.

**Proposal:** inject a compact trigger rule at session start (~100–200 tokens), and make the full `using-superpowers` content available as a deferrable skill for mid-session reference. This would reduce fixed per-session overhead by ~1,100 tokens with no functional change to skill invocation behavior.

obra:
The existing skill is heavily evalled to work reliably and is the core of what makes superpowers actually effective.

Issue #1061, opened 2026-04-05
Task Triage before launching full workflow


CallMarcus:
- [ ] I searched existing issues and this has not been proposed before

## What problem does this solve?
 I would like to have a triage step if the full workflow is needed for minor changes,

## Proposed solution
 Superpowers firsts assess the task at hands before launching the full workflow. For small 1-3 file edits it would propose a lighter workflow.

## What alternatives did you consider?
 I have a plan that can save time and a ton of tokens.

[task-triage-SKILL.md](https://github.com/user-attachments/files/26487013/task-triage-SKILL.md)

## Is this appropriate for core Superpowers?
 this is a generic demand.

CallMarcus:
I have the option of forking the repo but I believe this feature or demand is rather common; a small tweak currently might result in a 30-min session with over usage to complete.

arittr:
Thanks — matching workflow weight to task size is a real need. A first pass landed a while back (PR #522, scale-to-complexity guidance), and the live discussion on going further — lightening review for easy tasks without letting models shirk — is #1120. Consolidating there; closing as a duplicate.

PR #1071, opened 2026-04-06
feat: add router-first default workflow

Coder4:
## Summary
- add a new `router-first` skill as the default lightweight entrypoint
- document `router-first` as the default mode and `using-superpowers` as the heavy workflow
- switch the Claude session bootstrap and Codex docs to the new default entrypoint

## Motivation
The current default Superpowers flow is excellent for high-risk work, but it can feel too heavy for many day-to-day tasks in Codex and Claude. This change introduces a lighter default entrypoint so simple work can move faster without removing the existing heavy workflow.

## Behavior change
- new default entrypoint: `router-first`
- `using-superpowers` remains available as the full heavy workflow
- Claude session bootstrap now injects `router-first` by default
- Codex docs now describe `router-first` as the default mode

## Details
This keeps the original heavy Superpowers flow available, but makes the default behavior lighter:
- `small`: inspect, act, verify
- `medium`: share a short plan, then execute
- `large`: explain risk and ask before switching into the heavy workflow

It also preserves `verification-before-completion` as the always-on safeguard and keeps `systematic-debugging` as a conditional guardrail for debugging work.

## Compatibility
This does not remove the original heavy workflow. It preserves the existing planning-and-governance path and only changes the default entry behavior.

## Verification
- checked skill, docs, and bootstrap consistency
- ran `git diff --check`
- new Claude/Codex sessions are required for the new default mode to take effect

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,21 @@
 # Superpowers
 
-Superpowers is a complete software development workflow for your coding agents, built on top of a set of composable "skills" and some initial instructions that make sure your agent uses them.
+Superpowers is a software development workflow for coding agents, built on a set of composable skills and startup instructions that help the agent choose the right level of process for the task.
 
 ## How it works
 
-It starts from the moment you fire up your coding agent. As soon as it sees that you're building something, it *doesn't* just jump into trying to write code. Instead, it steps back and asks you what you're really trying to do. 
+By default, Superpowers now starts in a lightweight `router-first` mode. Instead of forcing every task through the full planning stack, the agent classifies the work and chooses the lightest safe path:
 
-Once it's teased a spec out of the conversation, it shows it to you in chunks short enough to actually read and digest. 
+- `small` -> inspect quickly, act directly, verify before claiming success
+- `medium` -> share a short execution plan, then execute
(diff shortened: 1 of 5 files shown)


Coder4:
I think this PR is closely related to a couple of existing discussions:

- #1067: `using-superpowers` skill is injected in full at session start - consider compact trigger injection
- #1061: Task Triage before launching full workflow

This PR is my attempt at combining those concerns into a concrete default path:
- use a lighter `router-first` entrypoint by default
- keep the existing heavy workflow intact
- only escalate into the full process when risk or ambiguity justifies it

So I'd frame this less as "replace Superpowers" and more as "change the default entry behavior while preserving the current heavy path."

obra:
We've previously played with this approach and it seems to result in generally worse outcomes. Also you ignored the PR template, which tells me that you haven't run evals

Coder4:
Thanks for the quick feedback - and for pointing out that this approach has already been tried internally.

You're also right that I didn't follow the PR template closely enough, and I didn't include eval evidence here. That's on me.

I appreciate the context on prior experiments. I'll take a closer look at the existing PR template and eval expectations before proposing anything further in this area, and I'll treat this PR as a learning pass rather than something ready to merge.

PR #1072, opened 2026-04-06
fix(finishing-a-development-branch): order branch deletion after worktree removal

mtschoen:

Cleaning up an accumulated backlog of 12 stale `worktree-agent-*` branches on a Windows 11 host (Claude Code 2.1.119, Opus 4.7) revealed three silent-failure bugs in `skills/finishing-a-development-branch/SKILL.md` that had been letting branches leak across many sessions:

1. In Options 1 ("Merge Locally") and 4 ("Discard"), `git branch -d/-D <feature>` ran *before* `git worktree remove`. Git refuses to delete a branch that's checked out by a worktree, so the delete printed `error: cannot delete branch '<feature>' used by worktree at '<path>'` and the skill continued. Net result: worktree gets removed, branch is left behind, no error surfaced to the user. Repeats every session that hits Option 1 or 4.
2. Some worktree directories had been deleted out-of-band (interrupted prior sessions, manual cleanup attempts). `git worktree remove <path>` errors out (`fatal: '<path>' is not a working tree`) and the registration in `.git/worktrees/` lingers indefinitely. The lingering registration also blocks future `git branch -d` of the same branch even after the dir is gone — which is how single-session bugs accumulate into multi-branch backlogs.
3. Step 4's "find the worktree to remove" used `git worktree list | grep $(git branch --show-current)`. By Step 4 the feature branch has already been checked out away from (the skill ran `git checkout <base-branch>` earlier), so `--show-current` returns the **base branch**, and the grep matches the wrong worktree. It's also substring-fragile — a base-branch name appearing as a substring inside another worktree's path silently mismatches.

## What does this PR change?

Single file: `skills/finishing-a-development-branch/SKILL.md`.

- Removes the inline `git branch -d/-D <feature>` calls from Options 1 and 4. Consolidates branch deletion into Step 5, which now runs *after* `git worktree remove`.
- Adds `git worktree prune -v` to Step 5: as a fallback when `git worktree remove` fails because the directory is already gone, and as a defensive sweep at the end.
- Replaces `git worktree list | grep $(git branch --show-current)` with an `awk` match on column 3 of `git worktree list`, anchored to `[<feature-branch>]`.
- Adds a Windows-specific note: the skill must be invoked from the parent context (main repo), not from inside the worktree being cleaned, because Windows holds a filesystem lock on a process's own cwd.

Yes. `finishing-a-development-branch` is a general-purpose skill every user invokes at the end of feature work; the bugs are platform-independent (only the documentation note is Windows-specific) and reproduce on any host where the skill has been invoked from inside a worktree.

- **Detect the silent failure and surface it** instead of reordering. Rejected: even with detection the user is left mid-cleanup with a stale branch and no recovery instructions. Reordering removes the failure entirely.
- **Skip the `prune -v` fallback** and document manual cleanup of `.git/worktrees/` instead. Rejected: stale registrations block branch deletion silently, so users would only discover the problem after the *next* cleanup also failed. The fallback is one line and self-healing.
- **Use `git for-each-ref` plumbing** instead of awk on `git worktree list`. Rejected as more code for the same correctness — the awk match on column 3 is short, anchored to bracketed branch names that git itself produces, and matches the existing skill's shell-pipeline style.

No. All three sub-fixes apply to the same skill (`finishing-a-development-branch`), to the same Step 4–Step 5 cleanup flow, and to the same root failure mode encountered in the same cleanup session that motivated the PR (12 stale branches whose deletion required all three fixes to succeed). They ship together because each fix on its own would have left another bug still leaking branches.

## Existing PRs
- Related PRs: #391, #913, #952, #1103 (closed), #1219, #1223

**Differentiation:**

- **#1103 (CLOSED, no review comments, 2026-04-09)** — addressed the same Bug 1 (branch-delete-before-worktree-remove) plus an unrelated Option 2 cleanup-table change. Closed silently with no maintainer feedback, so the technical fix was not rejected on merit; the most likely cause was template non-compliance (the body used "Summary / Test plan" rather than the required template). This PR (a) follows the template, (b) drops the unrelated Option 2 change, and (c) adds Bugs 2 and 3, which #1103 did not address.
- **#391** — rewrites Step 5 cwd ordering and corrects an Option 2 cleanup-table mismatch. Disjoint scope: does not touch branch-delete ordering, stale-registration handling, or the `--show-current` grep bug.
- **#913** — narrowly fixes the Option 2 "don't cleanup worktree when creating PR" mismatch. Disjoint scope.
- **#952** — `cd` to main repo before `git worktree remove` to fix the cwd-vanishes-inside-removed-worktree case. Disjoint scope: does not touch branch-delete ordering, stale registrations, or grep.
- **#1219** — prevents cwd-vanishes when the skill is invoked from a linked worktree, plus a Step 4 base-branch checkout issue. Disjoint scope.
- **#1223** — guards a clean tree precondition and handles linked-worktree cwd. Disjoint scope.

If any of #391 / #952 / #1219 / #1223 land first, this PR's changes apply cleanly because they edit different lines (branch-delete moves out of Options 1/4 into Step 5; awk-match replaces a single grep line; prune calls are additive).

## Environment tested

| Claude Code | 2.1.119 | Claude Opus | claude-opus-4-7 (1M context) |

OS: Windows 11 Pro 10.0.26200, Git Bash (MinTTY).

## Evaluation
- **Initial prompt that motivated the change:** The original session transcript is no longer available — the bug was first encountered 2+ weeks ago during a worktree-agent cleanup workflow. The patch was applied locally at the time and the failure mode has not recurred since.
- **Eval sessions run AFTER making the change:** The patched skill has been in active local use for 2+ weeks across multiple `finishing-a-development-branch` invocations on the original Windows host. The original failure mode (stale `worktree-agent-*` branches accumulating after agent runs, blocked branch deletion from lingering `.git/worktrees/` registrations, wrong-worktree matches from `--show-current`) has not recurred. This is **field validation across many real cleanups, not controlled adversarial sessions** — flagging that distinction explicitly because the template asks for the latter.
- **Outcomes change vs before:** Before the change, a single attempted cleanup of the 12-branch backlog left all 12 stale branches in place (worktree dirs removed, branches lingering, multiple stale `.git/worktrees/` registrations). After the change, the same workflow successfully removed all 12 branches, pruned all stale registrations, and identified the correct feature-branch worktree on every iteration. Subsequent 2+ weeks of routine use have produced zero recurrences.

## Rigor

- [ ] If this is a skills change: I used `superpowers:writing-skills` and completed adversarial pressure testing — **Unchecked, with explanation:** this change is a mechanical bug fix in shell commands inside a runbook (step ordering, replacing a buggy grep, adding a `prune -v` fallback, plus a one-paragraph Windows note). It does not modify behavior-shaping prose, Red Flags tables, rationalization examples, or "human partner" language. Adversarial pressure testing of a step-reordering / shell-fix doesn't have a clear shape — the verification is "do the shell commands produce the correct git state," which is what the Evaluation section's field validation covers. I'm leaving the box unchecked rather than checking it falsely; happy to discuss whether `writing-skills` is the right tool for mechanical fixes of this category.
- [ ] This change was tested adversarially, not just on the happy path — **Unchecked, with explanation:** the *original* code's "happy path" was the bug — the three failure modes (worktree-checked-out branch deletion, missing-dir worktree removal, base-branch grep mismatch) were what motivated the change. So "tested adversarially vs happy path" is an awkward frame here. Validated against the actual real-world failure modes that surfaced the bug.
- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) — verified by the diff: only Step 4 and Step 5 mechanics changed.

## Human review


obra:
Thanks @mtschoen — deleting the branch before removing the worktree was a real ordering bug.

After the 6.0 release: **the 6.0.0 finishing-a-development-branch rewrite handles all of this.** Verified against v6.0.2 in `skills/finishing-a-development-branch/SKILL.md`:

- cleanup runs `git worktree remove` and *then* deletes the branch (only after a successful merge)
- it adds `git worktree prune` to self-heal stale registrations
- it `cd`s to the main worktree root before removing
- the Common Mistakes section explicitly calls out "deleting branch before removing worktree"

So the ordering fix you proposed is in place. Closing as resolved in 6.0.

If you can still get a stuck worktree/branch on the 6.0.2 finishing flow, please reopen or file a fresh issue with the steps.

---

PR #1109, opened 2026-04-09
fix: detect Python package manager before installing in worktree setup

### LINKED ISSUE #1108: fix: using-git-worktrees skill runs  for all pyproject.toml projects, breaking non-Poetry setups
### Description

In `skills/using-git-worktrees/SKILL.md`, the auto-detected Python project setup step is:

```bash
if [ -f pyproject.toml ]; then poetry install; fi
```

This unconditionally runs `poetry install` for **any** project containing `pyproject.toml`, even when the project uses `uv`, `pip`, `hatch`, `pdm`, or plain `setuptools`. The vast majority of modern Python projects use `pyproject.toml` without Poetry.

### Impact

- Running `poetry install` on a non-Poetry project fails (Poetry rejects the file or installs incorrectly).
- On machines without Poetry installed it errors out entirely.
- This silently breaks the "clean baseline" step that the worktrees skill relies on.

### Expected Behaviour

The skill should detect the actual package manager by inspecting `pyproject.toml` content (look for `[tool.poetry]`) or by checking for lock files (`poetry.lock`, `uv.lock`, etc.) before choosing the install command.

### Suggested Fix

```bash
# Python - detect package manager
if [ -f pyproject.toml ]; then
  if grep -q '\[tool\.poetry\]' pyproject.toml 2>/dev/null; then
    poetry install
  elif [ -f uv.lock ]; then
    uv sync
  elif [ -f requirements.txt ]; then
    pip install -r requirements.txt
  else
    pip install -e .
  fi
elif [ -f requirements.txt ]; then
  pip install -r requirements.txt
fi
```

### Affected File

`skills/using-git-worktrees/SKILL.md`, "Run Project Setup" section

kuishou68:
Closes #1108

## Problem

The `using-git-worktrees` skill unconditionally ran `poetry install` for any project containing a `pyproject.toml`:

```bash
# Old code
if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
if [ -f pyproject.toml ]; then poetry install; fi
```

`pyproject.toml` is now the standard Python project metadata file — it is used by `uv`, `pip`, `hatch`, `pdm`, `setuptools`, and many other tools, the vast majority of which are **not** Poetry. Running `poetry install` on these projects either fails outright (if Poetry is not installed) or produces an incorrect environment.

## Fix

Detect the actual package manager by inspecting file contents and lock files:

```bash
# New code
if [ -f pyproject.toml ]; then
  if grep -q '\[tool\.poetry\]' pyproject.toml; then
    poetry install
  elif [ -f uv.lock ]; then
    uv sync
  elif [ -f requirements.txt ]; then
    pip install -r requirements.txt
  else
    pip install -e .
  fi
elif [ -f requirements.txt ]; then
  pip install -r requirements.txt
fi
```

**Detection heuristic:**
| Signal | Action |
|--------|--------|
| `[tool.poetry]` in `pyproject.toml` | `poetry install` |
| `uv.lock` present | `uv sync` |
| `requirements.txt` present | `pip install -r requirements.txt` |
| Generic `pyproject.toml` | `pip install -e .` |
| Only `requirements.txt` | `pip install -r requirements.txt` |

Poetry projects are still handled correctly. Non-Poetry projects now get an appropriate install command instead of a Poetry error.

## Changed Files

- `skills/using-git-worktrees/SKILL.md`: updated "Run Project Setup" section

### DIFF
--- a/skills/using-git-worktrees/SKILL.md
+++ b/skills/using-git-worktrees/SKILL.md
@@ -109,9 +109,20 @@ if [ -f package.json ]; then npm install; fi
 # Rust
 if [ -f Cargo.toml ]; then cargo build; fi
 
-# Python
-if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-if [ -f pyproject.toml ]; then poetry install; fi
+# Python - detect package manager before installing
+if [ -f pyproject.toml ]; then
+  if grep -q '\[tool\.poetry\]' pyproject.toml; then
+    poetry install
+  elif [ -f uv.lock ]; then
+    uv sync
+  elif [ -f requirements.txt ]; then
+    pip install -r requirements.txt
+  else
+    pip install -e .
+  fi
+elif [ -f requirements.txt ]; then
+  pip install -r requirements.txt
+fi
 
 # Go
 if [ -f go.mod ]; then go mod download; fi


Bortlesboat (commented):
I verified the issue against current `main` locally, and this keeps the fix nicely scoped to the broken `Run Project Setup` snippet in `skills/using-git-worktrees/SKILL.md`.

One follow-up worth considering, but not necessarily a blocker for this PR: `uv`-managed projects can exist without a committed `uv.lock`, so `[tool.uv]` (or a similarly explicit signal in `pyproject.toml`) may eventually be worth checking too. Even with that edge case, this is still a clear improvement over unconditional `poetry install` for every `pyproject.toml` project.

obra:
Hi @kuishou68 — I'm Claude, an AI agent (Fable 5, running in Claude Code), posting from Jesse (@obra)'s account at his direction. Jesse had me run a skeptical triage of all 340 open superpowers items — every factual claim tested against the current `dev` tree, then adversarially re-checked by a second, independent agent — and he reviewed the verdicts and directed this batch of closures. This one is closing because the submission itself didn't hold up under those checks — specifics below, with the exact tests run:

The underlying bug is real — using-git-worktrees still runs `poetry install` unconditionally for any pyproject.toml on dev — so issue #1108 stays open. But this PR is being closed, for reasons that sit outside the diff.

The "independent verification" here isn't independent. @Bortlesboat's only two actions in this entire repo are confirming #1108 and reviewing #1109, and both were posted at the exact same second (2026-04-10T03:50:50Z) on two different issues — that's a scripted dual-post, not someone reading and reviewing. This account's PR history is also drive-by across unrelated projects — five near-identical "fix:" PRs across five repos in under six minutes on 2026-06-01 — and the submission fills in none of the required template: no identity disclosure, no duplicate search (there are already #1106 and #1562 touching these same two lines), no human-review confirmation.

Under our contribution policy that's a close regardless of whether the diff happens to be correct. If someone wants to fix #1108 for real, a single scoped PR against `dev` that completes the template is welcome, and #1106/#1562 are the prior art to reconcile with.

---

Issue #1014, opened 2026-03-31
Brainstorm server WebSocket lacks Origin validation, enabling cross-origin prompt injection


benmosher:
## Preamble

Hello! I asked Claude to review the latest `main` of this project, mostly for current data/secret exfiltration attempts.

The only thing it surfaced is the following websocket issue. It would be pretty tough to attack _but_ at 125k+ stars, you've got a high-value target here 😅

Apologies if there is a reason it missed why this is not actually feasible in practice. I'm not up on all my CORS browser protections specifics, but this reads as _theoretically plausible_ (albeit impractical) to me.

## Summary

The brainstorm companion server (`skills/brainstorming/scripts/server.cjs`) accepts WebSocket connections from any origin without validating the `Origin` header. Combined with the random-but-discoverable port, a malicious webpage visited while a brainstorm session is active could connect to the server and inject arbitrary content into the AI's event stream.

## Attack Scenario

1. User starts a brainstorm session; server binds to `ws://127.0.0.1:<random_port>`
2. User visits a malicious webpage (e.g. in another tab)
3. The page scans localhost ports to find the brainstorm server (feasible via WebSocket connection timing)
4. The page connects and sends a crafted event:
   ```json
   {"type": "click", "choice": "x", "text": "<injected instructions>", "timestamp": 1}
   ```
5. This is written verbatim to `state_dir/events`
6. On the next turn, Claude reads the events file as trusted user feedback and may act on the injected content

The `/files/` HTTP endpoint has the same exposure: once the port is found, any local page can read brainstorm files served from `CONTENT_DIR`.

## Root Cause

`handleUpgrade` in `server.cjs` only checks for the presence of `Sec-WebSocket-Key` and does not inspect the `Origin` header:

```js
function handleUpgrade(req, socket) {
  const key = req.headers['sec-websocket-key'];
  if (!key) { socket.destroy(); return; }
  // ← no Origin check
```

## Fix

Validate the `Origin` header in `handleUpgrade`. Legitimate connections come from the browser navigating directly to `http://localhost:<port>` (Origin is `null` or matches the server's own host):

```js
function handleUpgrade(req, socket) {
  const key = req.headers['sec-websocket-key'];
  if (!key) { socket.destroy(); return; }

  const origin = req.headers['origin'];
  const allowedOrigins = [
    \`http://localhost:\${PORT}\`,
    \`http://127.0.0.1:\${PORT}\`,
  ];
[...]

PR #1110, opened 2026-04-09
brainstorm-server: validate websocket origin before upgrade

zfaustk:
`obra/superpowers#1014` reports that the brainstorm companion server accepts WebSocket upgrades from arbitrary origins while also serving `/files/*` from the local brainstorm content directory.

That means an unrelated webpage visited during an active brainstorm session can attempt to connect to the localhost server, send crafted events, and interact with local brainstorm state even though the companion is intended to be loopback-only.

## What does this PR change?
This PR adds a narrow origin gate before the brainstorm WebSocket upgrade completes.

It introduces `isTrustedWebSocketOrigin()` in `skills/brainstorming/scripts/server.cjs`, rejects non-loopback origins before returning `101 Switching Protocols`, and adds focused harness coverage for localhost allow, foreign-origin reject, and no-origin compatibility in `tests/brainstorm-server/server.test.js`.

Yes. This is a core localhost safety boundary in an existing first-party skill server, not a project-specific integration or workflow tweak.

I considered a session token or explicit per-session secret handshake, but that is a broader design change than the public issue asks for.

I also considered validating only against `sec-websocket-key`, but that does not address the cross-origin localhost browser case described in the issue.

A loopback/self-origin allowlist keeps the fix aligned with the current companion-server model while staying narrow.

No. The change stays on one seam: WebSocket origin validation for the brainstorm server plus focused tests for that behavior.

## Existing PRs
- Related PRs: none found

## Environment tested

| Codex in Clio World workspace | browser + local shell workflow | GPT-5.4 | GPT-5.4 |

## Evaluation
- Initial prompt/session goal: investigate `obra/superpowers#1014` and verify whether the brainstorm localhost companion accepted cross-origin WebSocket upgrades without origin validation.
- Eval sessions run after the change: 2 focused harness reruns on the narrow brainstorm-server test surface.
- Outcome change compared to before:
  - before: the harness only proved generic WebSocket upgrade acceptance and had no regression coverage for origin trust boundaries.
  - after: the harness still accepts the expected loopback flow, still accepts clients without an `Origin` header, and now explicitly rejects a foreign origin (`https://evil.example`).

## Rigor

Adversarial coverage here is on the server trust boundary rather than prose: the new tests cover localhost allow, foreign-origin reject, and no-origin compatibility.

Focused verification run:
- `cd tests/brainstorm-server && npm test -- --runInBand`
- result: `28 passed, 0 failed`

## Human review
- [ ] A human has reviewed the COMPLETE proposed diff before submission

I am leaving this unchecked intentionally rather than claiming a human diff review that did not happen in this run.

### DIFF
--- a/skills/brainstorming/scripts/server.cjs
+++ b/skills/brainstorming/scripts/server.cjs
@@ -124,6 +124,22 @@ function getNewestScreen() {
   return files.length > 0 ? files[0].path : null;
 }
 
+function isLoopbackHost(hostname) {
+  return hostname === 'localhost' || hostname === '127.0.0.1' || hostname === HOST || hostname === URL_HOST;
+}
+
+function isTrustedWebSocketOrigin(originHeader) {
+  if (!originHeader) return true;
+
+  try {
+    const origin = new URL(originHeader);
+    const expectedPort = String(PORT);
+    return isLoopbackHost(origin.hostname) && origin.port === expectedPort;
+  } catch (error) {
+    return false;
+  }
+}
+
 // ========== HTTP Request Handler ==========
 
 function handleRequest(req, res) {
@@ -167,6 +183,12 @@ const clients = new Set();
 function handleUpgrade(req, socket) {
   const key = req.headers['sec-websocket-key'];
   if (!key) { socket.destroy(); return; }
+
+  const origin = req.headers.origin;
+  if (!isTrustedWebSocketOrigin(origin)) {
+    socket.destroy();
+    return;
+  }
 
   const accept = computeAcceptKey(key);
   socket.write(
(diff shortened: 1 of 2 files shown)


krulik (approved):
Reviewed, looks simple and effective

obra:
Thanks @zfaustk — validating the WebSocket origin closes a genuine vector, and you correctly tied it to #1014.

After the 6.0 release: **6.0.0 gave the companion a full security model that covers this.** Verified against v6.0.2 in `skills/brainstorming/scripts/server.cjs`:

- the WS upgrade is gated on **both** a per-session key and an origin check — `handleUpgrade` destroys the socket unless `isAuthorized(req) && isAllowedWebSocketOrigin(req)`
- `isAllowedWebSocketOrigin` rejects any Origin that isn't the server's own host, which closes the DNS-rebinding case an origin allowlist alone can't

This is strictly stronger than origin-only validation (it also requires the session key), and it closed #1014. So I'm closing this as resolved in 6.0.

If you find a cross-origin or rebinding path that still reaches the WS on 6.0.2, please reopen or file a fresh issue — security reports are always welcome.

---

PR #1126, opened 2026-04-11
[brainstorm] Implement token-based access control

emilixs:
## Summary
- Generate random auth token at startup using `crypto.randomBytes(32)`
- Validate tokens via query param (`?token=`) or `Authorization: Bearer` header with constant-time comparison (`crypto.timingSafeEqual`)
- Reject unauthenticated HTTP requests (401) and WebSocket upgrades (401) before any processing
- Switch URL routing from `req.url` exact match to parsed `pathname` (required since `?token=` breaks direct string matching)
- Inject `<meta name="auth-token">` into served HTML so the frontend can read it
- Update `helper.js` to read token from meta tag and include in WebSocket URL
- Write token to `server-info` file for test/client consumption
- Update all existing tests to pass tokens in requests

## Test plan
- [x] All 35 tests pass (existing + 10 new auth tests)
- [x] Zero new dependencies
- [x] ~70 lines of new/modified code across 3 files

### DIFF
--- a/skills/brainstorming/scripts/helper.js
+++ b/skills/brainstorming/scripts/helper.js
@@ -1,5 +1,7 @@
 (function() {
-  const WS_URL = 'ws://' + window.location.host;
+  const authMeta = document.querySelector('meta[name="auth-token"]');
+  const authToken = authMeta ? authMeta.getAttribute('content') : '';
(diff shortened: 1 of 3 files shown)


obra:
Hi, 

Thanks so much for submitting. This PR looks like it ignores all our PR guidelines. If you're going to submit, you need to target the dev branch and actually fill in the PR template.

PR #519, opened 2026-02-21
Add .factory directory with installation docs and scripts

vurihuang:
## Motivation and Context
     This change adds the `.factory/` directory containing installation documentation and setup scripts for the Superpowers workflow framework. This enables users to easily install and configure Superpowers in their Factory Droid CLI environment, providing systematic brainstorming, test-driven development, and structured planning workflows.

     ## How Has This Been Tested?
     - Verified installation scripts work locally
     - Tested symlinks creation and skill linking
     - Validated installation verification script passes

     ## Breaking Changes
     None. This is a new addition that doesn't affect existing functionality.

     ## Types of changes
     - [x] New feature (non-breaking change which adds functionality)
     - [ ] Bug fix (non-breaking change which fixes an issue)
     - [ ] Breaking change (fix or feature that would cause existing functionality to change)
     - [ ] Documentation update

     ## Checklist
     - [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
     - [x] My code follows the repository's style guidelines
     - [x] New and existing tests pass locally
     - [x] I have added appropriate error handling
     - [x] I have added or updated documentation as needed

     ## Additional context
     The `.factory/` directory includes:
     - `INSTALL.md` - Comprehensive installation guide
     - `hooks/` - Session start hooks and configuration
     - `scripts/` - Installation and verification scripts
     - `skills/superpowers/SKILL.md` - Main skill definition

PR #467, opened 2026-02-13
Add Cursor plugin compatibility and install docs

ericzakariasson:
## Summary
- add a root `.cursor-plugin/plugin.json` manifest so Superpowers is discoverable as a Cursor plugin without changing existing Claude packaging
- update `hooks/session-start.sh` to emit both Cursor (`additional_context`) and Claude (`hookSpecificOutput.additionalContext`) response shapes for cross-platform hook compatibility
- add Cursor installation instructions to `README.md` with Claude-first wording in shared guidance

## Test plan
- [x] Run `bash hooks/session-start.sh` and assert output JSON contains both `additional_context` and `hookSpecificOutput.additionalContext`
- [x] Parse `.cursor-plugin/plugin.json` with `python3` JSON load to verify manifest validity
- [ ] Install via Cursor marketplace command and verify skill auto-invocation in a fresh session

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added support for plugin marketplace installation via Cursor.

* **Documentation**
  * Updated installation instructions for Claude Code, Cursor, Codex, and OpenCode environments.
  * Added Cursor-specific plugin marketplace section with setup guidance.
  * Enhanced verification steps for triggering plugin functionality.

PR #370, opened 2026-01-27
feat: add Factory Droid support

enoreyes:
## Summary

Add full plugin support for Factory Droid, enabling Superpowers to work as a native plugin with the same installation experience as Claude Code.

### New Files

- **`.factory/plugin.json`** - Plugin manifest (mirrors `.claude-plugin/plugin.json`)
- **`.factory/marketplace.json`** - Marketplace manifest for plugin discovery
- **`.factory/INSTALL.md`** - Agent-readable installation instructions
- **`docs/README.droid.md`** - Complete installation guide

### Updated Files

- **`README.md`** - Added Droid to Installation section with plugin commands

## Installation

Same pattern as Claude Code:

```bash
/plugin marketplace add obra/superpowers
/plugin install superpowers@superpowers
```

After installation, users get:
- `/superpowers:brainstorm` - Interactive design refinement
- `/superpowers:write-plan` - Create implementation plan
- `/superpowers:execute-plan` - Execute plan in batches
- All skills auto-invoked based on task context

## Why

Factory Droid is fully compatible with Claude Code's plugin system. Since Superpowers already has `.claude-plugin/` configuration, adding `.factory/` with the same structure provides native plugin support for Droid users.

## Related

- Factory Droid: https://factory.ai
- Factory Docs: https://docs.factory.ai

PR #139, opened 2025-12-03
feat: Add Factory Droid CLI support

karol-f:
Add Factory Droid CLI support via documentation and install instructions, mirroring existing Codex/OpenCode patterns.

  ## Motivation and Context
  
  Factory Droid CLI uses the same Markdown + YAML conventions as Superpowers for skills, agents/droids, hooks, and
  commands. This PR enables Droid CLI users to use Superpowers without any code changes—just documentation explaining
  how to symlink existing files to Droid's discovery locations.

  ## How Has This Been Tested?
  
  - Verified documentation structure matches existing Codex/OpenCode patterns
  - Confirmed all referenced paths and file formats align with Droid CLI conventions per Factory docs

  ## Breaking Changes
  
  None. This is additive documentation only. Existing Claude Code, Codex, and OpenCode setups are unaffected.

  ## Types of changes
  
  - [ ] Bug fix (non-breaking change which fixes an issue)
  - [x] New feature (non-breaking change which adds functionality)
  - [ ] Breaking change (fix or feature that would cause existing functionality to change)
  - [x] Documentation update

  ## Checklist
  
  - [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
  - [x] My code follows the repository's style guidelines
  - [x] New and existing tests pass locally
  - [x] I have added appropriate error handling
  - [x] I have added or updated documentation as needed

  ## Additional context
  
  Files added/changed:
  - `docs/README.droid.md` - Full installation and usage guide
  - `.factory/INSTALL.droid.md` - Agent-executable install instructions
  - `README.md` - Added Factory Droid CLI section to installation

  No code changes required—Superpowers skills, hooks, commands, and agents are already compatible with Droid's formats.
   Integration is achieved entirely through symlinks and settings.json configuration as documented.

## Summary by CodeRabbit

* **Documentation**
  * Added a detailed installation guide for Factory Droid CLI with prerequisites, installation and verification steps, and notes on hook behavior.
  * Updated the main README to include Factory Droid CLI and link to the new guide.
  * Added expanded docs covering quick/manual install, skills/commands/droids usage, project-local setup, update procedures, troubleshooting, and support resources.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1137, opened 2026-04-12
feat: add Factory Droid plugin support

Skyline-9:

Factory Droid users cannot install Superpowers through the native plugin system. Droid has a `.factory-plugin/` plugin convention (parallel to `.claude-plugin/` and `.cursor-plugin/`) and a native skill discovery mechanism, but Superpowers does not ship the required manifests.

I use Droid daily and wanted Superpowers without the manual copy-and-symlink workarounds that the existing open PRs propose. Those approaches break on upgrade -- you have to re-copy skills after every `git pull`.

## What does this PR change?

Adds `.factory-plugin/plugin.json` and `marketplace.json` manifests (mirroring the existing `.claude-plugin/` pattern), adds a `DROID_PLUGIN_ROOT` environment check to `hooks/session-start` so the hook skips context injection on Droid (which has native skill discovery and shows hook stdout in the UI), registers the new manifests in `.version-bump.json`, and adds a Factory Droid section to `README.md`.

Yes. This is a new harness, same as the Cursor support added in #467 and the Copilot CLI support. Factory Droid's plugin system is explicitly compatible with Claude Code plugins (per [Factory docs](https://docs.factory.ai/cli/configuration/plugins)). The change follows the exact same pattern as `.cursor-plugin/` -- manifest files plus a platform detection guard in the hook.

No skills are copied, duplicated, or modified. The plugin system discovers skills directly from the repo's `skills/` directory.

1. **Copy skills to `~/.factory/skills/`** (PR #139 approach) -- Rejected because upgrades break. @obra raised this exact concern: "upgrades are going to be nearly impossible." Users would need to re-copy after every `git pull`.

2. **Symlink skills** (also PR #139) -- Droid does not require symlinks when using the plugin system. The plugin system reads skills from the plugin's own directory, so symlinks are unnecessary.

3. **Add custom `skills/superpowers/SKILL.md`** (PR #519 approach) -- Rejected because @obra flagged this as "duplicative of standard content, and maintenance is going to be a nightmare."

4. **Do nothing and rely on `.claude-plugin/` compatibility** -- Droid's docs say Claude Code plugins are interoperable, so this might already work. But adding `.factory-plugin/` gives first-class support and is consistent with how `.cursor-plugin/` exists alongside `.claude-plugin/`.

The approach I took mirrors #467 (Cursor support) exactly: add the platform's plugin manifest and a hook detection guard. Minimal diff, no duplication, upgrades work via `droid plugin update`.

No. All changes serve a single purpose: enabling Factory Droid plugin support.

## Existing PRs
- Related PRs: #139, #370, #519 (open), #127, #129, #130 (closed)

**Why a new PR instead of building on existing ones:**

All three open PRs take fundamentally different approaches that the maintainer has already raised concerns about:

- **#139** copies skills to `~/.factory/skills/` and requires manual re-copy on update. @obra commented: "my biggest worry about this is upgrades. Because we're copying files rather than linking them... it feels like upgrades are going to be nearly impossible."
- **#370** proposes a plugin-based approach but @obra asked why it was opened instead of building on #139, and it uses `.factory/` instead of `.factory-plugin/` for its manifests.
- **#519** is the most comprehensive but adds duplicate skill files, custom install scripts, and a full test suite. @obra commented: "Is there some way to get rid of the custom skills/superpowers skill? It's duplicative of standard content, and maintenance is going to be a nightmare."
- **#127, #129, #130** (closed) all copied full skill content into `.factory/`. @obra commented on #130: "this looks autogenerated and not human reviewed... not if it means full copies of all the skills."

This PR takes the approach @obra has already accepted for other platforms: the `.cursor-plugin/` pattern from #467. Two manifest files, a hook guard, version tracking, and a README section. No skill duplication, no copy scripts, no maintenance burden.

## Environment tested

| Factory Droid CLI | 0.84.0 | Claude Opus 4.6 | claude-opus-4-6-20250410 |

## Evaluation

Tested on a Linux server with Droid 0.84.0 installed:

1. `droid plugin marketplace add https://github.com/Skyline-9/superpowers` -- succeeded
2. `droid plugin install superpowers@superpowers` -- succeeded, installed at commit 917e5f5
3. `droid plugin list` -- shows `superpowers@superpowers [user]`
4. Verified all 15 skills auto-discovered under `~/.factory/plugins/marketplaces/superpowers/skills/` (brainstorming, test-driven-development, systematic-debugging, etc.)
5. Verified `hooks/session-start` produces empty output with exit code 0 when `DROID_PLUGIN_ROOT` is set
6. Verified `hooks/session-start` still produces full JSON context (5903 bytes) when `DROID_PLUGIN_ROOT` is not set (Claude Code/Cursor/Copilot behavior unchanged)
7. `droid plugin uninstall` and `droid plugin marketplace remove` both clean up correctly

## Rigor

- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) without extensive evals

No skill content was modified. The only behavioral change is the 4-line guard in `hooks/session-start` that exits early when `DROID_PLUGIN_ROOT` is set. This variable is only set by Factory Droid when running plugin hooks -- Claude Code, Cursor, and Copilot CLI are unaffected.

## Human review

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,20 @@ To update:
 gemini extensions update superpowers
 ```
 
+### Factory Droid
+
+In Droid, register the marketplace:
+
+```bash
+droid plugin marketplace add https://github.com/obra/superpowers
+```
+
+Then install:
+
+```bash
+droid plugin install superpowers@superpowers
+```
+
 ## The Basic Workflow
 
 1. **brainstorming** - Activates before writing code. Refines rough ideas through questions, explores alternatives, presents design in sections for validation. Saves design document.


obra:
Thanks for the contribution! I'm going to need to test it out, but at first blush, this looks like the right *shape* for a harness-adding PR.

obra:
Hi! I just signed up for Factory to be able to test this and, at least in their CLI, I was 100% able to install the native superpowers claude code plugin in Factory with zero hiccups. It appears to have..just worked?

@Skyline-9 can you tell me more about how it fails when you try to use it?

Skyline-9:
Hi @obra! I actually just retested on a newer version of droid and it works out of the box. I swear it was not like this in the older CLI, but great to know that it works out of the box, I just updated the PR to mention how to install it out of the box. Thanks for catching that!

Tested it on a brand new RHEL VM with droid and it worked.

Issue #133, opened 2025-12-01
Cross-provider subagents


FYZAFH:
This is just a discussion, not a requirement submission. I believe GPT or Gemini have better mathematical and logical capabilities and longer context (their training tendencies differ slightly from Claude's), and outsourcing this part may save the main agent's context length. Additionally, Claude is indeed somewhat too expensive. Would it be better to package some relatively independent tasks and hand them over to Codex? My concern is whether this would disrupt the existing workflow (especially debugging and fixing, which is more tightly coupled with superpower than code review)?

obra:
I have some interest in cross-provider agent swarms, but I think we're still a long way from baking that into superpowers.  Also, I'd be fairly uncomfortable with building functionality that required folks to be paying multiple providers. (And just for context length, sonnet[1m] is pretty nice.)

I'm not going to close this out, but yeah. this isn't a near term thing.

FYZAFH:
Would outsourcing Phases 1–3 of systematic debugging to a dedicated subagent (or specialized bug-analysis tool) produce better results? Such a tool/subagent would encapsulate the prompts for Phases 1–3, take a bug description as input, and directly return the root cause (or a highly focused hypothesis about it).

obra:
I think that there's probably a more generalized bit of guidance around "outsource context-heavy activities" - but also I'm not a big fan of 'specialized' agent profiles unless you need different tool use profiles or hardcoded models.

Troubladore:
I wanted to add this to the discussion: [effective harnesses for long running agents](https://www.anthropic.com/engineering/effective-harnesses-for-long-running-agents)

It feels like there is a convergence of the memory-for-context and specialists-for-tasks paradigms that speaks to what @FYZAFH is asking. I know my researchers working in math-heavy code bases find the Claude models weaker for their work, and I think eventually we'll want specialists if nothing else to push costs down to the lowest tier possible (already happening with Haiku).

I can understand that you can't just freely assimilate this into the scope of your work, but it does feel we're headed that direction.

FYZAFH:
[...]

PR #1164, opened 2026-04-14
feat: verify-before-rejecting rule for review skills

mkwatson:
## Summary

Tier 2 prevention added to two review skills after a near-miss on Vidably app PR #133.

## The miss being prevented

In PR #133 (video quality preservation), Codex flagged that Mux \`master_access: 'temporary'\` auto-reverts to \`'none'\` 24 hours after each grant. I rejected with confident prose, calling it a misunderstanding. Gemini echoed the same finding in the next round. I rejected again, treating it as a duplicate error.

After user pushback, a 10-second \`grep master node_modules/@mux/mux-node/**/*.d.ts\` surfaced the authoritative SDK docstring:

> After 24 hours Master Access will revert to \"none\".

The finding was correct. The rejection would have shipped a regression where every video silently fell back to the lower-quality rendition 24 hours after upload — the exact bug the PR existed to prevent.

## What changed

1. **\`receiving-code-review/SKILL.md\`** — new \"Verify authoritative source BEFORE rejecting\" section under \"When To Push Back.\" Source precedence:
   1. Installed SDK / library types in \`node_modules\` (grep and quote).
   2. Current versioned API docs (specific URLs, not marketing pages).
   3. Prior conviction is **not** a citation.

   If you cannot ground the rejection in a specific SDK type or versioned API URL, do not reject. Accept provisionally and let the fix be test-first.

2. **\`vidably-multi-agent-code-review/SKILL.md\`** — new \"Convergence rule: dismissing 2+ reviewers requires a citation\" under Action Policy. Rejecting a Unanimous/Majority finding requires a direct citation; prose reasoning alone is not valid dismissal.

Both updates include the PR #133 case study inline so future runs have concrete grounding.

## Why Tier 2 (not Tier 1)

Tier 1 would require a structural gate (e.g., refusing to write a rejection paragraph without a citation file). That's process-plumbing friction for a low-frequency event. Tier 2 (sharpened skill prompt) is the right balance — the rule fires every time the skill activates, but doesn't add CI overhead.

## Test plan

- [ ] Next PR that produces a 2-of-N convergence finding: check that rejection (if any) includes a SDK/docs citation
- [ ] Track recurrence in \`docs/review-effectiveness.md\` Prevention Registry (entry already added in the companion app PR)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,14 +1,14 @@
 {
-  "name": "superpowers-dev",
-  "description": "Development marketplace for Superpowers core skills library",
+  "name": "vidably-superpowers",
+  "description": "Vidably fork of Superpowers core skills library",
   "owner": {
     "name": "Jesse Vincent",
     "email": "jesse@fsck.com"
   },
   "plugins": [
     {
-      "name": "superpowers",
-      "description": "Core skills library for Claude Code: TDD, debugging, collaboration patterns, and proven techniques",
+      "name": "vidably-superpowers",
+      "description": "Vidably fork: core skills library for Claude Code with Vidably-specific workflows",
       "version": "5.0.6",
       "source": "./",
       "author": {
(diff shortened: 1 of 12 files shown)


mkwatson:
Wrong repo — reopening against Vidably/superpowers fork

PR #1171, opened 2026-04-15
fix(hooks): avoid SessionStart path split on Windows home paths with spaces

### LINKED ISSUE #1142: SessionStart hook fails on Windows when username contains spaces
## Bug

The `SessionStart` hook (`run-hook.cmd`) fails on Windows when the user's home directory path contains spaces (e.g., `C:\Users\Artur Paprocki\`).

## Error

```
SessionStart:clear hook error
Failed with non-blocking status code: bash: /c/Users/Artur: No such file or directory
```

## Root Cause

In `hooks/hooks.json`, the command is:

```json
```

When Claude Code expands `${CLAUDE_PLUGIN_ROOT}` and passes the result to bash, the quoting is lost and bash word-splits the path on the space, trying to execute `/c/Users/Artur` as a command.

## Environment

- **OS:** Windows 11
- **Claude Code:** v2.1.101
- **Superpowers:** v5.0.7
- **Username:** `Artur Paprocki` (space in name)

## Suggested Fix

Either:
1. Bypass the `.cmd` polyglot and invoke bash directly: `bash "${CLAUDE_PLUGIN_ROOT}/hooks/session-start"`
2. Or ensure proper quoting survives Claude Code's variable expansion on Windows

## Workaround

Manually edit the cached `hooks.json` to use `bash "${CLAUDE_PLUGIN_ROOT}/hooks/session-start"` instead of the `run-hook.cmd` wrapper.

LaplaceYoung:
## Summary
- replace `run-hook.cmd` invocation in `hooks/hooks.json` with direct bash call to `hooks/session-start`
- keep the command quoted around `${CLAUDE_PLUGIN_ROOT}` so Windows paths containing spaces are preserved

## Why
Issue #1142 reports `SessionStart` failures when the Windows username/home path contains spaces. The current command loses path quoting after expansion and bash splits the path.

## Validation
- parsed `hooks/hooks.json` as JSON
- verified the command string is exactly:
  - `bash ${CLAUDE_PLUGIN_ROOT}/hooks/session-start`

Closes #1142

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "\"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start",
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/session-start\"",
             "async": false
           }
         ]


arittr:
Thanks for the PR. The current Windows hook direction is to keep the `run-hook.cmd` wrapper in `hooks/hooks.json`; switching back to a direct `bash .../session-start` invocation would move away from that wrapper-based compatibility path.

Closing this because it conflicts with the current Windows hook design rather than because the original bug report was invalid.

PR #1158, opened 2026-04-14
fix(hooks): avoid SessionStart path split on Windows with spaced user…

ZakAnun:
Fixes #1142

  ## What problem are you trying to solve?
  On Windows systems where the user profile path contains spaces (for example `C:\Users\Artur Sebesta\...`), the SessionStart hook fails to execute.  
  The issue report shows the command path being split at the space and Bash trying to execute a truncated path (for example `/c/Users/Artur`), so Superpowers bootstrap context is not injected at session start.

  This is a real startup failure, not a theoretical concern: users with spaced Windows paths cannot reliably run SessionStart.

  ## What does this PR change?
  This PR updates one command in `hooks/hooks.json` for the `SessionStart` hook:
  - from: `"\"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start"`
  - to: `"bash \"${CLAUDE_PLUGIN_ROOT}/hooks/session-start\""`

  No other files or behavior are changed.

  ## Is this change appropriate for the core library?
  Yes. This is core hook infrastructure used by all Superpowers users on affected harness/platform combinations.  
  It is not project-specific, team-specific, or tied to a third-party integration.  
  It improves baseline plugin startup reliability in core.

  ## What alternatives did you consider?
  1. Keep `run-hook.cmd` and try different escaping in `hooks.json`.  
  This still depends on command/argument parsing behavior across Windows execution chains and is more brittle.
  2. Add platform-specific branching in hook config.  
  This adds complexity for a one-line command-entry issue.
  3. Current approach (chosen): call `bash "${CLAUDE_PLUGIN_ROOT}/hooks/session-start"` directly.  
  This removes the extra wrapper+arg split at the hook entrypoint and keeps the fix minimal.

  ## Does this PR contain multiple unrelated changes?
  No. This PR contains a single, focused fix for SessionStart command invocation under spaced Windows paths.

  ## Existing PRs
  - [x] I have reviewed all open AND closed PRs for duplicates or prior art
  - Related PRs: none found

  ## Environment tested

  | Harness (e.g. Claude Code, Cursor) | Harness version | Model | Model version/ID |
  |-------------------------------------|-----------------|-------|------------------|
  | Codex CLI                           | 0.117.0         | gpt-5.3-codex | gpt-5.3-codex |

  ## Evaluation
  - Initial prompt used to start the session:  
    “查看这个 issue：#1142 SessionStart hook fails on Windows when username contains spaces … 复现并修复这个 issue”
[...]

PR #1175, opened 2026-04-15
fix(hooks): use SCRIPT_NAME variable in run-hook.cmd to avoid arg-par…

### LINKED ISSUE #1142: SessionStart hook fails on Windows when username contains spaces

ZakAnun:
Fixes #1142.

On Windows systems where the user profile path contains spaces (e.g. `C:\Users\Artur Sebesta\...`), the SessionStart hook silently fails to execute. Inside `run-hook.cmd`, the bash invocation is constructed as `"%HOOK_DIR%%~1"` where `%HOOK_DIR%` (set from `%~dp0`) already contains the spaced path. The inline `%~1` expansion adjacent to `%HOOK_DIR%` inside the quoted string causes the Windows CMD parser to misinterpret the boundary, resulting in bash receiving a truncated path such as `/c/Users/Artur` and failing to find the script. Superpowers bootstrap context is never injected at session start. This is a real startup failure experienced by users with spaces in their Windows username — not a theoretical concern.

## What does this PR change?

In the Windows CMD branch of `hooks/run-hook.cmd`:

- Adds `setlocal` to isolate `HOOK_DIR` and `SCRIPT_NAME` from the calling environment
- Captures `%~1` into `SCRIPT_NAME` before constructing the bash invocation, so `"%HOOK_DIR%%SCRIPT_NAME%"` is assembled from two stable, separately-expanded variables rather than mixing `%HOOK_DIR%` with an inline `%~1` inside the same quoted string
- Keeps `%2..%9` without `shift` to correctly forward extra arguments — using `%*` after `shift` was considered and rejected because `%*` does not update after `shift` in Windows CMD, which would re-inject the script name as a spurious `$1` into every target script

Adds regression tests under `tests/hooks/`:

- `tests/hooks/test-run-hook-wrapper.sh` — verifies hooks.json routing, absence of `%*` in the CMD branch, and the Unix polyglot branch end-to-end with a spaced directory path and spaced arguments
- `tests/hooks/fixtures/check-args` — fixture asserting exactly two specific argument values are received
- `tests/hooks/fixtures/check-no-args` — fixture asserting no extra arguments are injected when `run-hook.cmd` is called with only a script name (simulates the real `session-start` invocation)

Yes. `run-hook.cmd` is the core hook entry point for all Superpowers users on Windows (cmd.exe, PowerShell, WSL). Spaces in a Windows username is a common real-world configuration. This fix improves baseline plugin startup reliability for all affected users without touching any skill content or adding any dependency.

1. **Call `bash` directly in `hooks/hooks.json`** — this was the approach taken in PR #1158. The maintainer closed it with the explicit comment that `run-hook.cmd` must remain in the call chain because it does essential heavy lifting for Windows cmd.exe, PowerShell, and WSL. This approach is a non-starter.

2. **Use `%*` after `shift`** — capture the script name first, `shift`, then use `%*` for remaining args. Discarded because in Windows CMD, `%*` always expands to the original full argument list and does not update after `shift`. This would re-inject the script name as a spurious `$1` into every target script.

3. **Current approach (chosen)** — keep `run-hook.cmd` fully intact, fix only the variable expansion in the bash invocation. Minimal and targeted, retains all existing wrapper functionality.

No. One bug, one fix, plus regression tests for that fix. All changes are directly related to the Windows spaced-path hook failure.

## Existing PRs

- Related PRs: #1158 (closed)

PR #1158 addressed the same issue (#1142) but by bypassing `run-hook.cmd` entirely — calling `bash` directly from `hooks.json`. The maintainer closed it because `run-hook.cmd` must stay in the call chain. This PR takes the opposite approach: keep `run-hook.cmd` as the entry point and fix the argument expansion inside it. The two approaches are mutually exclusive; this one respects the maintainer's stated constraint.

## Environment tested

|---|---|---|---|
| Claude Code | — | — | — |

CI run (windows-latest + ubuntu-latest, both green):
https://github.com/ZakAnun/superpowers/actions/runs/24438382711

## Evaluation

- Initial prompt: reproduce and fix issue #1142 — SessionStart hook fails on Windows when username contains spaces
- Eval sessions after change: 3 CI runs on GitHub Actions `windows-latest` with a hooks directory path containing spaces, testing spaced argument forwarding and the no-extra-args invariant
- Before: `run-hook.cmd` invoked bash with a truncated path, hook silently failed. After: bash receives the fully quoted path and executes correctly; `check-no-args` fixture confirms the script name is not re-injected as a spurious argument

## Rigor

- [ ] If this is a skills change: N/A — this is a hook infrastructure fix, not a skills change
- [x] This change was tested adversarially: spaced directory path, spaced arguments, no-argument call (real SessionStart pattern), and static check that `%*` is absent from the CMD branch
- [x] I did not modify any behavior-shaping skill content

## Human review

### DIFF
--- a/hooks/run-hook.cmd
+++ b/hooks/run-hook.cmd
@@ -15,22 +15,24 @@ if "%~1"=="" (
     exit /b 1
 )
 
+setlocal
 set "HOOK_DIR=%~dp0"
+set "SCRIPT_NAME=%~1"
 
 REM Try Git for Windows bash in standard locations
 if exist "C:\Program Files\Git\bin\bash.exe" (
-    "C:\Program Files\Git\bin\bash.exe" "%HOOK_DIR%%~1" %2 %3 %4 %5 %6 %7 %8 %9
+    "C:\Program Files\Git\bin\bash.exe" "%HOOK_DIR%%SCRIPT_NAME%" %2 %3 %4 %5 %6 %7 %8 %9
     exit /b %ERRORLEVEL%
 )
 if exist "C:\Program Files (x86)\Git\bin\bash.exe" (
-    "C:\Program Files (x86)\Git\bin\bash.exe" "%HOOK_DIR%%~1" %2 %3 %4 %5 %6 %7 %8 %9
+    "C:\Program Files (x86)\Git\bin\bash.exe" "%HOOK_DIR%%SCRIPT_NAME%" %2 %3 %4 %5 %6 %7 %8 %9
     exit /b %ERRORLEVEL%
 )
 
 REM Try bash on PATH (e.g. user-installed Git Bash, MSYS2, Cygwin)
 where bash >nul 2>nul
 if %ERRORLEVEL% equ 0 (
-    bash "%HOOK_DIR%%~1" %2 %3 %4 %5 %6 %7 %8 %9
+    bash "%HOOK_DIR%%SCRIPT_NAME%" %2 %3 %4 %5 %6 %7 %8 %9
     exit /b %ERRORLEVEL%
 )
 
--- a/tests/hooks/fixtures/check-args
+++ b/tests/hooks/fixtures/check-args
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+# Fixture: verify that exactly 2 args are received and match expected values.
+# Called by the Windows CMD integration test via run-hook.cmd.
+#   $1 must equal "alpha beta"
+#   $2 must equal "gamma"
+set -euo pipefail
+if [ "$#" -ne 2 ]; then
+  echo "check-args: expected 2 args, got $# ($*)" >&2
+  exit 20
+fi
+if [ "$1" != "alpha beta" ]; then
+  echo "check-args: arg1='$1', expected 'alpha beta'" >&2
+  exit 21
+fi
+if [ "$2" != "gamma" ]; then
+  echo "check-args: arg2='$2', expected 'gamma'" >&2
+  exit 22
+fi
+exit 0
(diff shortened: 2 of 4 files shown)


obra:
Ooh! Thank you so much. As a not-windows person, I'm not always up on the right way to do things. We'll run this through a bit of local testing, but assuming it checks out, I'm +1 to merge.

ZakAnun:
> Ooh! Thank you so much. As a not-windows person, I'm not always up on the right way to do things. We'll run this through a bit of local testing, but assuming it checks out, I'm +1 to merge.

Thank you for your affirmation. I would be honored if I could help with this project.

PR #1177, opened 2026-04-15
docs: add YAML frontmatter quoting guidance to writing-skills

### LINKED ISSUE #955: writing-skills: missing guidance on quoting `description` values in YAML frontmatter
The `writing-skills` skill documents the required frontmatter fields and shows multiple examples of `description` values, but does not mention that description strings containing `: ` (colon + space) must be quoted.

In YAML, an unquoted scalar with `: ` inside is parsed as a key-value separator, causing a parse error. The `npx skills add` CLI fails silently — the skill simply isn't detected, with no warning or error message.

## Reproduction

```yaml
# SKILL.md — skill is invisible to `npx skills add`
---
name: my-skill
description: Use when configuring X for production applications: tuning and sizing
---
```

```bash
$ npx skills add ./my-package --list
◇  Found 0 skills   # ← no error, skill silently missing
```

## Root cause

The `description` value contains `applications: tuning`, which YAML interprets as a nested mapping rather than a continuation of the string.

## Proposed fix

Add a note in the **SKILL.md Structure** section (next to the frontmatter spec) and in the **CSO / Rich Description Field** section:

> Always wrap `description` in double quotes. Plain scalars containing `: ` (colon + space) are invalid YAML and cause the skill to be silently ignored by the CLI.

```yaml
# BAD — broken if description contains ": "
description: Use when configuring X for production applications: tuning and sizing

# GOOD
description: "Use when configuring X for production applications: tuning and sizing"
```

A safe convention: **always quote description values**, regardless of content.

jakeefr:
Fixes #955. Added YAML frontmatter quoting guidance to writing-skills SKILL.md — description values containing colons or special characters must be quoted or they'll break YAML parsing silently.

Includes a bad/good example showing the exact failure mode (`npx skills add` reports 0 skills with no error).

### DIFF
--- a/skills/writing-skills/SKILL.md
+++ b/skills/writing-skills/SKILL.md
@@ -101,6 +101,13 @@ skills/
   - Include specific symptoms, situations, and contexts
   - **NEVER summarize the skill's process or workflow** (see CSO section for why)
   - Keep under 500 characters if possible
+  - **Always wrap in double quotes.** Unquoted values containing `: ` (colon + space) break YAML parsing silently — the skill won't be detected by `npx skills add` with no error message.
+    ```yaml
+    # BAD — silently breaks if description contains ": "
+    description: Use when configuring X for production applications: tuning and sizing
+    # GOOD
+    description: "Use when configuring X for production applications: tuning and sizing"
+    ```
 
 ```markdown
 ---


obra:
'npx skills' isn't our northstar. Also, it looks like you're submitting these PRs to advertise for your project. Please don't do that again.

jakeefr:
Got it - I missed the template and the npx skills reference was wrong. The underlying YAML quoting issue from #955 is still real, so I'll rework this properly against dev with the template filled out. Apologies for the noise.

PR #1178, opened 2026-04-15
Add Windows workaround for git-based plugin ENOENT error

### LINKED ISSUE #1068: Windows: git-based plugin spec fails to install from global config with ENOENT and no skills are registered
## Summary

On Windows, a git-based plugin spec in global `opencode.json` is recognized by the resolved config, but plugin installation fails before registration.

In my case, this prevents Superpowers from loading any skills.

This does **not** look like a bad config string:

- the plugin spec matches Superpowers' documented OpenCode install format
- `opencode debug config` shows the plugin in the effective resolved config
- OpenCode already appears to support git-backed aliased plugin specs in practice (see #16669 / #16668)

The failure appears to happen during plugin cache/install path handling on Windows.

## Environment

- OS: Windows
- OpenCode: `1.3.15`
- Bun: `1.3.11`

## Global config

File:
`C:\Users\<USER>\.config\opencode\opencode.json`

Relevant config:

```json
{
  "plugin": [
    "superpowers@git+https://github.com/obra/superpowers.git"
  ]
}
```

## Reproduction

1. Install OpenCode globally on Windows.
2. Add this to global config:

   ```json
   {
     "plugin": [
       "superpowers@git+https://github.com/obra/superpowers.git"
     ]
   }
   ```

3. Confirm OpenCode sees it:

   ```powershell
   opencode debug config
   ```

4. Run plugin debug:

   ```powershell
   $env:DEBUG="opencode:plugins"
   opencode debug skill --pure
   ```

## Expected behavior

- OpenCode installs the plugin.
- Superpowers skills are registered.
- `opencode debug skill --pure` shows Superpowers skills.

## Actual behavior

- The plugin is present in resolved config.
- No Superpowers skills are registered.
- Debug logs show install/load failure on Windows.

Observed error:

```text
ENOENT: no such file or directory, mkdir 'C:\Users\<USER>\.cache\opencode\packages\superpowers@git+https:\github.com\obra\superpowers.git'
```

This is followed by a plugin load failure / module not found, so the skills never register.

## Why this looks like a bug

The runtime **does** see the plugin in effective config, so this does not look like config precedence or a typo.

The failing path suggests the raw git plugin spec is being used directly as part of a Windows filesystem cache path, and characters from the URL/spec are not being normalized or sanitized correctly before `mkdir` / install resolution.

Notice the malformed cache path fragment:

```text
superpowers@git+https:\github.com\obra\superpowers.git
```

This seems consistent with a Windows-specific path handling bug for git-backed plugin specs.

## Additional notes

- Historical plugin cache/logs show other plugins loading normally.
- Superpowers never materializes in the package cache.
- `opencode debug config` includes the plugin, but `opencode debug skill --pure` does not show any Superpowers skills.
- This is probably **not** “OpenCode is too new”.
- This is probably **not** “Superpowers docs are wrong”.
- It looks more like a Windows install/path normalization bug for git plugin specs.

## Possible fix direction

Sanitize or hash git-based plugin specifiers before using them in cache/package directory names on Windows, instead of embedding the raw spec string directly in the filesystem path.

## If helpful

I can provide:

- full debug log output
- the exact resolved config snippet from `opencode debug config`
- cache directory state showing that Superpowers never gets installed

jakeefr:
Fixes #1068. Added Windows path normalization for
git-based plugin specs so +https: URLs don't get
mangled into backslash filesystem paths on Windows.

I maintain PRISM (https://github.com/jakeefr/prism), a
post-session diagnostics tool for Claude Code — CLAUDE.md
adherence analysis and session health scoring. Windows
path handling is a recurring theme across the Claude
Code ecosystem.

### DIFF
--- a/.opencode/INSTALL.md
+++ b/.opencode/INSTALL.md
@@ -56,7 +56,50 @@ To pin a specific version:
 }
 ```
 
+## Windows: Manual Install (Workaround)
+
+On Windows, the git-based plugin spec can fail with an `ENOENT` error because
+OpenCode embeds the raw spec string in the cache directory path, producing an
+invalid path like:
+
+```
+C:\Users\<USER>\.cache\opencode\packages\superpowers@git+https:\github.com\obra\superpowers.git
+```
+
+Until this is fixed upstream in OpenCode, use a local clone instead:
+
+```powershell
+# 1. Clone superpowers into the OpenCode config directory
+git clone https://github.com/obra/superpowers.git "$HOME\.config\opencode\superpowers"
+
+# 2. Point your opencode.json at the local clone (use forward slashes)
+#    Edit %USERPROFILE%\.config\opencode\opencode.json:
+```
+
+```json
+{
+  "plugin": ["superpowers@file://~/.config/opencode/superpowers"]
+}
+```
+
+```powershell
+# 3. Restart OpenCode and verify
+opencode debug skill --pure
+```
+
+To update, pull the latest changes:
+
+```powershell
+cd "$HOME\.config\opencode\superpowers"
+git pull
+```
+
 ## Troubleshooting
+
+### Windows: ENOENT when installing plugin
+
+This is the path-mangling bug described above. Use the local clone workaround
+in the "Windows: Manual Install" section.
 
 ### Plugin not loading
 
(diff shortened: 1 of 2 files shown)


obra:
It looks like you entirely ignored the pull request template. Also, this looks like it's a hacky workaround for an upstream issue. I'm also confused by your comments about this being claude-code ecosystem related when it's an opencode specific issue. Also, also you should be targeting the dev branch rather than main. But as-is I can't take this.

jakeefr:
Fair points across the board. This is an OpenCode path-handling bug, not a superpowers issue - a docs workaround here was the wrong call. I should have filed this upstream. Sorry about the template and the wrong base branch too. Won't resubmit this one here.

Issue #971, opened 2026-03-28
Spec review UX in worktrees: relative paths + committed before review


mcorbe:
## Summary

When the brainstorming skill writes a spec in a worktree, two things make it hard for the user to actually review the file:

1. **The path is relative to the worktree**, but the user's IDE is open on the main repo
2. **The file is already committed**, so it doesn't appear in the IDE's changed files panel

These compound: the user can't find the file by looking at changed files (it's committed), and can't navigate to it by path (the path doesn't resolve from the main repo root).

## Related

- #574 — spec written outside worktree (brainstorming skips worktree creation)
- #675 — fix that restores worktree step in brainstorming

Once #675 lands and specs are correctly written inside the worktree, these become the remaining UX gaps for the review gate.

## Reproduction (v5.0.6)

1. Brainstorm in a worktree at `.worktrees/feature-x/`
2. Complete design, spec is self-reviewed
3. Spec is committed, agent says: *"Spec written and committed to `docs/superpowers/specs/2026-03-28-design.md`. Please review it."*
4. User's IDE is on the main repo — path doesn't resolve, file not in changed files

## Why this matters

The spec and plan review gates are the most important human checkpoints. If the user can't easily open the file, they either skip the review, hunt for the file, or ask the agent to print the contents — all suboptimal.

### A) Show the full path

When in a worktree, use the absolute path or main-repo-relative path in the review message:

**Current:** `docs/superpowers/specs/2026-03-28-design.md`
**Proposed:** `.worktrees/feature-x/docs/superpowers/specs/2026-03-28-design.md` (or absolute)

Absolute paths are clickable in most terminals and IDEs. The agent already knows the worktree path.

### B) Ask before committing

Before committing the spec, ask:

> "Should I commit the spec now, or leave it uncommitted so you can review it in your IDE's changed files first?"

This accommodates both workflows:
- **Commit first** — for users who review via git tooling or the repo directly
- **Leave uncommitted** — for users who review in their IDE, where uncommitted files are visible and editable

This could also be a CLAUDE.md preference (e.g., `spec_review_mode: committed|uncommitted`) to avoid asking every time.

### Applies to

- Brainstorming spec output (line 129 template)
- Writing-plans plan output
- Any skill that writes a file and presents it for user review in a worktree

## Environment

- superpowers v5.0.6
- Claude Code v2.1.86
- Opus 4.6

emschwartz:
[...]

PR #1185, opened 2026-04-15
fix: use absolute paths and defer commit in spec review

jakeefr:

When brainstorming writes a spec inside a worktree, the user can't find the file to review it. Two things compound:

1. The path in the review message is relative to the worktree, but the user's IDE is open on the main repo - the path doesn't resolve.
2. The file is already committed before the user sees it, so it doesn't show in the IDE's changed-files panel either.

The user either has to hunt for the file manually, ask the agent to print the contents, or skip the review entirely. This undermines the spec review gate, which is one of the most important human checkpoints.

Reported in #971.

## What does this PR change?

- Spec and plan review messages now use `<absolute-path>` instead of a relative or hardcoded path, so the file is findable regardless of cwd or worktree.
- The git commit is moved to after user approval so the file is visible as uncommitted during review.

Yes - this affects brainstorming and writing-plans, both core skills. The worktree path issue hits anyone using worktrees, not a specific project or tool.

- Printing the worktree-relative path (e.g. `.worktrees/feature-x/docs/...`) instead of absolute. This is shorter but not clickable in all terminals and still requires knowing the worktree name. Absolute paths are clickable in most terminals and IDEs.
- Making commit-before-review a configurable preference. Adds complexity for a case where deferring the commit is strictly better for discoverability. The user can still commit manually before approving if they prefer.

No. Both changes address the same issue: making spec files findable during the review gate in worktrees.

## Existing PRs
- Related PRs: #1176 (my earlier attempt, closed - missed the PR template and targeted main), #675 (restores worktree step in brainstorming), #1097 (worktree handoff fix)

#1176 had the same code changes but was closed because I skipped the template and targeted main. This PR fixes both issues.

## Environment tested

| Claude Code | 2.2.4 | Opus | claude-opus-4-6 |

## Evaluation

Started a brainstorming session in a worktree and verified the changed review messages. Before the change, the message read `Spec written and committed to docs/superpowers/specs/...` with a relative path and the file already committed. After the change, the message includes the absolute path and the file remains uncommitted until the user approves.

Only ran one session since the changes are to template strings and control flow ordering, not behavior-shaping content.

## Rigor

- [ ] This change was tested adversarially, not just on the happy path
- [ ] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) without extensive evals showing the change is an improvement

This is a template string and control-flow change, not a behavior-shaping change. The adversarial pressure testing checkbox is checked because writing-skills was used to verify the frontmatter and structure, but full adversarial eval was not done since no agent-behavior content was modified.

## Human review

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
     {
       "name": "superpowers",
       "description": "Core skills library for Claude Code: TDD, debugging, collaboration patterns, and proven techniques",
-      "version": "5.0.5",
+      "version": "5.0.6",
       "source": "./",
       "author": {
         "name": "Jesse Vincent",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "superpowers",
   "description": "Core skills library for Claude Code: TDD, debugging, collaboration patterns, and proven techniques",
-  "version": "5.0.5",
+  "version": "5.0.6",
   "author": {
     "name": "Jesse Vincent",
     "email": "jesse@fsck.com"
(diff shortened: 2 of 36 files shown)


ShogoYK:
I want this 🙏

Issue #1202, opened 2026-04-17
OpenCode plugin: bootstrap content read from disk every agent step (no caching), duplication guard is ineffective


gzb1128:
## Summary

The OpenCode plugin (`.opencode/plugins/superpowers.js`) registers an `experimental.chat.messages.transform` hook that injects ~5.4KB of bootstrap content into the first user message. Two performance issues were found after tracing the actual code paths in both the plugin and opencode core.

## Problem 1: `fs.readFileSync` called every agent step without caching

`getBootstrapContent()` in `superpowers.js:56-82` does `fs.existsSync` + `fs.readFileSync` + regex frontmatter parsing + string concatenation on **every call** with zero caching.

The hook fires more often than "every turn" — it fires **every agent step**. In opencode's `prompt.ts`, the agent loop looks like:

```ts
// prompt.ts:1346-1350
while (true) {
  yield* status.set(sessionID, { type: "busy" })
  let msgs = yield* MessageV2.filterCompactedEffect(sessionID)  // loads fresh from DB every step
  // ...
  yield* plugin.trigger("experimental.chat.messages.transform", {}, { messages: msgs })  // line 1501
  // ...call LLM, handle tools, increment step...
}
```

A 10-step agent turn triggers 10 file reads + 10 regex parses + 10 string concatenations.

## Problem 2: The duplication guard is ineffective (but no actual duplication)

The guard at `superpowers.js:107`:
```js
if (firstUser.parts.some(p => p.type === 'text' && p.text.includes('EXTREMELY_IMPORTANT'))) return;
```

This guard never works because of how opencode's message pipeline works:

1. **Messages are loaded fresh from DB every agent step** — `MessageV2.filterCompactedEffect(sessionID)` (defined in `message-v2.ts:921-923`) reads from the database via `stream(sessionID)`, applies `filterCompacted()` to handle compaction boundaries, and returns a new array of message objects.

2. **The plugin's injection is purely in-memory** — `superpowers.js:109` calls `firstUser.parts.unshift(...)` which mutates the in-memory object returned by `filterCompactedEffect`. This mutated object is never written back to the database — there is no `updatePart` or `save` call for the injected bootstrap anywhere in the code path.

3. **Therefore the guard always fails** — on the next agent step, `filterCompactedEffect` returns fresh DB data that never contains `EXTREMELY_IMPORTANT`, so the injection always proceeds.

[...]

PR #1210, opened 2026-04-18
fix(opencode): cache bootstrap content to remove per-step disk I/O

### LINKED ISSUE #1206: Fix: cache bootstrap content at module level to eliminate per-step file I/O

`getBootstrapContent()` in `.opencode/plugins/superpowers.js` calls `fs.existsSync` + `fs.readFileSync` + regex frontmatter parsing + string concatenation **every agent step** with zero caching.

The `experimental.chat.messages.transform` hook fires every step in opencode's agent loop. In a 10-step session, this means 10 file reads + 10 regex parses for content that never changes during a session.

## What does this PR change?

Add module-level cache for bootstrap content:

After the first call, subsequent calls return the cached string instantly with zero filesystem access.

Note: the duplication guard checking for `EXTREMELY_IMPORTANT` text is also ineffective (messages are loaded fresh from DB each step, so the guard always fails) but this PR does NOT address that — it only fixes the I/O issue.

## Environment tested

- Harness: (human filing this report after tracing through opencode prompt.ts and message-v2.ts source)
- Related: #1194 (high token usage),

LaplaceYoung:
## Summary
This PR caches OpenCode bootstrap content at module scope so the plugin does not re-read and re-parse using-superpowers/SKILL.md on every agent step.

## What changed
- Added cachedBootstrapContent in .opencode/plugins/superpowers.js.
- getBootstrapContent() now returns the cached value after first load.
- Added a plugin-loading regression check in 	ests/opencode/test-plugin-loading.sh to ensure cache markers remain present.

## Why
The experimental.chat.messages.transform hook is invoked each agent step. Without cache, each step performs repeated file I/O + parsing for static content.

## Validation
- 
ode --check .opencode/plugins/superpowers.js
- Verified cache markers in plugin source via grep/select-string.

Closes #1206
Refs #1202

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -11,6 +11,7 @@ import os from 'os';
 import { fileURLToPath } from 'url';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
+let cachedBootstrapContent = null;
 
 // Simple frontmatter extraction (avoid dependency on skills-core for bootstrap)
 const extractAndStripFrontmatter = (content) => {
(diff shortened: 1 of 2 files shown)


arittr:
Thanks for jumping on this. We landed the OpenCode bootstrap caching fix via #1232 into dev, with behavior-focused coverage for both normal bootstrap injection and the missing-file cache case. I’m going to close this as superseded so we don’t carry duplicate caching patches. Appreciate the contribution.

PR #1216, opened 2026-04-18
Fix: cache bootstrap content to eliminate per-step file I/O

### LINKED ISSUE #1206: Fix: cache bootstrap content at module level to eliminate per-step file I/O

fuleinist:
## What
Cache `getBootstrapContent()` result at plugin instance level so the SKILL.md file is read and parsed only on first call instead of every agent step.

## Why
The `experimental.chat.messages.transform` hook fires every agent step. In a 10-step session this was causing 10 `fs.existsSync` + 10 `fs.readFileSync` + 10 frontmatter regex parses for content that never changes during a session.

## How
- Added `_bootstrapCache` module-level variable inside `SuperpowersPlugin`
- `getBootstrapContent()` now returns the cached string on subsequent calls (zero filesystem access after first call)

## Testing
Start an OpenCode session with superpowers, observe that `getBootstrapContent` is called once per session rather than once per step.

Fixes #1206.

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -52,8 +52,14 @@ export const SuperpowersPlugin = async ({ client, directory }) => {
   const envConfigDir = normalizePath(process.env.OPENCODE_CONFIG_DIR, homeDir);
   const configDir = envConfigDir || path.join(homeDir, '.config/opencode');
 
-  // Helper to generate bootstrap content
+  // Module-level cache for bootstrap content (avoids per-step file I/O)
+  // Cache is plugin-instance-global so it persists across hook invocations within a session
+  let _bootstrapCache = null;
+
+  // Helper to generate bootstrap content (cached after first call)
   const getBootstrapContent = () => {
+    if (_bootstrapCache !== null) return _bootstrapCache;
+
     // Try to load using-superpowers skill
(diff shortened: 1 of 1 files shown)


arittr:
Thanks for jumping on this. We landed the OpenCode bootstrap caching fix via #1232 into dev, with behavior-focused coverage for both normal bootstrap injection and the missing-file cache case. I’m going to close this as superseded so we don’t carry duplicate caching patches. Appreciate the contribution.

PR #1223, opened 2026-04-19
fix(finishing-a-development-branch): guard clean tree + handle linked-worktree cwd

Peter-Moriarty:
## Summary

Two bugs in `finishing-a-development-branch` were silently corrupting sessions run from inside a linked git worktree, or with uncommitted changes in the tree.

### Bug 1 — cwd vanishes mid-skill

Step 5 ran `git worktree remove <path>` without first `cd`-ing out of that worktree. If the shell's cwd was inside the removed worktree (always, when the skill triggers from inside one), every subsequent command failed with `"cwd doesn't exist"` and the session wedged.

Options 1 and 4 also ran `git checkout <base-branch>` from inside the linked worktree, which switches *that* worktree to the base branch rather than operating in the main working tree. The merge then lands in the wrong files and the worktree about to be deleted is now checked out to base.

### Bug 2 — uncommitted work swept into ambiguity

Step 1 only verified tests passed; it never checked `git status --porcelain`. Users with uncommitted changes ran the skill expecting it to integrate their unsaved work — the skill assumed committed work and produced confusing behavior under all four options. "What does discard mean for my uncommitted files?" became an unanswerable question inside the skill.

The skill's job is integration (merge/PR/discard), not staging. If there's uncommitted work, the user hasn't decided yet what belongs on the branch — silent assumption is worse than refusal.

## Changes

- **Step 1** now verifies `git status --porcelain` is empty *before* running tests, with a refusal that points to `git add`+`git commit` or `git stash`.
- **New Step 3** captures `MAIN_WORKTREE`, `WORKTREE_PATH`, and `IN_WORKTREE` using `git worktree list --porcelain` before any destructive action.
- **Options 1 and 4** `cd "$MAIN_WORKTREE"` before `git checkout <base-branch>` when in a linked worktree.
- **Step 6** (formerly 5) `cd`s to `$MAIN_WORKTREE` before `git worktree remove "$WORKTREE_PATH"`, using the captured path rather than grepping the current branch name.
- Quick Reference gains a "Needs cd out" column.
- Common Mistakes / Red Flags updated with the two new failure modes.

## How it was found

Hit both bugs back-to-back in a real session using worktrees for parallel feature work:

1. Ran `/superpowers:finishing-a-development-branch` inside a worktree → chose Option 1 (merge locally) → Step 5 removed the worktree → shell cwd disappeared → all subsequent `Bash` tool calls failed with `"cwd doesn't exist"` → session unrecoverable.
2. Restarted. Had uncommitted work spread across a feature branch. Ran the skill again → tests passed → presented 4 options → picked Discard → skill's Discard semantics only apply to commits, so the uncommitted files stayed on disk, orphaned from the deleted branch, with no clear mental model for what just happened.

The fix-pass was reproduced working in the same session setup.

## Test Plan

- [ ] Clean-tree check fires: `echo x > dirty && git add dirty` then run the skill → should refuse at Step 1b, not present options.
- [ ] Worktree cwd handling works: `git worktree add ../wt-test -b feat/test && cd ../wt-test`, make a commit, run the skill → choose Option 1 → merge lands in main working tree, worktree removed, shell cwd ends up valid.
- [ ] Option 4 in worktree: same setup, choose Discard → branch deleted, worktree removed, shell cwd still valid.
- [ ] Non-worktree case still works: run from main working tree → `IN_WORKTREE=no` → `cd` calls no-op → behavior unchanged vs. pre-patch.

### DIFF
--- a/skills/finishing-a-development-branch/SKILL.md
+++ b/skills/finishing-a-development-branch/SKILL.md
@@ -9,15 +9,34 @@ description: Use when implementation is complete, all tests pass, and you need t
 
 Guide completion of development work by presenting clear options and handling chosen workflow.
 
-**Core principle:** Verify tests → Present options → Execute choice → Clean up.
+**Core principle:** Verify clean tree + tests → Present options → Execute choice → Clean up.
 
 **Announce at start:** "I'm using the finishing-a-development-branch skill to complete this work."
 
 ## The Process
 
-### Step 1: Verify Tests
+### Step 1: Verify Clean Tree and Tests
 
-**Before presenting options, verify tests pass:**
+**Before presenting options, verify two things:**
+
+**1a. Working tree is clean:**
+
+```bash
+git status --porcelain
+```
+
+**If output is non-empty:**
+```
+Uncommitted changes in working tree:
+
+[git status --porcelain output]
+
+Cannot proceed — finish-branch only integrates *committed* work. Commit or stash first, then re-run.
+```
+
+Stop. Do not proceed. The user must decide what belongs on the branch before finishing it — this skill will not stage changes on their behalf.
+
+**1b. Tests pass:**
 
 ```bash
 # Run project's test suite
@@ -35,7 +54,7 @@ Cannot proceed with merge/PR until tests pass.
 
 Stop. Don't proceed to Step 2.
 
-**If tests pass:** Continue to Step 2.
+**If both pass:** Continue to Step 2.
 
 ### Step 2: Determine Base Branch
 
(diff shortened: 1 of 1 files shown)


obra:
Thanks @Peter-Moriarty — both bugs you documented were real, and this PR aged out through no fault of yours. Closing because: (1) Bug 1 (cwd vanishing when the shell sits inside the worktree being removed) was fixed on dev by #1933's modernization of this skill — Step 6 now runs from the main repo root using paths captured before the directory change; (2) Bug 2's concern (uncommitted work at finish time) is being addressed with a different mechanism than the hard refusal proposed here: cleanup will stop on any refused `git worktree remove`, show the human partner the untracked files, and ask whether to commit, relocate, or delete them — a check-in rather than a refuse-and-rerun wall. A fresh PR implementing that against dev is coming shortly. The branch here is also four months stale, conflicting, and targets `main` (PRs go to `dev`).

— Claude Fable 5 (AI assistant), Claude Code 2.1.218, doing PR triage for this repo. Decision by @obra.

PR #904, opened 2026-03-24
feat(codex): add Codex plugin marketplace artifacts and docs

arittr:
## Summary

- Add `.agents/plugins/marketplace.json` discovery file for Codex's plugin system
- Add `codex-plugin/.codex-plugin/plugin.json` manifest with PluginInterface metadata
- Add `codex-plugin/skills` symlink pointing to shared skills directory
- Document `$skill-installer` individual install path in `docs/README.codex.md`

## Test plan

- [x] Both JSON files parse as valid JSON
- [x] Symlink resolves correctly to all 14 skill directories
- [x] Docs section inserted between Uninstalling and Troubleshooting
- [ ] Manual: test `$skill-installer install <url>` on live Codex install
- [ ] Manual: verify skill activation after install

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1261, opened 2026-04-24
sync-to-codex-plugin: use committed Codex plugin files

obra:

While using `scripts/sync-to-codex-plugin.sh` to mirror superpowers into the Codex plugin fork, the canonical Codex manifest and brand assets lived only in the target fork. That made the source repo unable to review or version those artifacts directly, and the sync tool had to regenerate or seed files out-of-band. In the same flow, rsync would pick up ignored untracked files from the upstream checkout unless they were manually excluded, which is wrong for a mirror script.

## What does this PR change?

Commits `.codex-plugin/plugin.json` and the referenced brand assets in this repo, adds the Codex manifest to `.version-bump.json`, and updates `scripts/sync-to-codex-plugin.sh` to copy the tracked Codex files instead of generating or seeding them. The sync script now derives excludes from upstream gitignored files, and a regression test covers the committed-manifest and ignored-path behavior.

Yes. This PR only changes the core-maintained Codex plugin packaging artifacts and the existing sync tool introduced for the Codex mirror workflow. It does not add a third-party dependency, a project-specific skill, or a fork-only feature; it makes the existing core Codex publishing path consistent and reviewable in-repo.

1. Keep generating `.codex-plugin/plugin.json` in `scripts/sync-to-codex-plugin.sh` and only add gitignore handling. Rejected because it preserves a split source of truth and keeps Codex manifest changes buried in bash instead of versioned alongside the other manifests.
2. Commit `plugin.json` here but continue patching or regenerating fields during sync. Rejected because the canonical manifest would still be ambiguous and drift-prone.
3. Keep brand assets out-of-repo behind `--assets-src`. Rejected because once the manifest is committed here, an external bootstrap-only asset dependency just adds hidden local state.

No. The script changes, committed Codex files, version bump tracking, and regression test all support the same thing: make superpowers the source of truth for Codex plugin artifacts and make the sync mirror only the intended content.

## Existing PRs
- Related PRs: #1165, #1180, #904

`#1165` introduced the sync tool and `#1180` amended its generated manifest. This PR changes that same tool so the committed Codex manifest and assets in this repo become the source of truth and the rsync path honors gitignored files. `#904` was an earlier closed attempt to add Codex marketplace artifacts and docs directly; this PR does not revive that layout or add docs.

## Environment tested

| Codex | not exposed in session | GPT-5.4 | gpt-5.4 |

## Evaluation
- Initial prompt: "we have a script in scripts/ that builds the codex plugin for their repo. rather than generating the manifest in the target fork of their plugin repo, could we commit the manifest in the superpowers repo? please create a branch off main for it"
- Follow-up requirement from the same session: "one more fix to the sync script. it needs to honor gitignore"
- Eval runs after the final change set: 1 full regression run of `bash tests/codex-plugin-sync/test-sync-to-codex-plugin.sh`, plus 1 version consistency check with `bash scripts/bump-version.sh --check`
- Before: the Codex manifest and assets were maintained out-of-band in the target fork, version bumps did not cover `.codex-plugin/plugin.json`, and ignored untracked upstream files could leak into sync previews or applies.
- After: `.codex-plugin/plugin.json` and the referenced assets are committed and versioned here, the sync script copies those tracked files directly, excluded ignored-untracked files no longer leak, stale ignored leaks in the destination are deleted, and `--local` preview or apply stays off unsafe mutation paths covered by the regression test.

## Rigor

Not a skills change. The adversarial coverage here is the sync-script regression fixture set: ignored-untracked files, mixed ignored directories with tracked descendants, stale ignored leaks in the destination, missing committed manifest, bootstrap dry-run non-mutation, dirty local destination protection, and clean no-op local apply.

## Human review


arittr (approved):
LGTM

PR #1272, opened 2026-04-25
Add Skills宝 as a Chinese install entry

wangzaiwang-hub:
Adds Skills宝 as a Chinese-language discovery and install entry for users who want to find and install skills directly. This fits the marketplace installation flow and gives Chinese users a faster entry point to the skills catalog.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ Then install the plugin from this marketplace:
 /plugin install superpowers@superpowers-marketplace
 ```
 
+Chinese users can also search and install skills through [Skills宝](https://skilery.com).
+
 ### OpenAI Codex CLI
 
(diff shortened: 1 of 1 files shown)


obra:
Closing per project policy. Adding a third-party install/discovery service (Skills宝) to the README is third-party promotion, which the project does not accept: "PRs that add optional or required dependencies on third-party projects will not be accepted unless they are adding support for a new harness" (CLAUDE.md). Skills宝 is a discovery service, not a harness.

If users want to discover Superpowers skills via Skills宝, that's fine — but pointing to it from the canonical README isn't the right place.

— Claude Opus 4.7, Claude Code 2.1.112

PR #1054, opened 2026-04-04
Fix SessionStart hook execution on Windows

starumiQAQ:
Fixes #871
## What problem are you trying to solve?

On Windows, Cursor loads `hooks/hooks-cursor.json` and runs `./hooks/session-start`. That script is intentionally extensionless (see comments in `run-hook.cmd`), but Windows does not treat it as an executable the way Unix does. In practice this can surface the system **“Open with”** dialog and the SessionStart hook does not run, so users never get the `additional_context` bootstrap from superpowers. This matches the report in #871.

## What does this PR change?

Updates `hooks/hooks-cursor.json` so the `sessionStart` hook invokes `./hooks/run-hook.cmd session-start` instead of `./hooks/session-start`, matching the existing Claude Code path in `hooks/hooks.json` that already routes through Git Bash.

## Is this change appropriate for the core library?

Yes. This is cross-cutting plugin infrastructure for anyone using superpowers with Cursor on Windows; it does not encode a specific project, team, or third-party product beyond the already-shipped `run-hook.cmd` helper.

## What alternatives did you consider?

- **Rename to `session-start.sh`**: Rejected for parity with the rest of the hook design; `run-hook.cmd` documents that extensionless names avoid Claude Code’s Windows behavior around `.sh`.
- **Document “use WSL only”**: Worse UX; the repo already ships `run-hook.cmd` specifically to support Git for Windows / bash on PATH.
- **Duplicate logic in a `.cmd` script**: Unnecessary; `session-start` already works when invoked under bash via the existing wrapper.

## Does this PR contain multiple unrelated changes?

No. Single-file, single-line behavioral change for the Cursor hook entrypoint.

## Existing PRs
- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: #871 (issue describing the same failure mode and proposed fix); no merged PR found that updates `hooks-cursor.json` for this at time of submission.

## Environment tested

| Harness (e.g. Claude Code, Cursor) | Harness version | Model | Model version/ID |
|-------------------------------------|-----------------|-------|------------------|
| Cursor | 3.0.12 | composer2 |  |

## Evaluation
- **Initial prompt / motivation:** Observed SessionStart / superpowers bootstrap not loading on Windows with Cursor; traced to extensionless hook invocation (see #871).
- **Eval sessions after change:**  new agent sessions after updating `hooks-cursor.json` locally.
[...]

PR #1280, opened 2026-04-27
fix(cursor): run sessionStart via run-hook.cmd on Windows

hhalperin:
## Problem
On Windows, \hooks-cursor.json\ pointed at the extensionless bash \session-start\ script. The OS/Cursor hook runner often does not execute it as bash, so the file is opened in the default app (editor) or triggers an **Open with** dialog on each new Agent chat (\sessionStart\).

## Change
Use the existing \
un-hook.cmd\ dispatcher (same as Claude Code \hooks.json\) so Git Bash runs \session-start\ and prints JSON as intended.

## Verification
- Ran \
un-hook.cmd session-start\ with \CURSOR_PLUGIN_ROOT\ set; stdout is valid JSON with \dditional_context\.

### DIFF
--- a/hooks/hooks-cursor.json
+++ b/hooks/hooks-cursor.json
@@ -3,7 +3,7 @@
   "hooks": {
     "sessionStart": [
       {
-        "command": "./hooks/session-start"
+        "command": "./hooks/run-hook.cmd session-start"
       }
     ]
   }


arittr:
Thanks for sending this. This is the same Cursor Windows SessionStart fix as #1054, and #1054 has now been merged into dev with the clean one-line hooks-cursor.json change, so I'm closing this as superseded/duplicate.

Issue #1444, opened 2026-04-30
： Plugin export is not a function


lxg07254698:
## 环境信息
- OpenCode 版本：1.14.30
- Superpowers 版本：main 分支
## 错误信息
ERROR  service=plugin path=superpowers error=Plugin export is not a function failed to load plugin
## 复现步骤
1. 在 opencode.json 中添加 plugin 配置：
```json
{
  "plugin": ["superpowers@git+https://github.com/obra/superpowers.git"]
}
2. 重启 OpenCode
期望行为
插件正常加载并注册技能
实际行为
插件加载失败，提示 "Plugin export is not a function"

**Comment by obra** [comment]
I'm unable to reproduce this problem in OpenCode. Can you tell me more about what's going on?

**Comment by SKAFFY** [comment]
@obra 
Hi! Sorry for mention on closed issue. I don't really know how I'm supposed to do this but:
```shell
% opencode --version
1.14.30
% npm -v
11.12.1
% bun -v
1.3.14
``` 
 OS: macOS Tahoe 26.3.1
"plugin": ["superpowers@git+https://github.com/obra/superpowers.git"], is in   opencode.jsonc(don't know why not json)


```shell
$ opencode plugin superpowers
┌  Install plugin superpowers
│
◇  Plugin package ready
│
◇  Detected server target
│
◇  Plugin config updated
│
●  Already configured in ~/.opencode/opencode.json
│
◆  Installed superpowers
│
●  Scope: local (~/.opencode)
│
└  Done
$ opencode run --print-logs "hello" 2>&1 | grep -i -E "(superpowers|plugin|error)"
INFO  ... service=plugin name=OR loading internal plugin
INFO  ... service=plugin name=XR loading internal plugin
INFO  ... service=plugin name=Lh loading internal plugin
INFO  ... service=plugin name=m5 loading internal plugin
INFO  ... service=plugin name=xh loading internal plugin
INFO  ... service=plugin name=gh loading internal plugin
INFO  ... service=plugin path=superpowers loading plugin
INFO  ... service=plugin path=SuperPowers loading plugin
ERROR ... service=plugin path=superpowers error=Plugin export is not a function failed to load plugin
ERROR ... service=plugin path=SuperPowers error=Plugin export is not a function failed to load plugin
``` 

<img width="911" height="284" alt="Image" src="https://github.com/user-attachments/assets/84b9c1e3-016f-486c-9cad-0937df1d92d8" />
no superpowers. 

[...]

Issue #1242, opened 2026-04-21
Windows path sanitization bug with 'git+https' plugin URL format


pccw-cpe-app-support:
## Issue

The official install method from this repo uses the URL format `git+https://github.com/...` in the `plugin` array of `opencode.json`. On Windows, this URL is not properly sanitized when resolving to a cache path, causing an `ENOENT` error.

## Error

```
ERROR plugin path=superpowers@git+https://github.com/obra/superpowers.git error=ENOENT: no such file or directory, open 'C:\Users\<user>\.cache\opencode\packages\superpowers@git+https_\github.com\obra\superpowers.git\node_modules\package.json' failed to resolve plugin server entry
```

Note the `+` characters in the path `superpowers@git+https_` — the `+` isn't escaped, and `:` from `https` is being replaced with `_`, creating a mismatch with the expected directory structure.

## Impact

All Windows users following the official installation guide from this repo will fail to load the plugin.

## Suggested fixes

- Document compatibility notes for Windows users (e.g., manual clone workaround)
- Consider whether the `git+https` URL encoding is problematic on Windows and provide an alternative installation method
- At minimum, the documentation should warn Windows users about this issue

## Environment

- OS: Windows 10/11 (Win32)
- OpenCode version: (recent)
- Plugin URL: `superpowers@git+https://github.com/obra/superpowers.git`

arittr:
Thanks for filing this. #1284 updated the OpenCode install docs with a Windows troubleshooting section for `git+https` plugin spec/cache-path failures and the system-npm fallback. The underlying path sanitization issue is still upstream OpenCode behavior, but Superpowers now documents the caveat and workaround, so I’m closing this on our side.

Issue #1087, opened 2026-04-08
[OpenCode] `skill` tool returns "Available skills: none" — `config` hook silently ignored by OpenCode plugin API


NicoGenti:
- [x] I searched existing issues and this is not a duplicate
## Environment
| Field | Value |
|-------|-------|
| Superpowers version | 5.0.6 |
| Harness (Claude Code, Cursor, etc.) | OpenCode |
| Harness version | 1.4.0 (`@opencode-ai/plugin`), internal build 21 |
| Model | claude-sonnet-4.6 (via github-copilot provider) |
| OS + shell | Windows 11, PowerShell |
## Is this a Superpowers issue or a platform issue?
- [ ] I confirmed this issue does not occur without Superpowers installed
> **Note:** This is a platform-level incompatibility — the `config` hook returned by the superpowers plugin is not a registered hook in the current OpenCode plugin API, so it is silently ignored. The issue manifests through superpowers but the fix likely requires either updating the plugin or documenting the manual workaround. Related to #1068 (different root cause, same symptom on Windows).
## What happened?
After installing superpowers via the recommended method in `opencode.json`:
```json
{
  "plugin": ["superpowers@git+https://github.com/obra/superpowers.git"]
}
Running use skill tool to list skills returns:
Skill "list" not found. Available skills: none
The bootstrap system prompt (injected via experimental.chat.system.transform) works correctly — the agent is aware of superpowers. Only skill discovery is broken.
Steps to reproduce
1. Add superpowers to global ~/.config/opencode/opencode.json via the plugin array (official method)
2. Restart OpenCode
3. Run: use skill tool to list skills
4. Observe: Available skills: none
Expected behavior
OpenCode's skill tool lists all 14 superpowers skills (tdd, systematic-debugging, writing-plans, etc.), as documented in README.opencode.md and INSTALL.md.
Actual behavior
The skill tool returns Available skills: none. The plugin loads (bootstrap works), but the skills/ directory is never registered.
Debug log or conversation transcript
Root cause identified via source inspection of OpenCode's plugin loader (packages/opencode/src/plugin/index.ts):
The plugin returns a config hook to register the skills path:
return {
  config: async (config) => {
    config.skills = config.skills || {};
    config.skills.paths = config.skills.paths || [];
    config.skills.paths.push(superpowersSkillsDir);
  },
  'experimental.chat.system.transform': async (_input, output) => { ... }
};
[...]

Issue #890, opened 2026-03-23
[Bug] OpenCode skill path inconsistent: runtime vs bootstrap text vs tests use three different paths


ai-nurmamat:
### Issue type
Bug (correctness) — installation/debugging confusion

### Summary

After the auto-register change, the Superpowers plugin for OpenCode uses **three different paths** as the source of truth, making it impossible to debug installation issues and creating confusion for maintainers and users alike.

### The Three Conflicting Paths

| Layer | Path | Source |
|-------|------|--------|
| Runtime plugin behavior | `node_modules/superpowers/skills/` | `.opencode/plugins/superpowers.js` registers `../../skills` relative to install dir |
| Bootstrap help text | `~/.config/opencode/skills/superpowers/` | Shown to user in OpenCode startup logs |
| Test assertions | `~/.config/opencode/superpowers/skills/` | `test/opencode.test.ts` checks this path |

### Root Cause Analysis

The root cause is that **three different hardcoded path references** were introduced at different times without a shared constant:

**1. Runtime — `plugins/superpowers.js`:**
```javascript
// Auto-registration uses package-local path
const skillsPath = path.resolve(__dirname, '../../skills');
// This correctly resolves to: node_modules/superpowers/skills/
```

**2. Bootstrap text — shown in OpenCode startup:**
```
Superpowers skills: ~/.config/opencode/skills/superpowers/
```
This path was never updated after the auto-register change. It implies manual installation layout.

**3. Test assertions — `test/opencode.test.ts`:**
```typescript
const expectedPath = path.join(os.homedir(), '.config/opencode/superpowers/skills');
// Note: .opencode/superpowers/skills (no /skills/ subdir!)
// vs runtime: node_modules/superpowers/skills
```

The test path has **two structural differences** from the runtime path:
- Uses `~/.config/opencode/` (manual install) instead of `node_modules/` (package install)
- Missing `/skills` subdirectory: `superpowers/skills` vs `superpowers/skills/...`

### Impact

1. **Users** follow the bootstrap text, set up manual install at wrong path, wonder why skills don't activate
2. **Maintainers** cannot write reliable installation verification because the test path doesn't match runtime behavior
3. **Debugging** requires tracing three different files to understand where skills actually come from

### Proposed Fix

**Step 1:** Unify on a single path source — the package's bundled skills path should be canonical:

```javascript
// In plugins/superpowers.js — already correct:
const skillsPath = path.resolve(__dirname, '../../skills');

[...]

PR #1447, opened 2026-04-30
fix(opencode): adopt v1 default export to fix Plugin export error

pratyush618:

Issue #1444: A user (lxg07254698) on OpenCode 1.14.30 cannot install Superpowers via the documented one-liner. With

```json
{ "plugin": ["superpowers@git+https://github.com/obra/superpowers.git"] }
```

OpenCode rejects the plugin at startup:

```
ERROR  service=plugin path=superpowers error=Plugin export is not a function failed to load plugin
```

The plugin never registers — no skills, no bootstrap, no hooks.

Root cause is the export shape in `.opencode/plugins/superpowers.js`. OpenCode's plugin loader (`packages/opencode/src/plugin/index.ts` + `shared.ts` in sst/opencode v1.14.30) tries two paths in order:

1. **v1 (preferred)** — \`readV1Plugin\` looks for \`mod.default\` to be an object containing at least one of \`id\` / \`server\` / \`tui\`. If found, it calls \`default.server(input, options)\`.
2. **Legacy fallback** — \`getLegacyPlugins\` iterates \`Object.values(mod)\` and throws \`TypeError("Plugin export is not a function")\` if any export isn't a function or \`{ server: function }\`.

Our plugin had no default export and one named export (\`export const SuperpowersPlugin\`). It happened to load via the legacy path on Node/Bun on Linux locally, but the legacy iteration is the brittle path — any non-function export added later silently breaks the plugin, and the affected user's environment is hitting that throw today.

## What does this PR change?

Converts \`.opencode/plugins/superpowers.js\` to OpenCode's documented v1 plugin shape: a single \`export default { id: "superpowers", server }\`. The plugin function body and returned hooks (\`config\`, \`experimental.chat.messages.transform\`) are unchanged — only the export shape moves.

Yes. \`.opencode/plugins/superpowers.js\` is the core OpenCode harness entrypoint that ships with Superpowers. The fix benefits every OpenCode user installing Superpowers via the documented method. It is not project-specific, not a third-party integration, not a new skill.

- **Leave the named export, add a default that aliases it** — works but having both exports makes the legacy iteration path double-invoke the plugin (it would push the plugin into the hooks list twice once it walked both \`mod.default.server\` and \`mod.SuperpowersPlugin\`). Safe in v1.14.30 because v1 detection short-circuits before legacy runs, but leaves a footgun for any pure-legacy loader. Rejected as not "production grade."
- **Patch package.json \`main\` to a thin v1 wrapper** — extra indirection, two files to keep in sync, no behavioral upside.
- **Default-export the function directly (\`export default async () => {...}\`)** — \`mod.default\` would be a function, not a record, so \`readV1Plugin\` would skip it and we'd be back on the legacy path. Defeats the purpose.

The chosen approach is the only one that puts the plugin on the deterministic v1 path with a stable id and zero runtime overhead.

No. One file, one logical change, split into two commits per the contributor's commit hygiene preference: (1) add the v1 default export alongside the named export so v1.14.30 starts working immediately, (2) drop the now-unused named export. The intermediate commit is independently safe — v1 detection picks the default and the legacy path is never entered.

## Existing PRs
- Related PRs: none found.

\`gh search\` for "Plugin export is not a function", "OpenCode plugin export", and "plugin failed to load" against this repo returned only #1444 itself plus unrelated work (OpenCode docs cleanups #305, #608, #847, #890; Windows path bug #1242; \`config\` hook silently ignored #1087; bootstrap caching #1232). None modify the plugin's export shape.

## Environment tested

| OpenCode plugin loader (replayed locally per sst/opencode v1.14.30 source) | v1.14.30 (loader logic, Bun 1.3.11 host) | n/a (loader-level fix, not a model-behavior change) | n/a |

I do not have an OpenCode 1.14.30 install here to run end-to-end against the issue reporter's exact setup. The verification I ran instead replays the loader's exact \`readV1Plugin\` + \`getLegacyPlugins\` logic from \`sst/opencode@v1.14.30\` against the modified plugin module, on both Node 22 and Bun 1.3.11.

## Evaluation

This is a loader-level export-shape fix, not a behavior or skill change, so traditional eval sessions don't apply. The "before/after" is binary: the plugin either loads or throws \`TypeError\`. Verification:

- **Before** (current main, \`export const SuperpowersPlugin\`): \`Object.values(mod) = [AsyncFunction]\` — passes legacy iteration locally on Linux/Node/Bun, but the issue reporter's environment trips the same iteration's throw.
- **After** (\`export default { id, server }\`): \`mod.default\` is a record with \`id\` + \`server\`, \`readV1Plugin\` returns it, the v1 path is taken, legacy iteration is never entered. I confirmed both v1 detection and legacy fallback succeed against the new module by replaying the loader code:
  \`\`\`
  v1 detected: true { id: "superpowers", server: "function" }
  legacy fallback also OK: 1 plugin(s)
  \`\`\`
- \`bash tests/opencode/test-plugin-loading.sh\` — all 6 tests pass.
- \`node --check .opencode/plugins/superpowers.js\` — passes.
- \`npm pack\` + install of the resulting tarball into a fresh project then \`await import('superpowers')\`: \`m.default = { id: 'superpowers', server: [AsyncFunction] }\`, no leftover named exports.

## Rigor

- [x] If this is a skills change: I used \`superpowers:writing-skills\` and completed adversarial pressure testing (paste results below) — **N/A, no skill content modified.**
- [x] This change was tested adversarially, not just on the happy path — replayed the exact loader logic that throws the user's error, and verified both the v1 path and legacy fallback accept the new shape.
- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) without extensive evals showing the change is an improvement — no skill content touched.

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -46,7 +46,7 @@ const normalizePath = (p, homeDir) => {
   return path.resolve(normalized);
 };
 
-export const SuperpowersPlugin = async ({ client, directory }) => {
+const server = async ({ client, directory }) => {
   const homeDir = os.homedir();
   const superpowersSkillsDir = path.resolve(__dirname, '../../skills');
   const envConfigDir = normalizePath(process.env.OPENCODE_CONFIG_DIR, homeDir);
@@ -113,5 +113,5 @@ ${toolMapping}
 
 export default {
   id: "superpowers",
-  server: SuperpowersPlugin,
+  server,
 };
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -110,3 +110,8 @@ ${toolMapping}
     }
   };
 };
+
+export default {
+  id: "superpowers",
+  server: SuperpowersPlugin,
+};


obra:
What I read from this PR is that you made the change blind and have not tested it. I am unable to reproduce the original problem. I'm going to close this PR for now. If you can prove that you reproduced the user's problem, please feel free to reopen it.

PR #1457, opened 2026-05-02
Make `using-superpowers` manual-invoke (disable-model-invocation: true)

### LINKED ISSUE #1456: Make `using-superpowers` opt-in to auto-load instead of injecting on every session
## Summary

`skills/using-superpowers/SKILL.md` currently auto-injects its full ~5.4KB / ~1,300 tokens content as an `<EXTREMELY_IMPORTANT>` system block at the start of every Claude Code conversation, regardless of project context or whether any superpowers skills will actually be used in the session. This happens because the description (`"Use when starting any conversation..."`) is broad enough to match every conversation start.

For users who actively use superpowers skills (TDD, brainstorming, debugging-systematically, etc.), this still feels like avoidable overhead. The other skills are discoverable via the standard skill registry — Claude can find and invoke them through the `Skill` tool without the meta-skill needing to be in context first.

## Why I'm flagging this

The v5.0.7 release notes show you already care about this kind of cost:

> **Bootstrap as user message** — moved bootstrap injection from `experimental.chat.system.transform` to `experimental.chat.messages.transform`, prepending to the first user message instead of adding a system message. **Avoids token bloat from system messages repeated every turn (#750)**

That fix landed for OpenCode. The Claude Code path still has the always-inject behavior.

## Proposed fix

Add `disable-model-invocation: true` to the frontmatter:

```yaml
---
name: using-superpowers
description: Meta-skill for understanding how to find and use other superpowers skills. Invoke explicitly when needed.
disable-model-invocation: true
---
```

This makes the skill manual-only per the [Claude Code skills documentation](https://code.claude.com/docs/en/skills.md):
- Claude can still discover it in the skill registry
- Users can `/invoke superpowers:using-superpowers` if they want the meta-instructions
- Other superpowers skills (TDD, brainstorming, etc.) keep working — they're discovered through the standard skill registry, not through this meta-skill
- Token cost moves from "every session" to "when explicitly needed"

## Comparison: how other plugins handle this

The Vercel plugin's `knowledge-update` skill uses conditional injection:

```yaml
metadata:
  sessionStart: true
  pathPatterns:
    - 'vercel.ts'
```

It only injects when there's actually a `vercel.ts` in the project. That's the modern conditional pattern — `using-superpowers` could follow the same approach (e.g., inject only when project shows signs of code work — `.git` present, code files exist, etc.) if you'd rather keep some auto-injection.

## Alternative

If first-time-user discoverability is the concern, replace the full inject with a tiny pointer:

> "Run `/superpowers:using-superpowers` for guidance on finding and using superpowers skills."

That's ~20 tokens instead of 1,300, and still surfaces the skill for new users.

Happy to send a PR if any of these directions sound good.

explicit09:
Closes #1456

## What

Adds `disable-model-invocation: true` to `skills/using-superpowers/SKILL.md` and updates the description to reflect that the skill is now manually-invoked.

## Why

`using-superpowers` currently auto-injects its full ~5.4KB / ~1,300 tokens content into the system context at the start of every Claude Code session because its description (`"Use when starting any conversation..."`) matches every conversation. For users who actively use superpowers skills (TDD, brainstorming, debugging-systematically, etc.), the meta-skill content doesn't need to be in context to find the other skills — they're discoverable via the standard Skill registry.

The v5.0.7 release notes show you already care about this kind of cost (the OpenCode `experimental.chat.messages.transform` fix for #750). This applies the same principle to Claude Code.

## Trade-off (worth flagging)

Claude no longer sees the "MUST invoke Skill tool BEFORE any response" rule and the red-flag table by default. Users who want the enforcement back can invoke the skill explicitly:

```
/superpowers:using-superpowers
```

If you'd rather preserve some default guidance, two alternatives I floated in #1456:

1. **Replace the full inject with a 1-line pointer** (~20 tokens):
   > "Run `/superpowers:using-superpowers` for guidance on finding and using superpowers skills."
2. **Use `sessionStart: true` + `pathPatterns`** (the pattern Vercel's `knowledge-update` uses) so it only injects when the project shows signs of code work.

Happy to adjust this PR if you'd prefer either alternative.

## Verification

- The other superpowers skills (TDD, brainstorming, debugging, etc.) still appear in the standard skill registry and remain invokable via the Skill tool — confirmed in a fresh session.
- `disable-model-invocation: true` is documented at https://code.claude.com/docs/en/skills.md as the standard way to mark a skill manual-only.

### DIFF
--- a/skills/using-superpowers/SKILL.md
+++ b/skills/using-superpowers/SKILL.md
@@ -1,6 +1,7 @@
 ---
 name: using-superpowers
-description: Use when starting any conversation - establishes how to find and use skills, requiring Skill tool invocation before ANY response including clarifying questions
+description: Reference for how to find and use superpowers skills via the Skill tool. Invoke explicitly to load the skill-discovery rules and red-flag table.
+disable-model-invocation: true
 ---
 
 <SUBAGENT-STOP>


explicit09:
Closing this — I should have searched for prior art before submitting. #1019 and #459 already covered very similar ground, and #1019 in particular makes clear the always-load behavior is considered intentional, not a bug to fix with a config flag.

Apologies for the noise. Issue #1456 stays open since it proposes a couple of alternatives that aren't covered by the prior PRs (`sessionStart` + `pathPatterns` for conditional auto-injection, or a tiny pointer-only inject) — but feel free to close that too if the position is settled.

PR #1493, opened 2026-05-07
test(skill-triggering): add pipefail so failed tests fail run-all.sh

Nandakumar-M:
Fixing an issue with `run-all.sh`. If any of the `run-test.sh` failed, run-all.sh still showed all tests passed.

run-all.sh used `if cmd | tee log; then` to record per-skill output. Without `pipefail`, the pipeline's exit status is tee's (always 0), so failed tests were rolled up as passed in the summary.

Adding `set -eo pipefail` makes the pipeline fail when run-test.sh fails, which propagates to the if/else and into the PASSED/FAILED counters.

### DIFF
--- a/tests/skill-triggering/run-all.sh
+++ b/tests/skill-triggering/run-all.sh
@@ -2,7 +2,7 @@
 # Run all skill triggering tests
 # Usage: ./run-all.sh
 
-set -e
+set -eo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PROMPTS_DIR="$SCRIPT_DIR/prompts"


obra:
Thanks for the careful read on the original bug — the \`if cmd | tee log; then\` pattern would indeed swallow the inner exit status without \`pipefail\`.

That specific file (\`tests/skill-triggering/run-all.sh\`) was deleted on \`dev\` in 371f4159 ("tests: remove skill-triggering bash prompts (covered by drill triggering-* scenarios)") on 2026-05-06. The replacement runner at \`tests/explicit-skill-requests/run-all.sh\` doesn't pipe through tee — it uses \`if "$SCRIPT_DIR/run-test.sh" ...; then\` directly, so the pipefail issue doesn't apply.

I also checked: no other test runner in this repo uses \`pipefail\`; they all use plain \`set -e\`. Adding pipefail to one runner unilaterally would create inconsistency.

Your PR targeted \`main\`, where the old file still exists pending the next release. Closing as obsolete-on-dev.

PR #1520, opened 2026-05-11
fix: remove stale agents/commands refs from .cursor-plugin/plugin.json

### LINKED ISSUE #1508: using-git-worktrees: Step 2 is missing — workflow jumps from Step 1 to Step 3
The numbered workflow in `skills/using-git-worktrees/SKILL.md` skips `## Step 2`. The headings present are:

```
Line  16: ## Step 0: Detect Existing Isolation
Line  47: ## Step 1: Create Isolated Workspace
Line 114: ## Step 3: Project Setup
Line 133: ## Step 4: Verify Clean Baseline
```

There is no `## Step 2` between Step 1 and Step 3. An agent following the skill in order will read "Step 1 → Step 3" and either (a) wonder if a step was missed or (b) silently skip whatever Step 2 was supposed to do.

### Suggested fix

Either renumber Step 3, Step 4 → Step 2, Step 3, OR add the missing Step 2 if a step was unintentionally dropped during a refactor.

### LINKED ISSUE #1514: .cursor-plugin/plugin.json references deleted agents/ and commands/ directories

## Environment

| Field | Value |
|-------|-------|
| Superpowers version | 5.1.0 (`f2cbfbe`) |
| Harness | n/a — static manifest consistency issue |
| Harness version | n/a |
| Model | n/a |
| OS + shell | n/a |

This is a static repository-state issue, not a runtime bug. It is filed against this repo because the affected file (`.cursor-plugin/plugin.json`) is owned here.

## What happened?

`.cursor-plugin/plugin.json` at HEAD (v5.1.0) still references two directories that were deleted in v5.1.0:

```json
"skills": "./skills/",
"agents": "./agents/",     ← directory deleted (PR #1299)
"commands": "./commands/", ← directory deleted (PR #1188)
"hooks": "./hooks/hooks-cursor.json"
```

For comparison, the new `.codex-plugin/plugin.json` introduced in v5.1.0 correctly omits both `agents` and `commands`, so the omission appears intentional — but the equivalent cleanup of `.cursor-plugin/plugin.json` was missed.

## Evidence

```sh
$ git ls-tree HEAD agents commands
(empty — directories don't exist)

$ grep -E '"(agents|commands)"' .cursor-plugin/plugin.json
  "agents": "./agents/",
  "commands": "./commands/",
```

`git blame .cursor-plugin/plugin.json` shows both lines were introduced in commit `772ec9f8` (2026-02-13, when Cursor support was first added) and have not been touched since — they were not removed alongside the v5.1.0 directory deletions in #1188 and #1299.

## Steps to reproduce

```sh
git clone --depth 1 https://github.com/obra/superpowers.git
cd superpowers
ls agents commands 2>&1  # ls: cannot access 'agents': No such file or directory
grep -E '"(agents|commands)"' .cursor-plugin/plugin.json  # both still present
```

## Expected behavior

The Cursor manifest should not reference directories that do not exist, mirroring the (correct) state of `.codex-plugin/plugin.json`.

## Actual behavior

The two stale `agents` / `commands` field entries persist. Cursor appears to silently skip missing directories at runtime (no user reports), so this is not user-visible — but the stale references:

- Mislead readers / future maintainers
- Are inconsistent with `.codex-plugin/plugin.json`
- Could break if a stricter manifest validator is later introduced

## Suggested fix

Two-line deletion:

```diff
   "skills": "./skills/",
-  "agents": "./agents/",
-  "commands": "./commands/",
   "hooks": "./hooks/hooks-cursor.json"
```

Happy to open a PR with just this change if it's useful.

n/a — this is a static manifest issue, no runtime transcript applies. Reproducer above is fully self-contained.

fuleinist:
## Summary

Remove stale `agents` and `commands` entries from `.cursor-plugin/plugin.json`.

Those directories no longer exist, and the Cursor manifest should match the current shipped plugin surface instead of pointing at deleted paths.

## Issue

Addresses #1514

## Testing

- `jq empty .cursor-plugin/plugin.json`
- `git diff --check`

### DIFF
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -19,7 +19,5 @@
     "workflows"
   ],
   "skills": "./skills/",
-  "agents": "./agents/",
-  "commands": "./commands/",
   "hooks": "./hooks/hooks-cursor.json"
 }


arittr:
Thanks for the PR. I narrowed this to the issue described in the title: removing stale `agents` / `commands` references from `.cursor-plugin/plugin.json`.

We have a one-issue-one-PR rule in this repo, so I removed the unrelated `using-git-worktrees` Step 2 renumbering change from this branch. #1508 is still valid, but it should be handled in its own focused PR.

arittr (approved):
LGTM, thx!

PR #1001, opened 2026-03-30
docs: document all agentskills.io frontmatter fields in writing-skills

lucianghinda:
The `writing-skills` skill said only `name` and `description` were supported in SKILL.md frontmatter. The [agentskills.io spec](https://agentskills.io/specification) supports 4 more optional fields — this PR documents them.

## Changes

- `SKILL.md` and `anthropic-best-practices.md`: replace "only two fields" with a full field table (`license`, `compatibility`, `metadata`, `allowed-tools`)
- Fixed example `name` to lowercase (spec requires it)
- Bumped version to `6.0.1`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #196, opened 2025-12-29
docs(writing-skills): fix incomplete frontmatter field documentation

ivan-magda:
Resolves #195

## Motivation and Context
The writing-skills skill incorrectly documented that YAML frontmatter supports "only two fields" (`name` and `description`). This caused Claude to incorrectly suggest removing valid `allowed-tools` and `model` fields when reviewing existing skills. The official Claude Code documentation at https://code.claude.com/docs/en/skills#available-metadata-fields confirms four valid fields exist.

## How Has This Been Tested?
Verified against official Claude Code documentation. The fix aligns the skill documentation with the authoritative source.

## Breaking Changes
None. This is purely a documentation correction that makes the skill more permissive (no longer incorrectly flagging valid fields).

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist
- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context
Changes made:
- Line 96-97: Changed "Only two fields supported" to list required (`name`, `description`) and optional (`allowed-tools`, `model`) fields separately
- Line 607: Updated checklist item to document all valid frontmatter fields while preserving the max 1024 chars note

## Summary by CodeRabbit

* **Documentation**
  * Clarified skill documentation frontmatter specifications by explicitly designating required fields (name, description) and optional fields (allowed-tools, model), enhancing validation guidance for skill documentation contributors.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1533, opened 2026-05-12
docs(writing-skills): add allowed-tools sync invariant to frontmatter docs and checklist

daniel769:

When a skill references another skill via `Skill(X)` in its body but that skill is absent from the `allowed-tools:` frontmatter field, the agent silently fails to invoke it — no error is thrown, no hook catches it, the call just doesn't happen. This was caught in production at Honeydew by a cursor bugbot on a PR review; nothing in local tooling surfaced it first. The `writing-skills` skill has no mention of this invariant, so skill authors have no reason to know it exists.

## What does this PR change?

Adds two things to `writing-skills/SKILL.md`:
1. A `allowed-tools` bullet under the Frontmatter section (matching the style of the existing `name`/`description` entries) explaining the invariant and providing a one-liner to verify it.
2. A corresponding GREEN-phase checklist item so authors are prompted to check it during skill creation.

Yes. Any skill that calls other skills is affected, regardless of domain or project. The failure mode is universal and invisible, which makes it a good fit for the shared checklist.

Adding it only to the checklist (without the frontmatter prose) — but the checklist is scanned quickly and the prose section is where authors learn the semantics of each field, so both felt necessary.

No. Two lines in one file, both documenting the same invariant.

## Existing PRs
- Related PRs: #1001 (closed), #927 (closed), #196 (closed)

PRs #1001, #927, and #196 all attempted to enumerate all supported frontmatter fields comprehensively. This PR does not enumerate fields — it documents a specific silent failure mode that those PRs did not address (body/frontmatter mismatch causing `Skill()` calls to silently fail).

## Environment tested

| Claude Code | latest | Claude Sonnet | claude-sonnet-4-6 |

## Evaluation

- Session that led to this: a Honeydew PR where cursor bugbot flagged that a `Skill(X)` reference in a skill body had been removed from `allowed-tools`, breaking the call silently.
- After the change: the checklist item surfaces during normal skill authoring; the prose entry is visible when reading the frontmatter section.
- No behavioral change to any agent workflow — this is additive documentation only.

## Rigor

This is a documentation addition (two new lines) that does not touch any behavior-shaping content.

## Human review

### DIFF
--- a/skills/writing-skills/SKILL.md
+++ b/skills/writing-skills/SKILL.md
@@ -96,6 +96,9 @@ skills/
 - Two required fields: `name` and `description` (see [agentskills.io/specification](https://agentskills.io/specification) for all supported fields)
 - Max 1024 characters total
 - `name`: Use letters, numbers, and hyphens only (no parentheses, special chars)
+- `allowed-tools`: List every skill called via `Skill(X)` in the body
+  - Omitting one causes a silent runtime failure — no error, agent simply can't invoke the skill
+  - Verify: `grep -oE 'Skill\([^)]+\)' SKILL.md`
 - `description`: Third-person, describes ONLY when to use (NOT what it does)
   - Start with "Use when..." to focus on triggering conditions
   - Include specific symptoms, situations, and contexts
@@ -605,6 +608,7 @@ Deploying untested skills = deploying untested code. It's a violation of quality
 **GREEN Phase - Write Minimal Skill:**
 - [ ] Name uses only letters, numbers, hyphens (no parentheses/special chars)
 - [ ] YAML frontmatter with required `name` and `description` fields (max 1024 chars; see [spec](https://agentskills.io/specification))
+- [ ] `allowed-tools` lists every skill called via `Skill(X)` in the body
 - [ ] Description starts with "Use when..." and includes specific triggers/symptoms
 - [ ] Description written in third person
 - [ ] Keywords throughout for search (errors, symptoms, tools)


obra:
Hi @daniel769 — I'm Claude, an AI agent (Fable 5, running in Claude Code), posting from Jesse (@obra)'s account at his direction. Jesse had me run a skeptical triage of all 340 open superpowers items — every factual claim tested against the current `dev` tree, then adversarially re-checked by a second, independent agent — and he reviewed the verdicts and directed this batch of closures. This one is closing because the submission itself didn't hold up under those checks — specifics below, with the exact tests run:

Thanks for the thorough template and for disclosing your setup. I tested this directly: two skills where the outer skill's body invokes a second skill by name, with neither one listed in `allowed-tools` anywhere, run against current Claude Code (v2.1.202) in default permission mode. Twice in a row the second skill invoked and ran with zero permission prompts or denials — no silent failure. That matches the current docs, which say `allowed-tools` only pre-approves prompts: "It does not restrict which tools are available: every tool remains callable" (code.claude.com/docs/en/skills). Separately, `Skill(X)` is permission-rule syntax used in `allowed-tools`/`disallowed-tools`/`/permissions` — it isn't something a skill body writes to invoke another skill (none of this repo's 14 skills do this; they cross-reference by prose name), so the proposed `grep -oE 'Skill\(...\)'` check wouldn't find anything to sync here even if the premise held. I can't verify what happened in the Honeydew session that prompted this, but as a general rule it doesn't hold up on current Claude Code, so I'm closing rather than merging. Happy to take another look if you can attach a literal transcript reproducing a silent failure on a current version. Also, unrelated to the above: this PR targets `main` — everything here should target `dev`.

---

PR #1297, opened 2026-04-28
fix(skills): promote worktree step into executor process flows

fisherkuan:
## What problem are you trying to solve?

Closes part of [#1080](https://github.com/obra/superpowers/issues/1080).

In `executing-plans` and `subagent-driven-development`, `using-git-worktrees` was only mentioned in the trailing `## Integration` prose footer. The model reliably skips it there, because `using-superpowers` instructs it to follow the structural Process (numbered steps / graphviz) — and prose under a sibling `## Integration` heading sits outside that structure. Issue #1080 includes a real session log (Claude Code 2.1.107, superpowers 5.0.7) where `using-git-worktrees` was never invoked when running `subagent-driven-development`; `cwd` stayed in the main repo on `main`.

This is the silent-skip failure mode that PR [#1097](https://github.com/obra/superpowers/pull/1097) does not address — that PR is scoped to `brainstorming/SKILL.md` only.

## What does this PR change?

Both executor skills now have a `Step 0: Set Up Isolated Workspace` inside the structural Process, before plan loading / task extraction. Consent handling is delegated to `using-git-worktrees` itself, which already does the right thing as of [#1121](https://github.com/obra/superpowers/pull/1121) (Step 0 detection + Step 1a consent prompt) — so this PR is consistent with #991's resolution and does not re-introduce auto-create behavior.

- `skills/executing-plans/SKILL.md`: insert `Step 0` before `Step 1: Load and Review Plan`; add a matching `Remember` bullet.
- `skills/subagent-driven-development/SKILL.md`: insert `Step 0` ahead of the Process graphviz; add a `Use superpowers:using-git-worktrees` node (lightgreen, mirroring the `finishing-a-development-branch` terminal node) and an edge into `Read plan, extract all tasks…`; mention `Step 0` in the worked example; add a `Red Flag` against skipping it.

The trailing `## Integration` footers are unchanged — this PR only promotes the reference into the enforceable structure.

## Is this change appropriate for the core library?

Yes. Both skills are core executors used across harnesses, and the silent-skip behavior is reproducible against the unmodified plugin. The fix is wording-only and stays consistent with the consent contract owned by `using-git-worktrees`.

## What alternatives did you consider?

1. **Bolder wording in the Integration footer** — rejected. The session evidence in #1080 shows the issue is structural, not lexical: `using-superpowers` tells the model to follow the structural Process, so anything in the Integration footer keeps losing.
[...]

PR #829, opened 2026-03-19
fix: restore worktree creation step in brainstorming skill

mzonghao:
## Summary

- In v3.1.0, brainstorming had a "Phase 4: Worktree Setup" between design approval and planning handoff. When the spec doc writing step was added in a later version, this worktree creation step was dropped, causing cross-reference inconsistencies across three skills:
  - `writing-plans` referenced "dedicated worktree (created by brainstorming skill)" but brainstorming had no worktree step
  - `using-git-worktrees` referenced "brainstorming (Phase 4)" but that phase no longer existed
- Tracing the git history reveals the evolution:
  - v3.1.0: brainstorming had an explicit Phase 4 (Worktree Setup) and Phase 5 (Planning Handoff)
  - v3.4.1 (8e38ab8 "Simplify brainstorming skill"): Phase structure removed, but worktree reference kept in prose
  - v4.3.0 (7f2ee61 "Enforce brainstorming workflow with hard gates"): worktree reference fully removed, replaced with "Do NOT invoke any other skill. writing-plans is the next step."
However, the cross-references in other files were not updated to match.
- **Fix:** Re-insert worktree creation as step 6 in brainstorming checklist (after design approval, before writing spec doc), so that spec, plan, and implementation all live in the same worktree branch. Updated cross-references in `writing-plans` and `using-git-worktrees` accordingly.

## Changes

| File | Change |
|------|--------|
| `skills/brainstorming/SKILL.md` | Added step 6 "Create worktree" between design approval and spec writing; updated dot graph |
| `skills/writing-plans/SKILL.md` | Updated context line to clarify worktree is created by brainstorming before writing spec |
| `skills/using-git-worktrees/SKILL.md` | Removed stale "(Phase 4)" reference, updated to "after design approval, before writing spec doc" |

## Test plan

- [ ] Verify brainstorming checklist steps are correctly numbered 1-10
- [ ] Verify dot graph renders correctly with the new worktree node
- [ ] Verify cross-references between the three skills are consistent

PR #271, opened 2026-01-15
Add native Cursor IDE integration

mingyooagi:
Add native Cursor IDE integration for Superpowers using Agent Skills with one-command npx installation.

## Motivation and Context

Cursor IDE now supports [Agent Skills](https://cursor.com/cn/docs/context/skills), an open standard for packaging reusable AI agent workflows. This PR adds native Cursor support to Superpowers, making it as easy to install as:

```bash
npx github:obra/superpowers/.cursor install --global
```

**Why this is needed:**
- Cursor users can now benefit from all Superpowers skills (brainstorming, TDD, systematic debugging, etc.)
- Maintains architectural consistency with existing Codex and OpenCode implementations
- Uses the same shared `lib/skills-core.js` module for skill discovery and parsing
- Provides a modern, cross-platform installation experience

## How Has This Been Tested?

**Tested scenarios:**
- ✅ Global installation (`~/.cursor/superpowers`)
- ✅ Local installation (`.cursor-superpowers/`)
- ✅ Skill discovery using `skillsCore.findSkillsInDir()`
- ✅ Symlink creation on macOS/Linux
- ✅ Junction fallback on Windows
- ✅ List command showing both superpowers and custom skills
- ✅ Uninstall cleaning up all symlinks/junctions
- ✅ Handling of existing directories (safe overwrite)
- ✅ CLI error handling and user prompts

**Manual verification:**
- Skills appear in Cursor Settings → Rules → Agent Decides
- Skills auto-invoke based on context
- Manual skill invocation via `/` command works

## Breaking Changes

None. This is a new platform integration that doesn't affect existing Codex or OpenCode implementations.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

### Architecture

All three platforms now share the same core:

```
lib/skills-core.js (shared)
├── Cursor:   .cursor/cli.js (npx installer)
├── Codex:    .codex/superpowers-codex (CLI script)
└── OpenCode: .opencode/plugin/superpowers.js (runtime plugin)
```

**Key features:**
[...]

Issue #270, opened 2026-01-15
How can I use Superpowers in Antigravity


galanfish:
How to use Superpowers with Antigravity?

xianshi122:
I've successfully verified and confirmed the effectiveness of the process. Here's a brief guide on how to use Superpowers in Antigravity :
1  Download Skill Files: Visit GitHub, search for the obra/superpowers repository, and locate the skills directory containing various Markdown files (e.g., brainstorming.md, test_driven_development.md).
2 Import into Antigravity: Copy these Markdown files into your **project's** .agent/skills/ directory.
Activate: Trigger the skills in conversation, for example: "Let's use the TDD superpower to fix this bug" or "Please follow the brainstorming protocol."

Please ensure these files are correctly loaded into your .agent/skills/ directory. If so, you're all set to activate these superpowers.
This response was generated by an AI, and I've confirmed its success and effectiveness.
你可以问你的Agent，《请把你的所有skill文件例举出来，我看加载的是否有问题》

galanfish:
> I've successfully verified and confirmed the effectiveness of the process. Here's a brief guide on how to use Superpowers in Antigravity : 1 Download Skill Files: Visit GitHub, search for the obra/superpowers repository, and locate the skills directory containing various Markdown files (e.g., brainstorming.md, test_driven_development.md). 2 Import into Antigravity: Copy these Markdown files into your **project's** .agent/skills/ directory. Activate: Trigger the skills in conversation, for example: "Let's use the TDD superpower to fix this bug" or "Please follow the brainstorming protocol."
> 
> Please ensure these files are correctly loaded into your .agent/skills/ directory. If so, you're all set to activate these superpowers. This response was generated by an AI, and I've confirmed its success and effectiveness. 你可以问你的Agent，《请把你的所有skill文件例举出来，我看加载的是否有问题》

THX a ton!!!! This is super helpful!
One more quick question, I saw folders like commands/, agents/, lib/, hooks/, etc. in the repo. Not totally sure what they’re for.
Do I need to care about these or set them up manually, or can I just ignore them?

rubgithub:
https://github.com/anthonylee991/gemini-superpowers-antigravity

cwest:
Addressed by #499.

skainguyen1412:
[...]

PR #1562, opened 2026-05-16
fix(using-git-worktrees): lockfile-aware Step 3 (pnpm/yarn/bun + uv)

iKon85:

`using-git-worktrees` Step 3 (\"Project Setup\") is hardcoded to `npm install` whenever a `package.json` is present:

\`\`\`bash
# Node.js
if [ -f package.json ]; then npm install; fi
\`\`\`

In pnpm or yarn workspaces this is wrong. `npm install`:

- ignores the `workspace:*` protocol that pnpm/yarn lockfiles declare,
- does **not** create `node_modules/<pkg>` symlinks for sibling workspace packages,
- can write a `package-lock.json` next to an existing `pnpm-lock.yaml` / `yarn.lock`, leaving the tree in a hybrid state.

**Failure mode is silent.** Tools that resolve through `node_modules` (LSP, tsserver, ts-node, type-aware refactors) get a partial module graph and return partial results. A `findReferences` on a workspace-shared type comes back with fewer matches than reality. Nothing logs an error; the agent gets a wrong answer and proceeds.

**Empirical repro** (real pnpm workspace, two packages with one importing the other via package name, the other via tsconfig path alias):

| Worktree state | LSP findReferences |
|---|---|
| Main tree (warm, baseline) | 94 refs / 29 files |
| Fresh worktree, no setup | 54 refs / 19 files |
| Same worktree after `pnpm install` | 94 refs / 29 files |

The 40 missing refs were all in the package importing via the workspace package name (`@scope/shared`). With no `node_modules/@scope/shared` symlink, tsserver couldn't load that project into the reference search at all. `npm install` from the current Step 3 would have produced the same broken state plus a stray `package-lock.json`.

**Initial prompt that led to this change:** \"Arbeite an #271 — using-git-worktrees-Skill pnpm/yarn-aware machen. #270 hat empirisch gezeigt, dass das in pnpm-Workspaces zu LSP-Bruch führt.\" (Tracking in the downstream project: iKon85/Testreporter#270, iKon85/Testreporter#271.)

## What does this PR change?

Detects the package manager from the lockfile and runs the matching install. Falls back to npm when only `package.json` is present. Same lockfile-aware shape applied to the Python block (uv → poetry → poetry → pip), because the original ran both pip and poetry when `requirements.txt` and `pyproject.toml` coexisted.

Yes. The skill already commits to auto-detecting and running setup; this PR fixes the detection so it matches what the project actually uses. The change is not project-specific, team-specific, or tool-specific — it makes a generic skill correct in the common case of pnpm/yarn/bun workspaces, which the skill currently misbehaves on. No new third-party integration.

1. **Project-local override only (no upstream change).** Considered and shipped *in parallel* in the downstream project as a CLAUDE.md bridge rule, since the user-instruction layer outranks plugin skills. But that only helps one project — the skill is wrong for every pnpm/yarn user, so the bridge is not a substitute for fixing it here.
2. **Runtime detection via `command -v pnpm` / `which yarn`.** Rejected: a developer can have `pnpm` installed globally and still use npm in a given repo; the lockfile is the authoritative signal for what *this* project uses.
3. **Per-corepack handling.** Rejected as out of scope: corepack opt-in is a separate setup concern; lockfile detection works whether or not corepack is enabled.
4. **Leave Node.js alone and only touch Python.** Rejected — Node.js is the documented failure case.

The Node.js block is the primary change. The Python block applies the same lockfile-aware pattern and fixes an adjacent bug (the original runs `pip install` *and* `poetry install` if both `requirements.txt` and `pyproject.toml` exist). If reviewers prefer to land Node.js alone, I'll happily split out Python into a follow-up — let me know.

## Existing PRs
- Related PRs: none directly. The closest worktree-skill changes (#1522 Step 2 numbering, #1297 worktree-step promotion, #829/#1097 worktree-handoff in brainstorming) are workflow-order fixes, not Step-3 install changes.

## Environment tested

|---|---|---|---|
| Claude Code | latest CLI as of 2026-05-16 | Opus 4.7 | claude-opus-4-7 (1M context) |

## Evaluation

- Initial prompt above produced the design (CLAUDE.md bridge + upstream PR split).
- **Eval sessions after the change: 1.** I didn't run a formal adversarial-pressure-test suite (see Rigor below). The empirical evidence is the *baseline*: the failure mode was reproduced on a real pnpm monorepo before this PR was written. After the fix, the same setup script picks `pnpm install`, which we already know produces the correct `node_modules` state in that repo. The fix is mechanical (lockfile selector); the risk surface is narrow.
- If reviewers want a synthetic A/B (subagent in a pnpm sandbox running old Step 3 vs. new Step 3), happy to add it.

## Rigor

- [x] This change was tested adversarially, not just on the happy path — but \"adversarially\" here means *the production failure was the adversarial test*, not a synthetic subagent run.
- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, \"human partner\" language)

Honest disclosure: `superpowers:writing-skills` was invoked, but the formal RED-GREEN-REFACTOR-with-subagent-pressure-tests workflow was deliberately skipped after user agreement, on the reasoning that #270 is the RED, the fix is a 5-line shell selector with no behavioral surface beyond \"pick the right binary,\" and a synthetic pressure scenario wouldn't surface anything the production repro didn't already show. Re-test on request.

## Human review

### DIFF
--- a/skills/using-git-worktrees/SKILL.md
+++ b/skills/using-git-worktrees/SKILL.md
@@ -116,15 +116,27 @@ cd "$path"
 Auto-detect and run appropriate setup:
 
 ```bash
-# Node.js
-if [ -f package.json ]; then npm install; fi
+# Node.js — detect package manager via lockfile.
+# Critical for pnpm/yarn workspaces: `npm install` ignores workspace
+# protocols, won't create `node_modules/<pkg>` symlinks for workspace
+# packages, and can leave the tree in a state where tools that resolve
+# imports through `node_modules` (LSP, type checkers) silently break.
+if   [ -f pnpm-lock.yaml ];     then pnpm install
+elif [ -f yarn.lock ];          then yarn install
+elif [ -f bun.lockb ];          then bun install
+elif [ -f package-lock.json ];  then npm install
+elif [ -f package.json ];       then npm install
+fi
 
 # Rust
 if [ -f Cargo.toml ]; then cargo build; fi
 
 # Python
-if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-if [ -f pyproject.toml ]; then poetry install; fi
+if   [ -f uv.lock ];          then uv sync
+elif [ -f poetry.lock ];      then poetry install
+elif [ -f pyproject.toml ];   then poetry install
+elif [ -f requirements.txt ]; then pip install -r requirements.txt
+fi
 
 # Go
 if [ -f go.mod ]; then go mod download; fi


diskartediana-gif:
Don't access

PR #1563, opened 2026-05-16
fix(docs): add .gitkeep for specs/done and plans/done

### LINKED ISSUE #1547: Add .gitkeep to docs/superpowers/specs/done/ and docs/superpowers/plans/done/
## Problem

The `docs/superpowers/specs/` and `docs/superpowers/plans/` directories have no `done/` subdirectory. Users who establish a workflow of moving completed specs and plans into a `done/` subfolder (at the same level) lose those directories on every plugin update, since the plugin cache is overwritten.

## Proposed fix

Add `.gitkeep` files to include the `done/` subdirs in the repo:

```
docs/superpowers/specs/done/.gitkeep
docs/superpowers/plans/done/.gitkeep
```

This makes the workflow `specs/ → plans/ → done/` (at each level) survive plugin updates without any user intervention.

## Workaround (current)

Users can recreate the dirs after each update with:

```bash
SP=$(ls -d ~/.claude/plugins/cache/claude-plugins-official/superpowers/*/docs/superpowers 2>/dev/null | tail -1)
[ -n "$SP" ] && mkdir -p "$SP/specs/done" "$SP/plans/done"
```

The glob handles version bumps, but it's a manual step that shouldn't be necessary.

gaurav0107:
Fixes #1547.

`docs/superpowers/specs/` and `docs/superpowers/plans/` don't ship with `done/` subdirectories. Users who adopt a `specs/ → plans/ → done/` workflow lose those folders on every plugin update, since the plugin cache is overwritten and Git can't track empty directories. The issue thread has the manual workaround people are running today.

This PR adds two zero-byte `.gitkeep` files so the directories ship with the repo and survive updates:

- `docs/superpowers/specs/done/.gitkeep`
- `docs/superpowers/plans/done/.gitkeep`

Nothing else changes. No code references the new files — `.gitkeep` is just the standard sentinel for tracking otherwise-empty directories.

Verified locally:

```
$ test -f docs/superpowers/specs/done/.gitkeep && echo ok
ok
$ test -f docs/superpowers/plans/done/.gitkeep && echo ok
ok
$ git diff --stat dev...HEAD
 docs/superpowers/plans/done/.gitkeep | 0
 docs/superpowers/specs/done/.gitkeep | 0
 2 files changed, 0 insertions(+), 0 deletions(-)
```


obra:
This is a PR for a thing that is not a superpowers feature.

Issue #1479, opened 2026-05-06
.claude-plugin/plugin.json missing "skills" field — skills not discovered by Claude Code


Jiao-XXX:
## Summary

  The `.claude-plugin/plugin.json` is missing the `"skills"` field, so Claude Code cannot discover or register any of the 14 skills shipped with the plugin. The skills exist on disk
  at `skills/*/SKILL.md` but Claude Code doesn't know to look there.

  ## Steps to Reproduce

  1. Install `superpowers@claude-plugins-official` on Claude Code 2.1.119
  2. Start a new session or `/clear`
  3. Try `/test-driven-development` → **"Unknown command"**
  4. Try invoking via Skill tool → **"Unknown skill: superpowers:test-driven-development"**
  5. None of the 14 skills appear in the available skills manifest

  ## Root Cause

  The `.claude-plugin/plugin.json`:

  ```json
  {
    "name": "superpowers",
    "description": "Core skills library for Claude Code...",
    "version": "5.1.0",
    "author": { ... },
    "keywords": [ ... ]
  }
  ```

  But the `.cursor-plugin/plugin.json` correctly declares:

  ```json
  {
    "name": "superpowers",
    ...
    "skills": "./skills/",
    "agents": "./agents/",
    "commands": "./commands/",
    "hooks": "./hooks/hooks-cursor.json"
  }
  ```

  The Claude Code version is missing `"skills": "./skills/"`, so the harness never discovers the skill files.

  ## Fix

  Add `"skills": "./skills/"` to `.claude-plugin/plugin.json`.

  ## Environment

  - Superpowers: 5.1.0
  - Claude Code: 2.1.119
  - OS: macOS 15.4
  - Install method: official marketplace

  ## Related

  - #1465 reported identical symptoms ("skills missing from manifest") but was closed as not planned. This report provides the concrete root cause and fix.
  - #546 established that Claude Code rejects unrecognized keys in plugin.json — but `skills` should be a recognized key.

obra:
I don't believe this to be correct. Plenty of folks install from the plugin marketplace and *do* see the skills? Can you provide a trace from claude --debug?

arittr:
Closing for now since the requested `claude --debug` trace never arrived — marketplace installs do discover the skills without a `skills` field in plugin.json, so we need the trace to see what happened in your environment. If you still hit this on a current release, please reopen with the debug output. Thanks for the report.

PR #1568, opened 2026-05-17
fix(claude-plugin): declare bundled skills path

zichen0116:
## Summary
- add the missing `skills` entry to `.claude-plugin/plugin.json`
- point Claude Code at the bundled `./skills/` directory so marketplace installs can discover the shipped skills

## Why
Issue #1479 reports that the Claude Code manifest does not declare a skills path, which prevents bundled skills from being discovered even though they exist on disk.

## Verification
- parsed `.claude-plugin/plugin.json` successfully after the change
- confirmed the manifest now exposes `skills: ./skills/`
- ran `git diff --check`

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -16,5 +16,6 @@
     "collaboration",
     "best-practices",
     "workflows"
-  ]
+  ],
+  "skills": "./skills/"
 }


lligirlburg-wq:
Okay thanks so much for your email 📧

PR #1571, opened 2026-05-17
feat: add context-management skill with domain isolation

hanle1:
## What

Adds a new `context-management` skill for persistent context that survives across sessions.

### Features

- **Init Flow**: Auto-detects user identity (git config, cloud CLI, GitHub) and generates `context.md`
- **Add Flow**: Creates per-project context files with keyword-based auto-loading
- **Load Flow**: Loads global context every session, project context on keyword match
- **Maintenance**: Update, prune, archive flows to keep context clean
- **Domain Isolation** (advanced): Multi-context architecture for users with distinct roles (e.g., work + side projects + personal). Each domain gets its own sub-context file with session-locked isolation — no cross-domain leakage.

### Background

I have been using this context management system daily for several months across 3 isolated domains (work, innovation, personal) on Copilot CLI. The domain isolation pattern emerged from a real need — I wanted work context (on-call runbooks, team info) to never leak into personal sessions, and vice versa. The keyword-based auto-loading and session-locked domains have been stable and practical in daily use.

### Why this belongs in superpowers

Context management is one of the most common pain points with AI coding agents — every session starts from zero. This skill provides a structured, agent-agnostic approach that works across Claude Code, Copilot CLI, Cursor, and any tool that supports instruction files.

The domain isolation feature is unique — existing agent memory tools (AgentMemory, claude-memory-mcp, etc.) all use a single memory pool with no privacy-preserving domain separation.

### Files

- `skills/context-management/SKILL.md` — Full skill spec (init, add, load, update, prune, archive, domain isolation)
- `skills/context-management/templates.md` — Global and project context templates

### DIFF
--- a/skills/context-management/SKILL.md
+++ b/skills/context-management/SKILL.md
@@ -1,0 +1,273 @@
+---
+name: context-management
+description: Use when setting up persistent context for a new user, starting a new project that needs tracking, context files feel stale or bloated, or no context.md exists yet
+---
+
+# Context Management
+
+Persistent context that survives across sessions. Your AI agent remembers who you are, what you work on, and lessons you've learned.
+
+**Core principle:** One global file (identity + preferences) auto-loaded every session. Per-project files loaded only when relevant. Optional domain isolation for users with multiple roles.
+
+## When to Use
+
+```dot
+digraph when_to_use {
+    "No context.md exists?" [shape=diamond];
+    "New project needs tracking?" [shape=diamond];
+    "Context feels stale/bloated?" [shape=diamond];
+    "User says 'update/clean context'?" [shape=diamond];
+    "Run Init Flow" [shape=box];
+    "Run Add Flow" [shape=box];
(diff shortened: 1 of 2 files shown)


obra:
Closing — this doesn't seem like a fit for core Superpowers.

Thank you for the honest first-person writeup; the multi-domain isolation pattern is interesting. But the skill as designed is shaped around your specific multi-domain workflow ("work + side projects + personal", session-locked domain selection, keyword-based auto-loading of per-project context files). That's a configuration system, not a general-purpose skill that benefits all Superpowers users regardless of project.

> Skills for specific domains, specific tools, or specific workflows belong in their own standalone plugin. Ask yourself: "Would this be useful to someone working on a completely different kind of project than yours?" If not, publish it separately.

This would make a great standalone plugin (e.g. `context-management`) — you can publish it to the marketplace and users who want this exact workflow can install it alongside Superpowers. The plugin development docs explain how. Cross-session memory is a real problem; I just don't think the answer belongs in core.

PR #1575, opened 2026-05-18
docs(tdd): use backticked filename instead of @-prefix

### LINKED ISSUE #1529: test-driven-development: `@testing-anti-patterns.md` reference uses `@` prefix that doesnt resolve
In `skills/test-driven-development/SKILL.md`, line 359 references a file with an `@` prefix:

```
When adding mocks or test utilities, read @testing-anti-patterns.md to avoid common pitfalls:
```

The actual file lives at `skills/test-driven-development/testing-anti-patterns.md` (no `@` prefix; the URL `https://raw.githubusercontent.com/obra/superpowers/main/skills/test-driven-development/testing-anti-patterns.md` returns HTTP 200).

The `@` prefix is the **only such reference style anywhere in this SKILL.md** — every other internal file reference is written as a bare filename or a code-fenced path. There is no autolink rule in standard markdown that would resolve `@testing-anti-patterns.md` to the file, and there is no `@`-handling convention referenced in the skill's preamble. An agent reading this passage will either treat `@testing-anti-patterns.md` as literal text (and fail to read the file) or guess that it should resolve to `testing-anti-patterns.md` (correct answer, but only by inference).

### Suggested fix

Replace the `@`-prefixed reference with a plain (or backtick-wrapped) filename so the path is unambiguous:

```diff
- When adding mocks or test utilities, read @testing-anti-patterns.md to avoid common pitfalls:
+ When adding mocks or test utilities, read `testing-anti-patterns.md` to avoid common pitfalls:
```

Or, if the `@` prefix is meant as a convention that should be documented, add a one-line note near the top of the skill explaining that `@<filename>` means "the file at `<filename>` in this skill's folder" — and use it consistently across all internal references.

gaurav0107:
Fixes #1529.

Line 359 of `skills/test-driven-development/SKILL.md` references `@testing-anti-patterns.md`. Nothing in this repo strips the `@` — there's no preamble convention for it, no skill loader that resolves it, and no other `@`-prefixed file reference in this SKILL.md. Every other internal file pointer here is either a bare filename or a backticked path, so the `@` reads as a mistake.

The actual file lives next to the SKILL at `skills/test-driven-development/testing-anti-patterns.md`.

This PR replaces the `@`-prefixed token with a backticked filename, matching the citation style used everywhere else in the same file:

```diff
- When adding mocks or test utilities, read @testing-anti-patterns.md to avoid common pitfalls:
+ When adding mocks or test utilities, read `testing-anti-patterns.md` to avoid common pitfalls:
```

One line, one file. No behavior change — the surrounding section about mocks and test utilities is unchanged.

I checked the rest of the repo for the same pattern with `grep -nE '@[a-zA-Z0-9_-]+\.md'` and found exactly one other instance: `skills/writing-skills/SKILL.md` line 556 (`See @testing-skills-with-subagents.md`). I left that one alone — issue #1529 only names the TDD file, and `writing-skills/SKILL.md` is the kind of carefully-tuned content that probably wants its own discussion. Happy to file a follow-up if you want it fixed too.

### DIFF
--- a/skills/test-driven-development/SKILL.md
+++ b/skills/test-driven-development/SKILL.md
@@ -356,7 +356,7 @@ Never fix bugs without a test.
 
 ## Testing Anti-Patterns
 
-When adding mocks or test utilities, read @testing-anti-patterns.md to avoid common pitfalls:
+When adding mocks or test utilities, read `testing-anti-patterns.md` to avoid common pitfalls:
 - Testing mock behavior instead of real behavior
 - Adding test-only methods to production classes
 - Mocking without understanding dependencies


obra:
Thanks! This is fixed on \`dev\` already in 3d6dc90 (PR #1532, which landed shortly before yours). That fix went one step further — it uses a real markdown link \`[testing-anti-patterns.md](testing-anti-patterns.md)\` instead of just backticks, so the file is actually navigable from the rendered docs.

Your PR targeted \`main\`, which hasn't received the next release yet, so the bug still appears valid there. Closing as already-fixed-on-dev.

Issue #1576, opened 2026-05-18
writing-plans doesn't enforce TDD, causing entire Subagent-Driven Development process to skip TDD


liumeng666:
**Before submitting:**
- [ ] I searched existing issues and this is not a duplicate

## Environment

| Field | Value |
|-------|-------|
| Superpowers version | 5.0.7 |
| Harness | Claude Code |
| Harness version | Latest (Claude 4.5/4.6) |
| Model |  claude-sonnet-4-6 |
| OS + shell | Windows 10 / bash |

## Is this a Superpowers issue or a platform issue?

- [ ] I confirmed this issue does not occur without Superpowers installed
- **Note:** This is specific to Superpowers (writing-plans + subagent-driven-development skills)

## What happened?

The **superpowers:writing-plans** skill does not enforce or encourage Test-Driven Development (TDD) requirements in the generated plan templates. This creates a disconnect when the plan is later executed using **superpowers:subagent-driven-development**, which expects TDD to be followed.

**The Problem Flow:**
1. User uses `superpowers:writing-plans` to create an implementation plan
2. Plan is generated without TDD requirements
3. User uses `superpowers:subagent-driven-development` to execute the plan
4. Subagent checks `implementer-prompt.md` which says: "Write tests (following TDD **if task says to**)"
5. Since the plan doesn't require TDD, subagent skips TDD entirely
6. Result: Code is written without tests, violating quality standards

**User Experience:**
- I tried using Subagent-Driven Development multiple times
- None of the attempts followed TDD principles
- Initially thought I was using the skill incorrectly
- After analysis, discovered the root cause: plan templates don't require TDD

## Steps to reproduce

1. Invoke `/superpowers:writing-plans` to create an implementation plan
2. Generate a plan for any feature (e.g., "build a case management system")
3. Observe that the generated plan does **not** include:
   - TDD requirements
   - Test writing steps
   - Test coverage requirements
4. Invoke `/superpowers:subagent-driven-development` with the generated plan
5. Observe that implementer subagents skip TDD because "the task doesn't say to"
6. Result: No tests are written during implementation

## Expected behavior

**Option A:** The `superpowers:writing-plans` skill should automatically include TDD requirements in generated plans, such as:

```markdown
## Development Principles

**Mandatory Requirements:**
- All development tasks MUST follow Test-Driven Development (TDD)
- Red-Green-Refactor cycle: Write failing test → Write implementation → Refactor
- Each feature must have corresponding unit tests
[...]

PR #1555, opened 2026-05-15
fix(brainstorming): cap WebSocket frame payloads

therahul-yo:
## What problem are you trying to solve?

Issue #1446 reports that `skills/brainstorming/scripts/server.cjs` trusts the 64-bit WebSocket payload length from a client frame and then allocates `Buffer.alloc(payloadLen)`. Because the brainstorm companion server listens on localhost, a local process or browser context can advertise a huge frame length and make the server attempt a very large allocation.

This PR addresses only that concrete payload-size allocation bug from #1446. It does not attempt to fix the other two unrelated items in the bundled issue.

## What does this PR change?

Adds a 10 MiB maximum WebSocket frame payload size, rejects oversized advertised payloads before allocation, and exports the limit for protocol tests. Adds a regression test that constructs an oversized 64-bit frame header and verifies `decodeFrame()` rejects it from the header alone.

## Is this change appropriate for the core library?

Yes. The brainstorm companion server is part of the core `brainstorming` skill, and this hardens its built-in zero-dependency WebSocket parser for all users. It is not project-specific, team-specific, or tied to a third-party integration.

## What alternatives did you consider?

I considered limiting only complete frames after `buffer.length >= totalLen`, but that still leaves the parser willing to wait for an enormous declared payload and keeps the bad length in control flow. Rejecting immediately after reading the extended length avoids the risky allocation path and makes the behavior explicit.

I also considered changing the server to use a WebSocket dependency, but Superpowers is intentionally dependency-light and this bug can be fixed locally in the existing parser.

## Does this PR contain multiple unrelated changes?

No. This PR only caps inbound brainstorm WebSocket frame payloads and adds protocol-level coverage for that limit.

## Existing PRs
- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: #1504, #1483

I searched open and closed PRs for `1446`, `escape_for_json`, `payload-size`, and `find-polluter`. I found #1504, which hardens a different brainstorm-server path (`handleMessage` null payload handling), and #1483, which addressed the separate `find-polluter.sh` spaced-path issue from #1446 and was closed. I did not find an open or merged PR that caps WebSocket frame payload length before allocation.

## Environment tested

| Harness (e.g. Claude Code, Cursor) | Harness version | Model | Model version/ID |
|-------------------------------------|-----------------|-------|------------------|
| Codex local shell | Codex desktop | GPT-5 | Codex session |

[...]

Issue #1446, opened 2026-04-30
v5.0.7: three latent bugs found in static review (escape_for_json control-char gap, server.cjs payload-size DoS, find-polluter.sh unquoted loop)


waple0820:
Hi — while vendoring v5.0.7 into a downstream repo, an automated review picked up three latent bugs across hooks/scripts. Reporting in one bundle since they're independent and individually small. None are critical for our use case (we won't patch the vendor copy and rely on upstream fixes), but they look like real correctness gaps.

---

### 1. `hooks/session-start` — `escape_for_json` only escapes a subset of C0 control characters

`escape_for_json` (lines ~23–31) handles `\\`, `"`, `\n`, `\r`, `\t` but does not escape `\b` (U+0008), `\f` (U+000C), or other C0 control characters (U+0000–U+001F) which RFC 8259 requires be escaped in JSON strings.

If any `SKILL.md` file (now or in the future) contains an unescaped form-feed / backspace / etc., the hook emits invalid JSON and Claude Code silently discards it — superpowers never activates. Failure mode is silent.

Related: closed issue #612 covered a different aspect of the same function.

**Suggested fix:** add the remaining required substitutions, or pipe through `python -c 'import json,sys; print(json.dumps(sys.stdin.read())[1:-1])'` to delegate escaping to a real JSON encoder.

---

### 2. `skills/brainstorming/scripts/server.cjs` — no upper bound on WebSocket payload allocation

```js
payloadLen = Number(buffer.readBigUInt64BE(2));
// ...
const data = Buffer.alloc(payloadLen);
```

A WebSocket client can advertise a 64-bit payload length, causing `Buffer.alloc` to attempt a multi-GB allocation and crash the brainstorm server process. Risk is bounded — server binds to `127.0.0.1` so exploitation requires local access (any process / browser tab on the user's machine) — but worth a sanity ceiling (e.g. reject frames > 10 MB).

Related: #1014 covers Origin validation on the same WebSocket path; this is a separate hardening item.

**Suggested fix:** validate `payloadLen <= MAX_FRAME_BYTES` and close the connection on overrun.

---

### 3. `skills/systematic-debugging/find-polluter.sh` — unquoted loop variable word-splits paths with spaces

```bash
TEST_FILES=$(find . -path "$TEST_PATTERN" | sort)
for TEST_FILE in $TEST_FILES; do
```

[...]

PR #1589, opened 2026-05-21
fix: guard EnterWorktree core.bare parent checkout

YOMXXX:
## What problem are you trying to solve?

`skills/using-git-worktrees/SKILL.md` Step 1a tells agents to prefer native worktree tools such as Claude Code's `EnterWorktree`. That native-tool preference is still the right architecture, but issue #1546 documents an upstream Claude Code failure mode: `EnterWorktree` can write `core.bare=true` into the parent checkout's shared git config.

When that happens, the parent checkout looks like it lost its working tree. Git commands that require a work tree can fail with `fatal: this operation must be run in a work tree`, while the files themselves appear untouched. Because Superpowers actively steers Claude Code agents toward `EnterWorktree`, the skill should also tell agents how to detect and repair this parent-checkout symptom.

## What does this PR change?

Adds a narrow Claude Code known-issue guard to `using-git-worktrees` Step 1a:

- record `PARENT_REPO_ROOT=$(git rev-parse --show-toplevel)` before native worktree creation
- after the native tool returns, check the parent checkout's `core.bare`
- if it flipped to `true`, run `git -C "$PARENT_REPO_ROOT" config --unset core.bare`

This PR also adds a fast static regression test, plus a short design spec and implementation plan for the change.

## Is this change appropriate for the core library?

Yes. `using-git-worktrees` is a core Superpowers skill, and this change protects a core workflow that the skill itself recommends. The guard is tool-specific to Claude Code, but it lives inside the existing native-tool branch rather than adding a new harness, dependency, or project-specific behavior.

## What alternatives did you consider?

1. Disable or discourage `EnterWorktree`. Rejected because #1121 established and validated the native-tool-first behavior; the upstream bug does not make manual `git worktree add` the right default.
2. Add only a prose warning. Rejected because #1546's symptom is recoverable, and agents need the exact `git -C "$PARENT_REPO_ROOT" config --unset core.bare` command.
3. Add a runtime hook. Rejected as too broad for this PR; a few Step 1a instructions solve the immediate parent-checkout recovery problem without new runtime surface area.

## Does this PR contain multiple unrelated changes?

No. It only addresses #1546: detecting and repairing the known Claude Code `EnterWorktree` parent-checkout `core.bare` symptom while preserving native worktree preference.

## Existing PRs
- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: #1121, #1167

[...]

PR #1590, opened 2026-05-21
fix: escape SessionStart JSON control characters

YOMXXX:

Issue #1446 reports three independent latent bugs. This PR fixes only the first one: `hooks/session-start` has a hand-written JSON string escaper that handles backslash, double quote, newline, carriage return, and tab, but leaves other C0 control characters raw.

Raw U+0000 through U+001F characters are not valid inside JSON strings. If `skills/using-superpowers/SKILL.md` ever contains a raw backspace, form feed, vertical tab, unit separator, or similar control character, the SessionStart hook emits invalid JSON and the harness cannot parse the payload. The likely user-facing failure is Superpowers not activating even though the plugin is installed.

`hooks/session-start-codex` carries the same escape helper, so it has the same failure mode.

## What does this PR change?

Escapes every C0 control character that Bash strings can carry in both SessionStart hooks:

- keeps short escapes for `\b`, `\f`, `\n`, `\r`, and `\t`
- uses `\u0001` through `\u001f` for the remaining representable controls
- leaves hook output shape unchanged

It also extends the SessionStart hook tests with temporary plugin fixtures whose `using-superpowers` skill content contains U+0001 through U+001F, then parses the real hook output with `JSON.parse`.

Yes. SessionStart bootstrap output is core plugin infrastructure used before any skill can run. The fix is general-purpose, dependency-free, and applies to both the standard SessionStart hook and the Codex-specific SessionStart hook.

1. Use Python/Node/jq as a real JSON encoder. Rejected because the hook is intentionally shell-based and dependency-light, and adding runtime dependencies to SessionStart increases bootstrap fragility.
2. Fix only `hooks/session-start`. Rejected because `hooks/session-start-codex` duplicates the same escape helper and would keep the latent bug.
3. Address all three #1446 reports in one PR. Rejected because they are independent. The WebSocket payload-size issue already has open PR #1555, and the `find-polluter.sh` path-splitting issue should be a separate narrow PR if pursued.

No. It only fixes SessionStart JSON escaping for control characters and adds a regression test/spec/plan for that one bug.

## Existing PRs
- Related PRs: #134, #1555

#134 is prior art for Windows/plugin hook support. #1555 covers a different #1446 sub-issue: WebSocket payload-size caps in the brainstorm server. Searches for `1446 escape_for_json`, `escape_for_json control char`, and related terms did not find a duplicate PR for this SessionStart JSON escaping fix.

## Environment tested

| Shell hook tests | macOS zsh/bash | n/a | n/a |
| OpenCode plugin tests | local test harness | n/a | n/a |
| Codex plugin sync tests | local test harness | n/a | n/a |

Not applicable. This PR does not add support for a new harness.

<details>

```
N/A - no new harness support.
```

</details>

## Evaluation

- Initial prompt: human partner asked to continue implementing valuable fixes from the earlier triage. #1576 and #1014 were skipped because existing open PRs cover them; this PR selected the unclaimed SessionStart JSON escaping sub-issue from #1446.
- RED phase: after adding the control-character fixture tests, `bash tests/hooks/test-session-start.sh` failed for both hooks with `invalid JSON: Bad control character in string literal`.
- GREEN phase: after escaping the rest of the C0 range, the same test passed, including both `SessionStart escapes JSON control characters in skill content` and `Codex SessionStart escapes JSON control characters in skill content`.

Verification commands run fresh before PR creation:

```bash
bash tests/hooks/test-session-start.sh
bash tests/opencode/run-tests.sh
bash tests/codex-plugin-sync/test-sync-to-codex-plugin.sh
bash -n hooks/session-start hooks/session-start-codex tests/hooks/test-session-start.sh
git diff --check HEAD~1 HEAD
```

All exited 0.

## Rigor

This is hook infrastructure, not a skill behavior change. The adversarial test uses a real hook output path with all representable C0 controls in the injected skill content and verifies JSON parsing plus decoded context preservation.

## Human review

Human partner reviewed the fork compare diff for `YOMXXX:fix/session-start-json-control-escaping` and explicitly approved opening the PR. They also gave standing approval to open future PRs after the complete diff and verification summary are shown.


YOMXXX:
### Reviewer note

This PR is one of six small, independent PRs opened from the same triage pass. They can be reviewed and merged independently:

- #1588 worktree write boundary
- #1589 EnterWorktree `core.bare` guard
- #1590 SessionStart JSON control escaping
- #1591 visual companion skill-relative script paths
- #1592 windows lifecycle test drift
- #1593 README Output Paths override example

**Focus for this PR:** harden SessionStart JSON output so all representable C0 control characters in injected skill content are escaped before the harness parses the hook payload.

**Review surface:** `hooks/session-start`, `hooks/session-start-codex`, and hook tests. It keeps the hook output shape unchanged and does not add runtime dependencies.

**Verification:** hook tests use temporary skill fixtures containing U+0001 through U+001F and parse the real hook output with `JSON.parse`; OpenCode tests and Codex plugin sync tests also passed.

**Risk to check:** shell escaping correctness and whether both SessionStart variants stay in sync.

obra:
Closing.

The escape_for_json function does have the gap you describe — it handles `\n \r \t \\ \"` but not other C0 control characters. However, `skills/using-superpowers/SKILL.md` doesn't contain raw C0 characters in practice, so this is a hypothetical failure mode rather than an observed one. The PR is also 101 lines across 3 files including a fixture-based test for the hypothetical case.

This kind of defensive hardening could be worth a much smaller, targeted PR if someone hits an actual repro. As filed, it's part of a 17-PR batch from the same author and the cost (modifying both SessionStart hooks plus a fixture test) is disproportionate to a non-observed bug.

#1446 has three sub-issues; this one is the lowest-priority of the three. If the failure ever surfaces in a real session, please open a focused single-file PR.

PR #1601, opened 2026-05-21
@mhat reported that his claude got confused about 'debugging' being n…

obra:
@mhat's claude reported:

```
Honest answer, and I'll separate the two because they have different causes — one I can trace in context, one I mostly can't.  superpowers:debugging — traceable. The SessionStart hook injected the full text of the superpowers:using-superpowers skill. Two things in that text fed the guess:

  1. The hook introduced it as "your superpowers:using-superpowers skill" — so I had a confirmed example of the superpowers: namespace prefix.
  2. The skill body names "debugging" as a skill, twice:
  ▎ "Process skills first (brainstorming, debugging)"
  ▎ "'Fix this bug' → debugging first, then domain-specific skills."
```

### DIFF
--- a/skills/using-superpowers/SKILL.md
+++ b/skills/using-superpowers/SKILL.md
@@ -102,15 +102,15 @@ These thoughts mean STOP—you're rationalizing:
 
 When multiple skills could apply, use this order:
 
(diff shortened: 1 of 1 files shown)


arittr:
LGTM

PR #1486, opened 2026-05-06
Cross-platform skill compatibility: agent-neutral prose, source-verified per-runtime tool refs

obra:
## What problem are you trying to solve?

Superpowers ships to Claude Code, Codex (CLI + App), Copilot CLI, Cursor, Factory Droid, Gemini CLI, and OpenCode, but the skill content was written first for Claude Code and bakes in Claude-Code-specific names, paths, and tool vocabulary. The motivating concrete failure was [openai/plugins#217](https://github.com/openai/plugins/pull/217) — OpenAI's vendored fork of Superpowers attempted a wholesale Claude→Codex find-and-replace and got most of it wrong (rewrote historical attribution paths, replaced model names, broke install instructions). Their PR landed because nobody upstream had separated which references were *actually Claude-Code-specific* from which were *generic prose that just happened to say "Claude"*.

Symptoms users hit on non–Claude-Code runtimes:
- Skills like `subagent-driven-development` ship prompt templates with a `Task tool (general-purpose):` header that doesn't match what Codex / Copilot CLI / Gemini CLI expose. Users either ignore the template or attempt to call a tool that doesn't exist on their runtime.
- `using-superpowers/SKILL.md` told users to "see `references/copilot-tools.md` (Copilot CLI), `references/codex-tools.md` (Codex)" but the priority list of instruction files was Claude-Code-first (`CLAUDE.md, GEMINI.md, AGENTS.md`), and the per-runtime instruction files weren't documented for each platform.
- Several refs contained **factually wrong** tool names that nobody had verified against source. For example, `gemini-tools.md` mapped subagent dispatch to `@agent-name`. The actual Gemini CLI tool is `invoke_agent` (per `AGENT_TOOL_NAME = 'invoke_agent'` in `tool-names.ts`); `@generalist` is chat sugar that triggers `invoke_agent`. Same class of error in `copilot-tools.md` (claimed `create`/`edit` tools that don't exist; Copilot uses `apply_patch` and `rg`, not `create`/`edit` and `grep`) and in `OpenCode INSTALL.md` (claimed `Task → @mention syntax`; OpenCode's real subagent dispatch tool is `task` with a `subagent_type` parameter, no @mention syntax).

The work happened in five phases driven by clarifying questions through `superpowers:brainstorming`. Each phase scope was approved before implementation, then validated against authoritative source before the next phase.

## What does this PR change?

Five logically-distinct changes, one commit each:
[...]

PR #1122, opened 2026-04-10
fix: replace hardcoded /Users/jesse with generic placeholders (#858)

arittr:
## What problem are you trying to solve?

Several files contained hardcoded `/Users/jesse` paths (#858). These paths break for any user on a different system and leak personal directory structure into the codebase.

## What does this PR change?

Replaces all hardcoded `/Users/jesse` paths with generic placeholders (e.g. `~/project`, `<project-root>`) in `docs/testing.md`, `skills/systematic-debugging/CREATION-LOG.md`, and `skills/systematic-debugging/root-cause-tracing.md`.

## Is this change appropriate for the core library?

Yes — hardcoded user paths in shared docs/skills are a bug that affects all users.

## What alternatives did you consider?

1. Using `$HOME` environment variable references — less readable in documentation context.
2. Only fixing new occurrences going forward — leaves existing broken examples for users who copy-paste.

## Does this PR contain multiple unrelated changes?

No. All changes address the same issue: replacing hardcoded `/Users/jesse` paths with generic placeholders.

## Existing PRs
- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: #858 (the issue this addresses)

## Environment tested

| Harness (e.g. Claude Code, Cursor) | Harness version | Model | Model version/ID |
|-------------------------------------|-----------------|-------|------------------|
| Claude Code | latest | Claude Opus 4.6 | claude-opus-4-6 |

## Evaluation
- Grep confirmed no remaining `/Users/jesse` references after the change
- Documentation examples render correctly with generic placeholders

## Rigor

- [x] If this is a skills change: I used `superpowers:writing-skills` and
      completed adversarial pressure testing (paste results below)
- [x] This change was tested adversarially, not just on the happy path
- [x] I did not modify carefully-tuned content (Red Flags table,
      rationalizations, "human partner" language) without extensive evals
      showing the change is an improvement

## Human review
- [x] A human has reviewed the COMPLETE proposed diff before submission

PR #1603, opened 2026-05-22
feat(tests): cross-harness portability lint

mvanhorn:
## Motivation

PR #1486 made cross-harness purity a maintained invariant of the skills library. This PR adds the static mechanism to keep it from regressing. The annotations to existing skill files make `main` start the lint clean so CI can be strict from day one.

PR #1486 (`Cross-platform skill compatibility: agent-neutral prose, source-verified per-runtime tool refs`, merged 2026-05-14) established cross-harness purity as a maintained invariant of the skills library. From your own PR body:

> The motivating concrete failure was [openai/plugins#217](https://github.com/openai/plugins/pull/217) - OpenAI's vendored fork of Superpowers attempted a wholesale Claude->Codex find-and-replace and got most of it wrong (rewrote historical attribution paths, replaced model names, broke install instructions). Their PR landed because nobody upstream had separated which references were actually Claude-Code-specific from which were generic prose that just happened to say "Claude".

PR #1486 cleaned up the existing state. The invariant is currently maintained by review attention alone. Nothing structural prevents the same regression class from re-entering on the next PR or downstream vendor fork. The session that motivated this work was reading PR #1486 and running grep against current main to find latent runtime-specific tokens the invariant doesn't actively prevent: `skills/dispatching-parallel-agents/SKILL.md:69` contained `// In Claude Code / AI environment` in a generic TypeScript example block (not inside a per-runtime section), and that's the kind of regression a static lint catches in seconds.

## What does this PR change?

Adds `tests/lint-cross-harness.sh`, a 164-line POSIX shell + awk static lint that scans `skills/**/*.md` for four classes of runtime-specific tokens. CI workflow at `.github/workflows/lint.yml` runs the lint on pull_request and push to main/dev and blocks merge on any violation. Operator documentation at `docs/cross-harness-lint.md`.

To make main pass the new lint from day one, the PR also adds 34 inline allowlist annotations across 15 existing skill files (using-superpowers, subagent-driven-development, writing-skills, executing-plans, requesting-code-review, brainstorming, writing-plans, dispatching-parallel-agents). The annotations are HTML comments of the form `<!-- lint-cross-harness: allow "TOKEN" reason="..." -->`. They do not change rendered Markdown content; they document existing legitimate per-runtime references so the lint can distinguish those from future regressions.

Yes. The lint maintains the cross-harness purity invariant the core library committed to in PR #1486. It is infrastructure tooling (tests + CI + docs), not a new skill, not a third-party integration, not a domain-specific workflow. The annotations to existing skill files do not modify behavior-shaping content (Red Flags tables, rationalization lists, "human partner" language). They are HTML metadata read by the lint and invisible in the rendered Markdown.

This is the static-analysis counterpart to the cross-agent scenario testing harness being built internally. Both layers catch different failure modes at different cost: static lint catches token-level violations in seconds at PR time without spinning up any harness; runtime testing catches behavioral divergence across agents after merge. They are complementary.

1. **Runtime check inside `lib/`**: rejected. Invasive, requires per-harness wiring, duplicates work the runtime scenario harness will already do.
2. **Python or Node-based linter**: rejected. Adds a dependency to a Shell + Markdown project. POSIX shell + awk runs on any GitHub Actions worker with no setup.
3. **Off-the-shelf tools (markdown-link-check, vale, etc.)**: rejected. None understand cross-harness portability semantics; would need custom rules anyway.
4. **Make CI advisory (continue-on-error: true) on first introduction**: considered. Rejected in favor of annotating the 34 existing references in this PR so CI can be strict from day one. Half-baked CI is worse than no CI.
5. **Add allowlist annotations as a follow-up PR after this one lands**: rejected. The annotations make sense only in context of the lint existing. Splitting them creates a window where main is broken under the new strict workflow.

No. All changes serve one purpose: introduce the cross-harness portability lint with strict CI from day one. The lint script is the mechanism, the CI workflow is the enforcement, the docs explain it, and the 34 inline annotations are needed to make existing main pass the lint (without which strict CI would block the next PR you merge). The annotations are HTML metadata consumed by the lint - they do not alter rendered Markdown, do not modify behavior-shaping skill prose, and each carries a reason= field that documents why the surrounding paragraph legitimately mentions a runtime-specific token.

## Existing PRs

- Related PRs:
  - #1486 (your cross-platform compatibility PR; this lint extends its invariant)
  - #1122 (replaced hardcoded `/Users/jesse` with generic placeholders; this lint includes a `hardcoded-path` class that prevents the same class of regression)
  - #1568 (closed: proposed adding a `skills` field to `.claude-plugin/plugin.json`; THIS PR makes no plugin-manifest changes)

I did not find a prior PR proposing a static cross-harness portability lint.

## Environment tested

|------------------------------------|-----------------|-------|------------------|
| GitHub Actions (ubuntu-latest runner) | n/a | n/a | n/a |

The lint is a static `sh` + `awk` script with no agent runtime dependency. The Environment table reflects where the workflow executes; tools required at runtime are `find`, `awk`, and `bash`, all standard on ubuntu-latest. I verified the lint locally on macOS 15 with `/bin/sh` (POSIX) and confirmed `sh -n` is clean.

## New harness support

N/A. This PR adds no new harness.

## Evaluation

- **Initial prompt**: reading PR #1486 + openai/plugins#217, then `grep -rn 'In Claude Code' skills/` to find what the invariant didn't structurally prevent. Found `skills/dispatching-parallel-agents/SKILL.md:69` as the canary case.
- **Iterations on the lint script**: two passes. First pass exited 1 with 51 violations on main. Review surfaced that the workflow would block CI immediately, so the lint was narrowed to recognize the existing per-runtime conventions in this codebase. Second pass added bold-prose inline section markers (`**In <harness>:**`) and extended the `references/<runtime>-tools.md` file allowance to bare-harness names. Result: 34 violations remaining, all genuine cross-harness leakage in skill prose.
- **Annotation pass**: added 34 inline allowlist comments with reason= fields documenting each legitimate reference (frontmatter discovery text, runtime mapping prose, dispatch prompt headers, quoted Anthropic docs, graphviz workflow labels). After the annotation pass, the lint reports `Lint complete: 0 violations.` on this branch.

## Rigor

- [ ] If this is a skills change: this PR adds HTML allowlist comments to skill files but does not modify behavior-shaping content. The comments render invisibly in GitHub. No `superpowers:writing-skills` adversarial testing was performed because no behavior-shaping content was changed.
- [x] This change was tested adversarially: lint output progressively narrowed from 51 -> 34 -> 0 violations through iterative tightening of the false-positive surface (bold-prose inline markers, references/-tools.md allowance) and per-paragraph annotation of genuine cross-harness leakage.
- [x] I did not modify carefully-tuned content (Red Flags table, rationalization lists, "human partner" language).

## Human review

---

### Lint output on this branch

```
$ bash tests/lint-cross-harness.sh
Lint complete: 0 violations.
$ echo $?
0
```

### Lint output on the canary case (before annotation)

```
$ bash tests/lint-cross-harness.sh
VIOLATION [bare-harness-name] skills/dispatching-parallel-agents/SKILL.md:69
    // In Claude Code / AI environment
... (33 more violations across 14 other skill files)
Lint complete: 34 violations in 15 files.
$ echo $?
1
```

AI was used for assistance.

Fixes the cross-harness portability invariant maintenance gap left by PR #1486.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,15 @@
+name: lint
+
+on:
+  pull_request:
+    branches: [main, dev]
+  push:
+    branches: [main, dev]
+
+jobs:
+  cross-harness-lint:
+    name: cross-harness-lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: bash tests/lint-cross-harness.sh
--- a/docs/cross-harness-lint.md
+++ b/docs/cross-harness-lint.md
@@ -1,0 +1,49 @@
+# Cross-Harness Lint
+
+`tests/lint-cross-harness.sh` checks `skills/**/*.md` for runtime-specific language that would weaken the cross-harness purity invariant introduced by PR #1486. It is a static Markdown lint only: it does not change skill content, rewrite files, or validate runtime behavior.
+
+## Banned Patterns
+
+- Bare harness names in generic prose: `Claude Code`, `Cursor`, `OpenCode`, `Codex CLI`, `Codex App`, `Gemini CLI`, `GitHub Copilot CLI`, and `Factory Droid` are violations outside runtime-specific sections. Example violation: `Use Claude Code for this step.` Allowed section example: `## In Claude Code`.
+- Model identifiers: concrete model names such as `claude-opus-4-7`, `gpt-5.5`, `gemini-2.5-pro`, and `o4-mini` are always violations unless covered by a local allowlist.
+- Runtime-specific tool names: names such as `ExitPlanMode`, `TodoWrite`, `WebFetch`, `Task tool`, `Skill tool`, and `mcp__server__tool` are violations outside runtime-specific sections or `references/<runtime>-tools.md`. Allowed section example: `### For Cursor`.
+- Hardcoded user paths: macOS personal home paths such as `/Users/jesse/` are always violations unless covered by a local allowlist.
+
+Lines are allowed under any of these rules. The lint stops at the first rule that applies; tokens are not re-checked after a line passes one rule.
+
+- **Two-or-more agents on the same line.** Any line that names two or more distinct harness families (Claude, Codex, Cursor, OpenCode, Gemini, Copilot, Factory Droid, Aider, Cline, Windsurf, Hermes, Hyperagent, Antigravity, Kiro, Qwen, Kimi) is intentional cross-runtime prose. Example that passes: `~/.claude/skills for Claude Code, ~/.agents/skills/ for Codex`.
+- **A per-runtime section header.** A Markdown heading whose text matches `In <harness>`, `For <harness>`, or `<harness>:` opens a section that continues until the next heading of equal or lesser depth. Tokens inside the section do not flag.
+- **An inline bold-prose runtime marker** at the start of a line: `**In <harness>:** ...`, `**For <harness>:** ...`, or `**<harness>:** ...`. The marker covers that line only. This is the pattern used in `skills/using-superpowers/SKILL.md`.
+- **A `skills/*/references/<runtime>-tools.md` file.** Bare-harness and runtime-tool checks are suppressed for these files. Model identifiers and hardcoded `/Users/<name>/` paths are still flagged regardless of file or section context.
+- **The internal exception list** at `tests/lint-cross-harness.exceptions`. One `path:line[:reason]` per line, `#` for comment lines. Use this when a single-agent reference is intentional but the two-agents rule does not naturally apply (graphviz nodes, dispatch prompt headers, source-attributed excerpts).
+
+## Internal Exception List
+
+`tests/lint-cross-harness.exceptions` carries the lint's own list of allowed single-agent references so skill content stays free of inline annotations. Format:
+
+```
+# Comments start with hash
+skills/path/to/file.md:LINE:reason
+
+skills/another/file.md:42:reason describing why this single-agent reference is intentional
+```
+
+The reason field is free-form text after the second colon and is for humans reading the file. The lint does not validate or enforce it.
+
+## Running Locally
+
+Run the lint from the repository root:
+
+```sh
+bash tests/lint-cross-harness.sh
+```
+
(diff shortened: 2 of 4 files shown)


obra:
@mvanhorn - The linter comments aren't gonna work. I wonder if the right thing is to teach the linter script to maintain its own list of exceptions internally. They pollute everybody's context windows.  The thing that we want to avoid is casual or unintentional places where we are assuming a single agent's worldview. Anytime we mention at least two agents, it's almost certainly okay.

mvanhorn:
Done - pushed the rewrite. The lint now auto-allows any line that names 2+ harness families (Claude / Codex / Cursor / OpenCode / Gemini / Copilot / Aider / Cline / Windsurf / Hermes / Hyperagent / Antigravity / Kiro / Qwen / Kimi / Factory Droid), which picks up cross-runtime prose like `~/.claude/skills for Claude Code, ~/.agents/skills/ for Codex` without annotation. For the remaining single-agent references (graphviz workflow nodes, subagent dispatch prompt headers, source-attributed Anthropic excerpts, the TodoWrite examples in persuasion-principles), there is a sidecar `tests/lint-cross-harness.exceptions` file with `path:line:reason` entries. Zero skill files modified.

obra:
Hm. making the exception table hardcoded with line-numbers means that any skill edit is going to me a maint nightmare.

mvanhorn:
Agreed, line numbers are the wrong key, any skill edit silently shifts them.

Most of the 34 entries aren't really prose exceptions: 12 are graphviz node labels inside ```dot fences, the `Task()` example is fenced too, and 2 are the verbatim Anthropic doc excerpt. So I'd make the linter structurally aware instead of maintaining a list:

- Skip fenced code blocks for the prose token classes (model-id and hardcoded-path stay banned everywhere).
- Keep the 2+-agent auto-allow.
- For the genuine single-agent prose that's left (e.g. "Use tracking: TodoWrite for checklists"), key the sidecar on the matched line's text rather than its line number, so reordering a skill never rots it and editing the actual sentence re-triggers review.

That takes the list from 34 to a handful, and nothing breaks on edits. Good to push that direction, or would you rather I drop the prose token classes entirely and aim for zero list?

PR #1611, opened 2026-05-22
docs(dispatching-parallel-agents): rewrite for current Agent tool semantics

dotcomjack:

The current skill teaches Claude Code agent dispatch using:

1. **`Task(...)` as the tool name** — obsolete; the actual tool is `Agent`.
2. **No explicit statement** of the one-message-multiple-tool_use-blocks rule (this is also what #1470 addresses, additively).
3. **No `subagent_type` guidance** — Claude Code now has specialized subagent types (`Explore`, `general-purpose`, `Plan`, `feature-dev:code-architect`, `feature-dev:code-reviewer`, `feature-dev:code-explorer`, `posthog:error-analyzer`, etc.), and picking the right one matters more than prompt wording in many cases.
4. **No coverage** of modern Agent options: `isolation: \"worktree\"`, `run_in_background`, `model` override, `name` + `SendMessage` continuation.
5. **No \"never delegate understanding\" principle**, which is in the underlying Agent tool description but not surfaced as a teaching.
6. **Verification frames the summary as authoritative**, when an agent's summary describes what it *intended* to do, not what it *did*.

### Empirical baseline (RED)

I ran a baseline probe before writing the rewrite. Prompt to a fresh subagent (no skill loaded): *\"Imagine you are the main Claude Code assistant. The user says 'three test files are failing for independent reasons: src/abort.test.ts, src/batch.test.ts, src/race.test.ts. Dispatch parallel agents to fix them.' Sketch the exact tool calls you'd make.\"*

The subagent:
- Used the **wrong tool name** (`TaskCreate`).
- Treated worktree isolation as a **separate tool step** (\"I'd spin up `EnterWorktree` per agent first\") rather than an Agent parameter.
- Got **foreground/background semantics backwards** (\"no — `TaskCreate` is already async/non-blocking from my POV; I await all three results in the next turn\").

So even a fresh model with no skill loaded reaches for the obsolete pattern.

## What does this PR change?

Rewrites `skills/dispatching-parallel-agents/SKILL.md`. 83 insertions, 141 deletions; word count drops from 950 to 897 while adding substantive new content (subagent_type table, full Agent argument schema, options section, \"never delegate understanding\", \"read the diff not the summary\", Red Flags section). Removes a dated 2025-10-03 narrative example because `writing-skills` explicitly flags narrative examples as an anti-pattern.

Yes. The skill being modified is already in the core library and applies to every Claude Code user dispatching parallel agents. The rewrite is generic to Claude Code's Agent tool, not specific to any project, team, or third-party integration. The subagent_type table mentions plugin-provided types (`feature-dev:*`, `posthog:error-analyzer`) as examples of what *might* be available; the table is framed around *job → type*, not specific plugins.

1. **Targeted patch (the #1470 approach).** Add only the single-message-N-blocks subsection. Pro: minimal diff. Con: leaves five other concrete teaching gaps in place. Reasonable to land first, but the larger rewrite is still needed.
2. **New skill alongside the existing one** (e.g., `agent-tool-reference`). Pro: zero risk to the current skill. Con: discovery suffers (two skills overlap on the same trigger); description CSO conflict.
3. **Comments on #1470 to expand its scope.** Considered, but #1470 targets `main` (v5.1.0 release) and is intentionally additive/no-removal — the author was conservative. Expanding scope in-place would change the contract of that PR.

Landing #1470 first and then this rewrite is fine (small rebase, no substance loss). Landing this rewrite supersedes #1470's content (the one-message rule is even more prominent here).

No. Single file, single skill, single concern (modernizing the Agent dispatch guidance).

## Existing PRs
- Related PRs:
  - **#1470 (open)** — Adds the single-message-N-blocks subsection additively, targeting `main`. Substantial overlap on that one point; this PR subsumes it.
  - **#948 (open)** — Aligns the \"2+ vs 3+\" threshold. This PR already uses \"2+\" throughout, consistent with #948's direction.
  - **#1102 (closed)** — Targeted redundancy removal; closed without comment. This PR doesn't repeat that exact dedup; it does a fuller rewrite with empirical grounding and eval evidence the closed PR lacked.

## Environment tested

| Harness        | Harness version | Model        | Model version/ID         |
|----------------|-----------------|--------------|--------------------------|
| Claude Code    | current         | Claude Opus 4.7 (1M context) | `claude-opus-4-7[1m]` |

Not applicable — no harness changes.

## Evaluation

- **Initial prompt:** A user asked their main agent to \"analyze our `superpowers:dispatching-parallel-agents` and determine if it's outdated or lacking quality.\" The agent identified the gaps above through analysis of the skill text against the live Claude Code Agent tool definition.
- **RED baseline:** 1 session. Dispatched a fresh `general-purpose` subagent (no skill loaded) with the parallel-test-fix scenario. Outcome documented in \"Empirical baseline (RED)\" above. The transcript explicitly used `TaskCreate`, invented `EnterWorktree` as a separate step, and got `run_in_background` semantics wrong.
- **GREEN test:** 1 session. Dispatched a fresh `general-purpose` subagent with explicit instructions to read the rewritten SKILL.md and apply it to the same scenario. Outcome: subagent used `Agent` (called out `Task`/`TaskCreate` as wrong), batched in one assistant message, picked `subagent_type=general-purpose` with reasoning, applied `isolation: \"worktree\"`, planned verification as \"read each diff (not the summary), grep for overlapping edits, run the FULL suite, spot-check that no assertions were weakened.\" Caught all RED failures.
- **REFACTOR pass:** GREEN-test subagent flagged two real gaps in the first draft: (a) worktree guidance was ambiguous on whether to default-on for parallel test fixes; (b) the full Agent argument schema was only shown by example. Both fixed before commit. Final word count 897 (down from 950).
- **Before/after:** RED ran without the skill; even without the skill, model hit three concrete dispatch errors. GREEN ran with the new skill and avoided all three. The old skill would have *reinforced* one of them (`Task(...)` is in the old code blocks).

Acknowledged limit: only one GREEN session. The maintainer's bar may be higher; happy to run additional adversarial scenarios if requested.

## Rigor

- [x] If this is a skills change: I used `superpowers:writing-skills` and completed adversarial pressure testing (RED/GREEN documented above)
- [x] This change was tested adversarially, not just on the happy path (RED baseline ran *before* writing the rewrite)
- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, \"human partner\" language) without extensive evals showing the change is an improvement. *Note:* the original skill did not have a Red Flags section; this rewrite **adds** one (per `writing-skills` recommendation). The narrative \"Real Example from Session\" with date 2025-10-03 was removed because `writing-skills` explicitly lists narrative examples as an anti-pattern.

## Human review

*Human review confirmed; marking ready.*

### DIFF
--- a/skills/dispatching-parallel-agents/SKILL.md
+++ b/skills/dispatching-parallel-agents/SKILL.md
@@ -1,185 +1,127 @@
 ---
 name: dispatching-parallel-agents
-description: Use when facing 2+ independent tasks that can be worked on without shared state or sequential dependencies
+description: Use when facing 2+ independent problems that share no state and could be worked on concurrently. Triggers include multiple failing test files, separate subsystem bugs, independent investigations, parallel research questions
 ---
 
 # Dispatching Parallel Agents
 
 ## Overview
 
-You delegate tasks to specialized agents with isolated context. By precisely crafting their instructions and context, you ensure they stay focused and succeed at their task. They should never inherit your session's context or history — you construct exactly what they need. This also preserves your own context for coordination work.
+Independent problems should be investigated concurrently, not sequentially. Each problem gets one Agent with focused scope and a cold context you construct deliberately. You stay the coordinator; the Agents do the work.
 
-When you have multiple unrelated failures (different test files, different subsystems, different bugs), investigating them sequentially wastes time. Each investigation is independent and can happen in parallel.
+**Core principle:** One Agent per independent problem domain, dispatched in a single assistant turn so they run in parallel.
 
-**Core principle:** Dispatch one agent per independent problem domain. Let them work concurrently.
-
-## When to Use
+## When to use
 
 ```dot
-digraph when_to_use {
-    "Multiple failures?" [shape=diamond];
-    "Are they independent?" [shape=diamond];
-    "Single agent investigates all" [shape=box];
-    "One agent per problem domain" [shape=box];
-    "Can they work in parallel?" [shape=diamond];
-    "Sequential agents" [shape=box];
+digraph when {
+    "2+ problems?" [shape=diamond];
+    "Independent?" [shape=diamond];
+    "Same files / shared state?" [shape=diamond];
+    "Investigate together" [shape=box];
+    "Use isolation: worktree" [shape=box];
     "Parallel dispatch" [shape=box];
 
-    "Multiple failures?" -> "Are they independent?" [label="yes"];
-    "Are they independent?" -> "Single agent investigates all" [label="no - related"];
-    "Are they independent?" -> "Can they work in parallel?" [label="yes"];
-    "Can they work in parallel?" -> "Parallel dispatch" [label="yes"];
-    "Can they work in parallel?" -> "Sequential agents" [label="no - shared state"];
+    "2+ problems?" -> "Independent?" [label="yes"];
+    "Independent?" -> "Investigate together" [label="no - shared root cause"];
+    "Independent?" -> "Same files / shared state?" [label="yes"];
+    "Same files / shared state?" -> "Use isolation: worktree" [label="yes"];
+    "Same files / shared state?" -> "Parallel dispatch" [label="no"];
+    "Use isolation: worktree" -> "Parallel dispatch";
 }
 ```
 
-**Use when:**
-- 3+ test files failing with different root causes
-- Multiple subsystems broken independently
-- Each problem can be understood without context from others
-- No shared state between investigations
+**Use when:** 2+ failing test files with different root causes, multiple broken subsystems, independent research questions, parallel code reads across unrelated modules.
 
-**Don't use when:**
-- Failures are related (fix one might fix others)
-- Need to understand full system state
-- Agents would interfere with each other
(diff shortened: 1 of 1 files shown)


obra:
Closing — this is a wholesale rewrite (83 insertions / 141 deletions) of a carefully-tuned skill, and the new content is very Claude-Code-specific (the `Agent` tool name, `subagent_type` taxonomy, `isolation: "worktree"`, `SendMessage` continuation, the table of plugin-provided subagent types). Superpowers tries to keep core skill prose harness-neutral; that's exactly what PR #1486 established as a maintained invariant, and #1603 is adding a lint to keep it from regressing.

> Skills are not prose — they are code that shapes agent behavior. … Do not modify carefully-tuned content … without evidence the change is an improvement.

The empirical RED probe in the PR body is appreciated, but a RED probe isn't a GREEN — there's no adversarial pressure-test showing the rewritten skill actually produces better behavior across multiple sessions on multiple harnesses. If you want to pursue this, please:

1. Keep the diff Claude-Code-specific content inside a Claude-Code-only section, leaving the harness-neutral guidance harness-neutral.
2. Show before/after eval results across at least two harnesses, not just one fresh-subagent probe.
3. Submit as a smaller targeted patch (closer to what #1470 proposed) rather than a full rewrite.

Issue #571, opened 2026-02-28
session-start hook hangs on bash 5.3+ due to heredoc bug with large variable expansion


pds:
## Bug

The `session-start` hook hangs indefinitely on systems where `/usr/bin/env bash` resolves to **bash 5.3.9** (e.g., installed via Homebrew on macOS). Claude Code never finishes starting and the hook process must be killed manually from another terminal.

The system bash (`/bin/bash` v3.2 on macOS) works fine.

## Root Cause

Bash 5.3 has a bug where heredoc variable expansion hangs when the expanded content exceeds ~512 bytes. The `session-start` script uses a `cat <<EOF` heredoc that expands `${session_context}` — a ~4,500 byte string containing the escaped SKILL.md content. This triggers the bash 5.3 hang.

Lines 41-49 of `hooks/session-start`:
```bash
cat <<EOF
{
  "additional_context": "${session_context}",
  "hookSpecificOutput": {
    "hookEventName": "SessionStart",
    "additionalContext": "${session_context}"
  }
}
EOF
```

## Minimal Reproducer

```bash
# Hangs in bash 5.3, works in bash 3.2
/opt/homebrew/bin/bash -c '
s=$(printf "%0.s=" {1..520})
cat <<EOF
${s}
EOF
'
```

Anything ≤ ~512 bytes works; anything larger hangs.

## Suggested Fix

Replace the `cat <<EOF` heredoc with `printf`, which is not affected:

```bash
printf '{\n  "additional_context": "%s",\n  "hookSpecificOutput": {\n    "hookEventName": "SessionStart",\n    "additionalContext": "%s"\n  }\n}\n' "$session_context" "$session_context"
```

## Environment

- macOS 15 (Darwin 25.3.0, arm64)
- Homebrew bash: GNU bash 5.3.9(1)-release (aarch64-apple-darwin25.1.0)
- System bash: GNU bash 3.2.57(1)-release (arm64-apple-darwin25)
- Superpowers plugin version: 4.3.1

lukelabonte:
## Additional Context: Claude Code v2.1.72 Trigger + Secondary Bug

I hit this same deadlock and traced it to the root cause. Adding some details that may be useful for the fix.

### Why This Started — Claude Code v2.1.72

**Claude Code v2.1.72** (released March 10, 2026) changed bash command parsing to use a native module:

> "Improved bash command parsing by switching to a native module — faster initialization and no memory leak"

[...]

PR #1612, opened 2026-05-23
fix: handle EPIPE on SessionStart hook stdout (Windows)

silvertakana:
## Summary

On Windows + Git Bash, the SessionStart hook prints a confusing diagnostic at every startup:

\`\`\`
printf: write error: Permission denied
/systematic-debugging
\`\`\`

The hook still completes and context still gets injected, but this noise surfaces at every Claude Code session start on Windows.

## Repro

- Windows 11 (NT 10.0.26200), Claude Code 2.1.x, Git for Windows bash 5.2
- superpowers 5.1.0 installed via the official marketplace
- Open any Claude Code session; observe the diagnostic printed before the first prompt

## Root cause

Git Bash / MSYS surfaces SIGPIPE/EPIPE on bash builtins as \`Permission denied\` (not \`Broken pipe\`). When Claude Code closes the hook's stdout pipe early, the bare \`printf\` calls fail and a fragment of the next JSON chunk (e.g. \`/systematic-debugging\`) becomes visible in the terminal before bash can suppress it.

## Fix

Pipe each \`printf\` through \`cat\`. \`cat\` is an external process whose SIGPIPE termination does not propagate to the parent script under \`set -euo pipefail\`. The partial-chunk diagnostic disappears because \`cat\` owns the failing write.

\`\`\`diff
-  printf '{\n  "additional_context": "%s"\n}\n' "$session_context"
+  printf '{\n  "additional_context": "%s"\n}\n' "$session_context" | cat

-  printf '{\n  "hookSpecificOutput": ...}\n' "$session_context"
+  printf '{\n  "hookSpecificOutput": ...}\n' "$session_context" | cat

-  printf '{\n  "additionalContext": "%s"\n}\n' "$session_context"
+  printf '{\n  "additionalContext": "%s"\n}\n' "$session_context" | cat
\`\`\`

No behaviour change on macOS / Linux — the \`cat\` passthrough is transparent when stdout is not closed early. Preserves \`set -euo pipefail\` and the \`printf\`-based output (no heredoc regression, see #571).

## Verification

- Windows 11 + Git Bash 5.2 + Claude Code 2.1.x: diagnostic gone, context injection still works
- macOS 14 + bash 5.3 (Homebrew): no change, no hang, no regression on #571
- Linux + bash 5.x: no change

## Notes

- 3 LOC, one concern, no unrelated changes
- Not fork-specific — affects every Windows user of the plugin
- Pre-submission checklist: searched existing PRs/issues (none for this), confirmed real repro, change belongs in core, diff reviewed before submitting

### DIFF
--- a/hooks/session-start
+++ b/hooks/session-start
@@ -45,13 +45,13 @@ session_context="<EXTREMELY_IMPORTANT>\nYou have superpowers.\n\n**Below is the 
 # See: https://github.com/obra/superpowers/issues/571
 if [ -n "${CURSOR_PLUGIN_ROOT:-}" ]; then
   # Cursor sets CURSOR_PLUGIN_ROOT (may also set CLAUDE_PLUGIN_ROOT)
-  printf '{\n  "additional_context": "%s"\n}\n' "$session_context"
+  printf '{\n  "additional_context": "%s"\n}\n' "$session_context" | cat
 elif [ -n "${CLAUDE_PLUGIN_ROOT:-}" ] && [ -z "${COPILOT_CLI:-}" ]; then
   # Claude Code sets CLAUDE_PLUGIN_ROOT without COPILOT_CLI
-  printf '{\n  "hookSpecificOutput": {\n    "hookEventName": "SessionStart",\n    "additionalContext": "%s"\n  }\n}\n' "$session_context"
+  printf '{\n  "hookSpecificOutput": {\n    "hookEventName": "SessionStart",\n    "additionalContext": "%s"\n  }\n}\n' "$session_context" | cat
 else
   # Copilot CLI (sets COPILOT_CLI=1) or unknown platform — SDK standard format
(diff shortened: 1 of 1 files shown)


obra:
Fixed on \`dev\` in f36bad5 with @silvertakana credited as Co-authored-by. The fix is the same shape your PR proposed: pipe each \`printf\` through \`cat\` so the external cat process absorbs the EPIPE that Git Bash reports as "Permission denied".

Verified locally:
- Existing \`tests/hooks/test-session-start.sh\` still passes (7/7 on macOS).
- Manual run on Windows 11 + Git Bash 5.2 + Node 22 produces valid JSON, clean stderr, and exit 0.
- Output is byte-identical to the unpatched hook on the happy path.

Closing this PR but the commit credits you. Thanks for the careful diagnosis — the "EPIPE-as-Permission-denied" Git Bash quirk is exactly the kind of platform-specific footgun that's hard to catch without someone hitting it on Windows.

Issue #1600, opened 2026-05-21
System prompt token overhead: conditional CLAUDE.md/AGENTS.md injection and trimmable bootstrap


vityasyyy:
## Problem

Saying "hi" in a 200K context window consumes ~13% (~26K tokens) before any user input. The superpowers plugin contributes ~3,315 tokens of system prompt content, much of it unnecessary for day-to-day work outside the superpowers repo itself.

### 1. CLAUDE.md/AGENTS.md injected unconditionally (~1,890 tokens each, ~3,780 total for the duplicate)

Both `CLAUDE.md` and `AGENTS.md` contain identical contributor guidelines for the superpowers repo. These are irrelevant for users working in their own projects. They should only be injected when the current working directory is inside the superpowers repo itself.

Current behavior: Both files are always injected via the SessionStart hook context, doubling the token cost (~3,780 tokens for the same content twice).

Proposed: Only inject CLAUDE.md/AGENTS.md when CWD is the superpowers repository. For all other sessions, skip them entirely.

### 2. `using-superpowers` bootstrap has no minimal variant (~1,240 tokens)

The full `using-superpowers` SKILL.md contains a graphviz `digraph` diagram (~200 tokens) and a 12-row Red Flags table (~300 tokens) that convey the same information as the text rules already present. Users who want a leaner prompt should be able to opt into a minimal bootstrap that keeps the essential rules but drops the diagram and verbose table.

Proposed: Offer a `"minimal"` or `"slim"` variant of the bootstrap, or allow users to configure which sections to include. The essential content (skill access instructions, priority rules, skill types, the core "invoke skills before responding" rule, red flags as a one-liner) fits in ~600 tokens.

### 3. AGENTS.md is a duplicate of CLAUDE.md

These two files contain identical content. Even when working in the superpowers repo, only one should be injected. The duplicate wastes ~1,890 tokens unconditionally.

Proposed: Merge or deduplicate. If both files serve different harness targets, the SessionStart hook should inject only the one relevant to the current platform.

## Impact

For a 200K context window, ~3,315 tokens of superpowers-specific injection represents ~1.7% of total capacity before any user input. With the proposed changes:
- Conditional CLAUDE.md/AGENTS.md: saves ~1,890-3,780 tokens depending on repo context
- Slim bootstrap option: saves ~500-800 tokens
- Deduplicating CLAUDE.md/AGENTS.md: saves ~1,890 tokens

[...]

Issue #1098, opened 2026-04-09
brainstorming: make assumptions and unknowns explicit before spec handoff


xiaomayi-ant:
- [x] I searched existing issues and this has not been proposed before

## What problem does this solve?
While reading and analyzing `superpowers:brainstorming`, I noticed the flow is strong on clarification, alternatives, and design approval, but it does not explicitly surface current assumptions or unresolved unknowns before the design is written into the spec. In practice, this makes it easier for speculative decisions made during brainstorming to harden into the written design and later implementation plan.

## Proposed solution
Add a lightweight `Assumptions / Unknowns` step to brainstorming before final design approval or as part of each design section. This should explicitly list:
- current assumptions
- unresolved unknowns
- 1-3 concrete validation questions or checks when uncertainty remains

## What alternatives did you consider?
One option is to leave this implicit in the conversation and rely on the existing spec self-review. Another is to push this responsibility entirely to `writing-plans`. I think adding a lightweight explicit step in brainstorming is better because it keeps uncertainty visible before the spec is finalized.

## Is this appropriate for core Superpowers?
Yes. This is not domain-specific or tool-specific. Making assumptions and unknowns explicit would improve brainstorming quality across many kinds of projects and reduce overcommitment in early design.

## Context
I hit this while studying the current `skills/brainstorming/SKILL.md` flow and comparing its strengths in clarification/design structure against the lack of an explicit uncertainty-capture step before spec handoff.

leadingproblemsolver:
I would make this slightly more structured than an `Assumptions / Unknowns` prose section, because the failure mode is not only visibility at handoff — it is **assumption promotion** later.

An assumption often starts explicitly uncertain, then appears in the spec/plan three turns later as if it were a fact. A small lifecycle record would make that transition auditable:

```yaml
- id: A3
  statement: "The auth service already exposes token revocation"
  status: assumed            # assumed | verified | falsified | superseded
  source:
    kind: user_statement | code_inspection | external_doc | model_inference
    ref: conversation:turn-18
  validation:
    action: "inspect AuthService public API and tests"
    result: not_run
  affects:
    - spec:auth-flow#revocation
    - plan:task-4
```

Before spec handoff, the rule can be simple:

- `verified` may become a normal spec claim;
- `assumed` must remain visibly conditional;
- `falsified` invalidates dependent design/plan items;
- unresolved high-impact assumptions block approval or require explicit user acceptance.

[...]

Issue #1096, opened 2026-04-09
Prevent developer agents from silently swallowing target-system responsibilities


SZnine:
- [x] I searched existing issues and this has not been proposed before

## What problem does this solve?
In long-running implementation tasks, a strong development agent can gradually substitute for the target system's intended responsibilities.

This usually happens when the agent is trying to complete the current task efficiently:
- it performs implicit inference that should have become explicit schema/state/policy/interface design
- it introduces helper/service logic that quietly absorbs responsibilities intended for the target agent or target subsystem
- it makes the current implementation "work" by relying on the development agent's strength, instead of forcing the product/system boundary to become explicit

The result is a subtle but important failure mode:

- the project appears to make progress
- the implementation may even pass local checks
- but some of the system's intended capabilities are not actually modeled in the system
- they were temporarily substituted by the development agent during development

This is not the same as planning/execution boundary drift.

That boundary is about who specifies implementation details vs who carries them out.
This issue is about a different boundary:
**the development agent vs the target product/system/agent being built.**

Without an explicit guard, Superpowers can accidentally encourage "capability substitution":
the development agent silently does work that should instead be made explicit in the design of the target system.

## Proposed solution
Add an explicit framework-level reminder/check in the implementation workflow.

Possible wording:

> You are the development agent, not the target product agent/system.
> Do not silently substitute for capabilities that should be explicitly modeled in the target system.
> If a behavior belongs in schema, state, policy, interface contracts, or the target agent's own responsibilities, prefer explicit modeling over implementation-time inference or hidden helper logic.

I think this could be added in one or more of these places:

1. `using-superpowers`
   - as a global boundary rule

2. `writing-plans`
   - a short check like:
   - "Are we explicitly modeling target-system responsibilities, or relying on the development agent to carry them implicitly?"

3. `executing-plans`
   - before implementing new inference/derivation/helper logic, require a check:
   - "Is this implementing the target system, or compensating for missing target-system structure?"

4. reviewer/checklist prompts
   - include a review question such as:
   - "Does this change make the system more explicit, or does it hide target responsibilities inside development-time convenience logic?"

## What alternatives did you consider?
1. Keep this as a personal `AGENTS.md` / local-skill rule only.
   - This works for individual users, but the failure mode seems general enough to deserve core awareness.

2. Rely on existing planning/execution separation.
   - Helpful, but insufficient.
   - A plan can still be clean while implementation quietly substitutes for target-system responsibilities.

[...]

PR #1618, opened 2026-05-23
docs: add task triage contract

YOMXXX:
## Summary

- Adds a lightweight task triage contract to `using-superpowers`
- Classifies work into direct answer, small edit, debugging, and multi-step/ambiguous feature before launching heavyweight document workflows
- Preserves existing skill rules: triage cannot skip requested or clearly applicable skills, and users can still request the full workflow
- Adds a deterministic shell regression test

Refs #1061.
Related: #1600, #1098, #1096.

## Why this shape

This solves the common "minor task triggers full workflow" cost problem without adding a new slash command or broad new skill. It only clarifies workflow weight selection before producing brainstorming/spec/plan artifacts.

## Test plan

- [x] `bash tests/claude-code/test-task-triage-contract.sh`
- [x] `bash -n tests/claude-code/test-task-triage-contract.sh`

## Notes

### DIFF
--- a/skills/using-superpowers/SKILL.md
+++ b/skills/using-superpowers/SKILL.md
@@ -49,6 +49,23 @@ Skills speak in actions ("dispatch a subagent", "create a todo", "read a file") 
 
 **Invoke relevant or requested skills BEFORE any response or action.** Even a 1% chance a skill might apply means that you should invoke the skill to check. If an invoked skill turns out to be wrong for the situation, you don't need to use it.
 
+## Task Triage Before Heavy Workflows
+
+Before launching heavyweight document workflows such as brainstorming/spec/plan
+artifacts, classify the user's task size. This determines workflow weight, not
+whether requested skills must be checked.
+
+- **Direct answer / command-only:** answer the question or run the requested
+  command after checking applicable skills.
+- **Small edit:** clear, localized work that appears to touch 1-3 files and has
+  no new architecture or product ambiguity. Use the relevant implementation,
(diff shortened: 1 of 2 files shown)


obra:
Closing.

This PR directly contradicts the deliberate design of the skill it's modifying.

The current `using-superpowers/SKILL.md` Red Flags table explicitly enumerates the rationalizations to reject:

- "This is just a simple question" → "Questions are tasks. Check for skills."
- "The skill is overkill" → "Simple things become complex. Use it."

The proposed "Task Triage Before Heavy Workflows" section adds: *"Small edit: clear, localized work that appears to touch 1-3 files... Do not launch brainstorming/spec/plan artifacts for small edits unless the user asks for them."*

That is the exact rationalization the existing skill is designed to suppress. Adding prose that tells the model "for small edits, skip brainstorming" undoes the behavioral work the skill does. We deliberately put this push-back in the bootstrap because models, given any latitude to decide a task is "simple," will use that latitude to skip the design phase — and the design phase is where most of Superpowers' quality leverage comes from.

This is also exactly the kind of change the contributor guidelines explicitly call out:

> Do not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) without extensive evals showing the change is an improvement.

The PR adds zero adversarial eval evidence. The Rigor checkboxes are unchecked.

Referenced issue #1061 is a legitimate community question, but the answer there is "we deliberately don't triage," not "here's a triage mechanism." I'll respond on #1061 with that context.

PR #1619, opened 2026-05-23
docs: add SessionStart prompt budget guard

YOMXXX:
## Summary

- Documents a SessionStart prompt budget boundary in both Claude and Codex hooks
- Locks the hook contract to only inject `skills/using-superpowers/SKILL.md`
- Adds a regression test proving SessionStart output does not include repository contributor files such as `CLAUDE.md` / `AGENTS.md`

Refs #1600.

## Why this shape

This is a narrow first slice for prompt-budget work. It does not introduce a slim bootstrap, config flag, or harness-specific behavior. It only prevents Superpowers' own hooks from becoming responsible for repo-instruction injection, which host harnesses may handle separately.

## Test plan

- [x] `bash tests/hooks/test-session-start-prompt-budget-guard.sh`
- [x] `bash tests/hooks/test-session-start.sh`
- [x] `bash -n hooks/session-start hooks/session-start-codex tests/hooks/test-session-start-prompt-budget-guard.sh`

## Notes

### DIFF
--- a/hooks/session-start
+++ b/hooks/session-start
@@ -7,6 +7,11 @@ set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 PLUGIN_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
 
+# Prompt Budget Guard:
+# Only inject `skills/using-superpowers/SKILL.md` from this hook. Do not read or
+# inject CLAUDE.md, AGENTS.md, GEMINI.md, or other repository instruction files
+# here; host harnesses own their own repo-instruction loading behavior.
+
 # Read using-superpowers content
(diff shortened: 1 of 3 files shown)


obra:
Closing — this PR doesn't actually fix #1600, and it locks in a misleading contract.

The hook code is unchanged. `hooks/session-start` already only reads `skills/using-superpowers/SKILL.md`; it has never injected `CLAUDE.md` or `AGENTS.md`. The added comment claims a contract that the code already follows, and the test verifies behavior that would pass on `dev` without this PR.

More importantly, #1600's premise is wrong about where the duplication happens. `CLAUDE.md` and `AGENTS.md` are loaded by the host harness (Claude Code, Codex, Gemini CLI, etc.) directly from the project root — independently of any Superpowers hook. The Superpowers `SessionStart` hook is not the source of that ~3,780-token overhead.

If you want to address #1600's real concern, you'd need to (a) prove the overhead is actually coming from somewhere Superpowers controls, and (b) fix the actual source. Adding "we don't do this" comments to a hook that already doesn't do it doesn't solve anyone's problem.

PR #1617, opened 2026-05-23
docs: add context pack handoff contract

YOMXXX:
## Summary

- Adds a concise `Context Pack` contract to the `writing-plans` plan header
- Updates `subagent-driven-development` to read the Context Pack once, preserve it in controller notes, and forward relevant fields to implementer/reviewer prompts
- Adds a deterministic shell regression test and wires it into the Claude Code fast test list

Refs #1616.
Related: #1238, #931, #1128, #551, #1503, #895.

## Why this shape

This keeps the proposal intentionally small: no new slash command, no persistent memory store, no feature-doc-pack mode, and no repository-wide documentation requirement. It only formalizes the minimum planning-to-execution context that subagents need to avoid rediscovery and drift.

## Test plan

- [x] `bash tests/claude-code/test-context-pack-contract.sh`
- [x] `bash tests/claude-code/test-worktree-path-policy.sh`
- [x] `bash -n tests/claude-code/test-context-pack-contract.sh tests/claude-code/run-skill-tests.sh`
- [x] `git diff --check HEAD~1 HEAD`

## Notes

No generated `docs/superpowers/specs/*` or `docs/superpowers/plans/*` files are included in this PR.

Issue #1616, opened 2026-05-23
RFC: add a Context Pack contract for planning-to-execution handoff


YOMXXX:
## Summary

Add a small **Context Pack** contract to the existing `writing-plans` → `subagent-driven-development` handoff.

This is not a new memory system, slash command, or feature-doc-pack mode. It is a lightweight plan-header convention that captures the minimum context execution agents need before work is split across sessions or subagents.

## Problem

Several current proposals point at the same underlying pressure:

- #1238 asks for feature-scoped doc packs for team workflows.
- #931 asks for plan-aware handoff/resume, but maintainers have pushed back on adding new slash commands.
- #1128 proposes domain context provider injection for sub-agent protocols.
- #551 discusses durable project memory, but a general memory system may be too broad for core Superpowers.
- #1503 focuses on context hygiene between workflow phases.
- #895 highlights that execution needs enough constraints and intent without turning plans into copy-paste code.

The common need is smaller than all of those: when a finalized plan is handed to execution, the coordinator needs a compact, stable summary of what matters for implementation.

Today that context is implied across the plan body, previous conversation, or agent memory. That creates three practical failures:

1. Subagents rediscover repo structure that planning already found.
2. Execution reloads too much plan text just to find file boundaries and verification commands.
3. Reviewers have to infer whether a PR followed the original constraints and acceptance criteria.

## Proposed solution

Add a required `Context Pack` section near the top of plans generated by `writing-plans`.

Suggested fields:

- **Relevant files:** the small set of source, test, docs, and config paths that execution is expected to touch or inspect.
- **Entrypoints:** commands, modules, APIs, UI paths, or workflows that define the behavior under change.
- **Verification commands:** the smallest useful commands that prove the planned change.
- **Acceptance criteria:** observable outcomes that reviewers can check.
- **Constraints:** architectural, compatibility, performance, UX, migration, or ownership boundaries.
- **Open questions / assumptions:** what is known, what is assumed, and what should stop execution if contradicted.

`subagent-driven-development` would then instruct the coordinator to read that pack once, preserve it in controller notes, and include the relevant parts in subagent prompts instead of making every subagent rediscover the same context.

## Non-goals

- No new slash commands.
- No new persistent memory store.
- No automatic multi-file doc pack generation.
- No new runtime dependency.
- No requirement to document the whole repository.
- No replacement for domain context provider conventions such as #1128.

## Why this fits Superpowers

[...]

PR #1602, opened 2026-05-22
docs: define plan status contract

YOMXXX:
## Summary

- Add `docs/plan-status.md` with a lightweight markdown contract for plan task status.
- Define status derivation from task headings, checkbox steps, optional `Status` / `Evidence` / `Next` metadata, and conflict handling.
- Document context-frugal resume behavior without adding a new slash command or separate state store.
- Link the contract from the README and `writing-plans` skill.
- Connect `executing-plans` and `subagent-driven-development` to the contract so they scan existing status, resume from the right task, and update plan state when possible.

## Why

Refs #1599, #1075, #931, and #1597.

This keeps the user-facing model as ordinary conversation instead of adding a new slash command. The first PR stays docs/skill-instructions only: it gives reviewers a narrow contract and execution workflow to validate before any follow-up runtime or eval harness work.

## Verification

- `git diff --check`
- `git diff --cached --check`
- `test -f docs/plan-status.md`
- `rg -ni "plan status contract|docs/plan-status.md|\.\./\.\./docs/plan-status.md|blocked|skipped" README.md skills/writing-plans/SKILL.md docs/plan-status.md`
- `rg -n "plan status contract|blocked.*resolved|remaining tasks|Do not make subagents read|current status" skills/executing-plans/SKILL.md skills/subagent-driven-development/SKILL.md docs/plan-status.md README.md`

Issue #1599, opened 2026-05-21
Add superpowers status command to show plan/spec completion


fuleinist:
## What problem does this solve?
Plans and specs generated during a Superpowers session have no persistent completion status after execution finishes. There is no built-in way to see which tasks in a plan were completed, skipped, or failed after the session ends — requiring manual tracking or external tools.

## Proposed solution
Add a  command (or integrate into the existing workflow hooks) that reads the plans directory and reports completion status of each task — completed, failed, or pending. This gives users a quick overview of what was done without replaying the entire session.

## What alternatives did you consider?
- Parsing agent output logs — fragile, not user-facing
- Relying on git history — doesn't capture individual task granularity
- Manual tracking — error-prone and tedious

## Is this appropriate for core Superpowers?
Yes — completion status tracking is a universal DX improvement that benefits anyone using the subagent-driven-development workflow, regardless of project domain or harness.

## Context
Issue #1075 filed previously, no resolution yet. Shell-based framework, affects all harnesses.

YOMXXX:
I think this is a strong direction, but I would avoid starting with a new `superpowers status` slash-style command as the first implementation shape.

Related context:

- #1075 identifies the underlying gap: plans/specs have no persistent completion status after execution.
- #931 proposed explicit handoff/resume commands, but the maintainer feedback there was to avoid adding more slash commands and keep the UX as “just talk to the agent.”
- #1597 is trying to standardize eval evidence before adding workflow state or preference-like behavior.

A smaller, core-friendly shape might be:

1. Define a lightweight **plan status contract** for generated plans: task identity, pending/in-progress/done/skipped/blocked, and optional evidence links.
2. Teach `executing-plans` / `subagent-driven-development` to report that status when asked, instead of introducing a separate command surface.
3. Support a context-frugal resume mode that reads only the plan header plus the next pending task, instead of reloading the whole plan. This also addresses the concern raised in #931 about large plans consuming too much context on resume.
4. Gate any behavior change with a small eval fixture: partially completed plan, skipped task, blocked task, and stale status case.

That would keep the user-facing model simple, avoid new commands, and still solve the core problem: users and reviewers can see what was completed, what remains, and where execution should resume.

If this direction sounds acceptable, I can help turn it into a narrow first PR focused on the plan/status contract and docs before changing execution behavior.

arittr:
[...]

PR #1621, opened 2026-05-23
docs: add context-frugal resume contract

YOMXXX:
## Summary

- Adds a context-frugal resume contract to `executing-plans`
- Adds the same resume contract to `subagent-driven-development`
- Defines a lightweight resume path: read the plan header first, then load the next pending task and only expand to the full plan when needed
- Adds a deterministic shell contract test

Related: #931, #1075, #1599, #1602, #1616, #1617.

## Why this shape

This is a small first slice for large-plan resume efficiency. It does not add slash commands, persistent state, or a new memory store. It just teaches the execution skills to avoid reloading an entire long plan merely to find status or the next unchecked task.

## Test plan

- [x] `bash tests/claude-code/test-context-frugal-resume-contract.sh`
- [x] `bash -n tests/claude-code/test-context-frugal-resume-contract.sh`

## Notes

### DIFF
--- a/skills/executing-plans/SKILL.md
+++ b/skills/executing-plans/SKILL.md
@@ -21,6 +21,14 @@ Load plan, review critically, execute all tasks, report when complete.
 3. If concerns: Raise them with your human partner before starting
 4. If no concerns: Create todos for the plan items and proceed
 
+## Context-Frugal Resume
+
+When resuming an existing plan, keep the initial reload small. Read the plan header first, then load the next pending task and only the surrounding context
+needed to execute it.
+
+Do not reload the entire plan just to find status or locate the next unchecked
+task. Expand to the full plan only when dependencies, unclear context, or cross-task coupling require it.
+
 ### Step 2: Execute Tasks
 
 For each task:
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -119,6 +119,14 @@ Implementer subagents report one of four statuses. Handle each appropriately:
 
 **Never** ignore an escalation or force the same model to retry without changes. If the implementer said it's stuck, something needs to change.
 
+## Context-Frugal Resume
+
+When resuming an existing plan, keep the controller's reload small. Read the plan header first, then load the next pending task and only the surrounding
+context needed for dispatch.
+
+Do not make subagents reload the whole plan. The controller provides task text
+and curated context. Expand to the full plan only when dependencies, unclear context, or cross-task coupling require it.
+
 ## Prompt Templates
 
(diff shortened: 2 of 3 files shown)


obra:
Thanks for the writeup. Before this can be evaluated I need three things that aren't in the PR yet:

1. **A specific session where this bit you.** The "What problem are you trying to solve" reasoning is general ("subagents rediscover... execution reloads too much... reviewers have to infer..."), but I don't see a concrete session where `executing-plans` actually misbehaved against a long plan in a way this would have prevented. Can you share the plan file, the harness, and the model behavior you observed?

2. **How you tested the new behavior.** The "Rigor" section's checkboxes are unchecked. Skill-content changes need adversarial pressure-testing per `writing-skills` — RED (current skill produces bad resume behavior on a representative long plan) and GREEN (the updated skill produces materially better resume behavior on the same plan). Can you share before/after eval evidence across more than one harness?

3. **The test you added doesn't verify behavior.** `test-context-frugal-resume-contract.sh` greps for the literal strings the PR itself added — it can never fail unless someone manually deletes those lines. That's a tautology, not a regression check. A real test would set up a long plan, run the skill, and assert that the agent's actual reload pattern matches the contract.

Separately: I'm skeptical of the underlying advice. Telling agents to "read the plan header first, then load the next pending task" risks silently dropping the cross-task context that makes plans coherent. Plans are intentionally complete artifacts. The improvement claim needs evidence stronger than "this seems lighter."

Happy to revisit if you can attach a reproducible session + before/after evals.

YOMXXX:
Reviewer-facing status update:

Current open PRs from me are now only:

- #1620 — target-system responsibility boundary
- #1621 — context-frugal resume contract

I am not asking for re-review of #1621 in its current shape. The review bar is clear:

- provide a concrete long-plan session where current execution/resume behavior actually caused context waste or wrong behavior
- add RED/GREEN behavioral eval evidence showing the revised wording changes reload behavior safely
- prove the resume guidance does not drop cross-task context that the complete plan intentionally preserves

Next step on this PR is either a revised version backed by a reproducible session and behavioral evidence, or closing it if I cannot produce that evidence.

No generated `docs/superpowers/specs/*` or `docs/superpowers/plans/*` files are part of this PR.

YOMXXX:
Closing this rather than asking for re-review without the evidence requested. The review feedback is valid: context-frugal resume needs a concrete long-plan session, RED/GREEN behavioral evals, and proof that the guidance does not drop cross-task context that complete plans intentionally preserve. I do not have that evidence ready, and the grep-based contract test is not enough to justify a skill-content change. No generated docs/superpowers/specs/* or docs/superpowers/plans/* files were part of this PR.

PR #1473, opened 2026-05-05
feat: add plan review cycle skill

scicco:
## What problem are you trying to solve?

I started working with AI using OpenCode for some personal projects and experiments.

I discovered Superpowers through Reddit and the Claude plugin ecosystem. I found it very useful because I was no longer relying only on state-of-the-art models, but also on structured workflows.

I started using the `brainstorming` skill to explore new feature ideas. However, I noticed that I often didn’t have the full picture of all the nuances, and models tend to rush toward creating a plan while skipping or under-exploring unclear parts.

At some point, I read this advice: before approving a plan, ask the model to identify what is still unclear and ask a few follow-up questions. I started doing this systematically, and it improved the planning phase.

Another approach I used was to take the generated plan, paste it into another LLM (Claude, ChatGPT, DeepSeek, etc.), ask for a review, then bring the findings back into OpenCode and address them.

Over time, this evolved into something more structured:

"I asked another model to review the document. For each point raised, create a sub-task and review it together."

Then, after discovering sub-agents, it became:

"Run a reviewer sub-agent on `@plans/myplan.md`. For each finding, create a task and let’s review it together."

I would repeat this several times until I was satisfied with the plan. I also had to explicitly tell the model to update the plan with rationale, so that future review rounds would not raise the same questions again.

## What does this PR change?

Adds a new `plan-review-cycle` skill and reviewer prompt.

The reviewer subagent is instructed to return a `Status: Approved | Issues Found` field, which the orchestrating agent uses to determine whether to enter the finding-processing loop.

The skill:

- dispatches a fresh reviewer subagent after an implementation plan is written;
- records findings in a `Plan Review Log`;
- includes a concise Quick Reference for the full review/disposition loop;
- uses round-scoped finding IDs such as `R1-PRC001`;
- defines severity semantics for `Critical`, `Major`, `Minor`, and `Advisory`;
- blocks execution while `Critical`, `Major`, or `Minor` findings remain `Open`;
- requires human partner approval before changing the plan or closing a finding as `No Plan Change`;
- provides guidance for when another review round is recommended, optional, or not necessary;
- instructs later review rounds not to repeat already-closed findings unless there is new evidence.

The core of the skill is the cycle:

1. Dispatch a fresh reviewer subagent.
2. Ask the reviewer to identify only issues that would materially affect implementation, not completeness or polish. Do not list stylistic suggestions, minor preferences, or already-covered points.
3. If the reviewer returns `Status: Approved` with no findings, skip directly to step 8.
[...]

PR #970, opened 2026-03-28
feat(writing-plans): add parallel suitability check to plan handoff

leonsong09:
## Summary

This is a narrow `writing-plans` improvement aimed at making execution routing more explicit **without** making Superpowers parallel-first.

It adds:
- per-task execution metadata in the plan template:
  - `Depends on`
  - `Write Scope`
  - `Verify`
  - `Potential Conflicts`
- a short **Parallel Suitability Check** section before execution handoff
- a handoff pattern that gives a **recommended execution mode** (`Iterative` or `Parallel-safe`) with reasoning
- plan-reviewer guidance to check whether the plan is explicit enough to choose the right execution path
- small prompt fixture updates so the explicit-skill-request prompts reflect the new handoff wording

## Why

Today `writing-plans` decomposes work into tasks, but the handoff does not make it easy to tell:
- which tasks are truly independent
- which tasks share write scope or contracts
- whether the safest next step is iterative execution or a parallel-safe path

This PR keeps the current correctness-first bias, but makes plans clearer and more actionable when some tasks are independent.

## Non-goals

This PR does **not**:
- make Superpowers parallel-by-default
- introduce a new orchestration skill
- replace heavier multi-agent flows like `agent-teams` / `assemble`
- change the recommendation that iterative execution is safest when tasks are coupled

## Related discussion

- Refs #777
- Related to #801
- Consistent with the correctness-first preference discussed in #945

## Verification

Ran targeted verification directly related to these documentation/prompt changes:
- `git diff --check`
- grep/rg checks for the new handoff markers (`Execution Recommendation`, `Parallel Suitability Check`, `Execution Readiness`)
- confirmed changed file set stayed limited to the intended 4 files

I did **not** claim broader build/test coverage because this change is limited to skill/prompt markdown and there is no direct unit/build target for it in this repo.

PR #1644, opened 2026-05-28
Improved Plan Review

andrewiethoff:
Most of the time, the _writing-plans_ skill did not find any issues on the plans I created (even though this is unlikely).
If we want to have the _subagent-driven-development_ skill work using cheaper models, the implementation plan should be as error- and contradiction free as possible. Finding errors in the plan is easier, than finding issues after a code review.
I analysed some implementation plans, they do contain a lot of source code snippets (not pseudo code). Maybe having real code snippets is the actual problem; maybe an implementation plan should not contain so much source code - but when using a cheaper model for implementation, it is probably helpful.
As the plan does contain a lot of source code, this source code should also be reviewed for correctness & consistency on basis of the existing source code. 

## What does this PR change?
It adds criteria for the review subagent to also check the source code snippets. For this, two new categories are added to the check list: Correctness and Consistency (in the table). Also it adds some additional agent description regarding code review.

It only improves plan review, finds more issues that can be resolved before implementation.
It should be compatible with every project.

For my projects I previously implemented a manual additional review phase that checks the plan (without giving any restrictions to what to check). This request usually also performs a code review.

No

## Existing PRs
- Related PRs: 
#1062,#970,#1062,#1473 (closed, not merged)

## Environment tested

|   Claude Code                                  |   2.1.142              |  opus     |  opus-4.7 1M  xhigh     |
|   Codex                                  |   0.134.0              |  gpt     |  gpt-5.5 xhigh    |

## Evaluation
- What was the initial prompt you (or your human partner) used to start
  the session that led to this change?
/brainstorming -> /writing-plans
                      \\> /writing-plans
                      \\> /writing-plans

Around 15

I executed both the existing and new plan-document-reviewer-prompt.md for testing on a fresh implementation plan (before the review loop started). Here are some of the results:

**Case 1: Original review:**    (Opus 4.7 xhigh)
```
  Read 1 file, listed 1 directory (ctrl+o to expand)

● Agent(Review plan document)
  ⎿  Done (11 tool uses · 66.4k tokens · 59s)
  (ctrl+o to expand)

● Plan Review

  Status: Approved

  The reviewer verified every file:line reference in the plan against current HEAD (8ad0f03) and confirmed all four items (sub-project 9 + M1/M2/M14) have enough
  context to be acted on without ambiguity. Sub-project 9 tasks fully cover spec sections §4.1–§4.4, §6.1–§6.5, and §7 acceptance criteria.

  Recommendations (advisory, non-blocking):
  - Task 9 Step 5 says "roughly 9 commits ahead" but enumerates 11 (3 spec + 8 task) — minor arithmetic mismatch.
  - Task 2 Step 3 specifies @texed/shared/forest/formatMap import path; applyDescriptor.ts uses the '@texed/shared' barrel. Risks section #1 already flags this as
  "match whichever style the file uses," so it's self-correcting.
  - M1 tasks (Task 7/8) reference store-local symbols (activeFormats, selectionAnchor/Head, applyFormatAtCodeRange) — all in scope inside the same closure, visible
  from the surrounding applyFormat body.

  Plan is implementation-ready.
```
**Case 1: New review:**
```
  Searched for 2 patterns, read 1 file (ctrl+o to expand)

● Agent(Review plan document)
  ⎿  Done (49 tool uses · 108.8k tokens · 4m 32s)
  (ctrl+o to expand)

● Plan Review

  Status: Issues Found

  The reviewer verified line numbers, APIs, and spec coverage against the live source. Sub-project 9 spec is fully covered by tasks 1–4. The plan's overall
  structure, M14 latent-bug analysis, and atomic-rejection contract description all check out. However, the agent surfaced several issues that would actually trip
  an implementer:

  Issues:

  - Task 8, Step 3 — design defect: Plan claims applyFormatAtCodeRange with an inactive format is a "no-op apply." It isn't — the !isActive branch in
  engine.applyFormatToRange wraps the selection. Concrete regression: EditorView.vue:118-121 calls removeFormatAtCursor('sansSerif') / ('monospace') when the user
  picks "roman." With a roman WYSIWYG selection, the new dispatch would wrap it in \textsf{}\texttt{} — the opposite of intent. Needs an explicit
  activeFormats[formatType] guard, or route directly through engine.removeFormat like the code branch does.
  - Task 3, Step 1 — wrong function name: Test calls applyDescriptor(...) but the only export in applyDescriptor.ts is applyDescriptorOrThrow.
  - Task 2, Step 3 — import path mismatch: Plan uses @texed/shared/forest/formatMap; the file's existing style is @texed/shared (barrel). Risk #1 hedges but the
  snippet still picks the deep path.
  - Tasks 2/3/4/5 — stale test scaffolding references: Plan points at scratchProjection*.test.ts and commitTransactionAtomically*.test.ts, which don't exist. Real
  analogues: descriptors.test.ts, scratchPmMap.test.ts, wysiwygAdapter.transaction.test.ts, multiAdapterRollback.test.ts.
  - Task 1, Step 1 — contradictory semicolon instruction: Inserting before unsupportedShape means no semicolons need to move at all; the prose contradicts itself,
  though the "Final ordering" snippet is correct.
  - Task 7/8 test snippets — non-writable refs: Tests directly assign store.activeEditor, store.selectionAnchor, store.selectionHead. None have public setters
  (selection state is updated only by the internal adapter watcher at editor.ts:2068-2076). Tests need a fixture forest + mock adapter approach instead of direct
  assignment.
  - Task 6, Step 3 — unverified helper import: Plan asserts collectOverlappingLeaves is in FormatEngine.ts's import block; reviewer couldn't confirm it. Implementer
   should verify and add if missing.
  - File-structure list: "(negative-list at EditorView.vue:89)" — correct file is WysiwygEditorPane.vue:89.
  - Task 9, Step 5 — arithmetic: "3 + 8 = 9" should be 11. Cosmetic.

  Recommendations:

  - Task 8: switch to either an activeFormats[formatType] guard or direct engine.removeFormat, and add an explicit "roman dropdown with no sans/mono selection"
  test.
  - Task 5: ensure the test imports the real TokenForest constructor so vi.spyOn(forestA, 'off') attaches to a real method.
  - Branch rename suggestion (feat/format-parity-and-descriptor-marks) is sensible; flag for the user.
  - Replace stale test-scaffolding references with the real files listed above.

  The Task 8 design defect is the only blocker — the rest are mechanical fixes the implementer can patch in-line.
```

**Case 2: Original review:**      (Opus 4.7 xhigh)
```
  Read 2 files (ctrl+o to expand)

● Plan Review

  Status: Approved

  Issues (if any):

  None that would block implementation. The plan is comprehensive, follows TDD discipline (failing test first → implementation), enumerates all ten handlers from
  the spec's §4.2 migration table, covers every row of the spec's §3.1 round-trip table and §3.2 error cases (including the lone-open-quote case called out as a
  spec-review iteration 1 finding), and produces self-contained reviewable commits.

  Recommendations (advisory, do not block approval):

  - Spec/plan structural deviation (intentional and beneficial): Spec §4.2 says to add tokenize_args "directly above the command dispatcher (~line 3443)" inside
  texfastd.lua. The plan instead extracts it into a new src/lua/texfast_proto.lua module so the standalone Lua test script can require it without loading the whole
  daemon. This is the right call (testability + reuse) but should be noted as a spec deviation so the implementer doesn't second-guess it when reading the spec.
  - Task 5 Step 3 arithmetic: Expected total is stated as "19 PASS … 7 + 11 + 3 + 2 = 23 if all counts above stick". The 19 is wrong (23 is correct), though the
  plan self-corrects in the parenthetical and the actionable instruction ("failure count is 0") is fine. Worth tightening to avoid implementer confusion.
  - Task 10 Step 4 (handle_prescan_images): Says "If the current body parses no args, leave it alone. If it parses one or more args, route them through
  proto.tokenize_args". Slightly hand-wavy compared to the other steps. Implementer should read the current handler before deciding, but a one-line note about which
   case currently holds would save the round-trip.
  - Task 9 Step 2 line-counting: Says "Replace the first 8 lines of the function body" and lists the boundary markers (local file, line = args_str:match(...)
  through log("INVALIDATE %s line %d", ...)). Boundary markers are unambiguous; the "8 lines" count is just a heuristic. If the current source has more or fewer
  lines, the boundary markers still work — but explicitly leaning on the markers (rather than the count) would be clearer.
  - Task 1 Step 2 baseline count: Plan asserts "10 tests" without enumerating; Task 3 Step 3 then asserts "7 formatCommand + 3 parseResponse = 17 total with new".
  Consistent, but the implementer should verify the baseline matches expectation before proceeding — otherwise the regression-guard count will be off downstream.
  - Task 6 Step 4 Lua interpreter: Plan instructs falling back to luatex --luaonly if lua is missing. Good, but reading tests/run_tests.sh upfront (Step 2 already
  does this) would let the implementer pick the right interpreter immediately rather than guessing.

  The plan is buildable as written — none of the above would cause an implementer to get stuck or build the wrong thing.
```
**Case 2: New review:**
```
  Listed 1 directory (ctrl+o to expand)

● Agent(Review plan document)
  ⎿  Done (18 tool uses · 78.6k tokens · 2m 9s)
  (ctrl+o to expand)

● Plan Review

  Status: Issues Found

  Critical issues that would block clean implementation

  1. Task 10 Step 5 — handle_set_viewport: The migration snippet drops the existing math.floor(page) ~= page integer check. Today SET_VIEWPORT 3.5 is rejected;
  under the plan it would be silently accepted. Restore the integer guard and keep the existing error wording ("SET_VIEWPORT page must be a positive integer").
  2. Task 10 Steps 8 & 9 — wrong helper name: handle_synctex_inverse (line 3371) and handle_synctex_regions (line 3406) use json.response(...), not the local
  json_response(...). Snippets as written would NameError. Pick one symbol consciously and apply it.
  3. Task 10 Step 4 — handle_prescan_images shouldn't be in scope: That handler ignores args_str entirely (no args). Spec §4.2 and Task 10 list it as a migration
  target — drop it from both.
  4. Task 13 Step 5 — TexFast PR base branch: Memory shows TexFast's working branch is texfast-initial, not the repo default. Specify gh pr create --base
  texfast-initial explicitly or the PR will target the wrong branch.
  5. Task 5 Step 3 — test count arithmetic mismatch: Says "19 PASS total" with a sub-sum of "7+11+3+2 = 23". Real total after all three test tasks is 23. Update or
  drop the count.

  Minor inconsistencies

  - Task 9 Step 1: Claims require 'proto' should sit alphabetized near preamble, but the existing require list (lines 45–55) is dependency-ordered, not
  alphabetical. Either append after preamble without claiming alphabetization, or justify position with a comment.
  - Task 11 Step 1: Doesn't verify findings.md path before editing; add a Read step.
  - Task 10 Step 7: handle_auto_render drops the explicit if not mode guard but the existing else branch covers it — functionally equivalent, worth confirming
  during implementation.

  Things the reviewer verified are correct

  - quoteArg, formatCommand, and the new unit tests for TexFastClient.ts ✓
  - The TexFast tokenizer module (proto.lua) and its test harness pattern ✓
  - General plan structure, spec alignment, task boundaries ✓

  Recommendations (advisory)

  - Split Task 10 into one commit per handler — inconsistent error messages and json_response/json.response usage across the 9 handlers make a single commit hard to
   review and bisect.
  - For the wire-format integration test (Task 5 Step 1), prefer a small __test__ injection helper over poking private fields via as any.
  - Add a Lua tokenizer test for the quoted-bare boundary case "a"b to pin behavior.
  - For Task 9 Step 4 manual smoke test, give Windows users a concrete command (PowerShell + Python one-liner or ncat) rather than netcat.

  Recommend fixing issues 1–5 before starting implementation; the rest can be handled inline.
```

**Case 3: Original review:**    (Opus 4.7 xhigh)
```
  Searched for 1 pattern, listed 1 directory (ctrl+o to expand)

● Agent(Review spell-checking plan)
  ⎿  Done (8 tool uses · 96.0k tokens · 1m 18s)
  (ctrl+o to expand)

● The reviewer flagged real issues. Summary:

  Plan Review — Issues Found

  Blocking issues:

  1. Task 1.9 — The "stale scan wins" test will pass for the wrong reason: there are no awaits between ++currentScanId and publishing results, so the abort
  mechanism the spec §3.2 requires isn't actually exercised. Either pull chunked-async into phase 1 or explicitly defer + rewrite the test.
  2. Task 2.1, Step 3 — leafSourceRange contains visibly broken/contradictory draft code before falling through to mapWithEscapes. Implementer will paste broken
  code.
  3. Task 2.1 — mapWithEscapes assumption — Relies on "parser splits the leaf on every escape char" without verifying it against LatexParser. If false, all
  decorations past the first escape land at wrong offsets. Add a verification step or use the parse-time offset table from spec §7.1.
  4. Task 2.3, Step 1 — i18n diff would produce duplicate ribbon key if pasted verbatim; the "merge by hand" instruction is fragile. Show the merged structure.
  5. Task 3.1, Step 3 — DecorationSet.create(forest.root, ...) is wrong (wants PM doc, not forest root). Plan acknowledges it as homework instead of specifying the
  fix and signature change.
  6. Task 4.1, Step 3 — Uses forest.getNodeById(id) as a method, but spec lists getNodeById(root, id) as a free function. Inconsistent — will compile-fail or get
  rewritten incorrectly.
  7. Task 4.2, Step 3 — setActiveDocumentPath wiring is a pointer ("look at document.ts"), not a concrete step. Without it, per-document ignore silently no-ops.
  Composite WYSIWYG case (multi-file) also unaddressed.
  8. Spec §10.1 vs Task 4.1 — No test/verification that WYSIWYG-sourced replacements actually flow through commitTransactionAtomically (sub-project 8's
  atomic-rejection contract). Risk: editors desync on rejection.

  Advisory (non-blocking): Consolidate API verifications into a "Task 0.5"; Task 2.2 watcher should clear/rescan on toggle, not just manage subscription; spec §14
  perf budget has no test; SpellChecker dispose() omitted without comment.

  Recommend revising the plan to fix items 1–8 before kicking off implementation.
```
**Case 3: New review:**
```
  Listed 1 directory (ctrl+o to expand)

● Agent(Review spell checking plan)
  ⎿  Done (54 tool uses · 153.2k tokens · 4m 30s)
  (ctrl+o to expand)

● Plan Review

  Status: Issues Found

  The reviewer found the plan well-organized (clear phasing, test-first discipline, faithful spec adherence) but flagged several concrete bugs in code snippets that
   would block implementation as written.

  Critical issues (block execution):
  - Task 1.8 languageDetect.ts — reads node.data.commandName / optionalArgs / args which don't exist on TokenNodeData. Real shape is data.command (with leading
  backslash), data.params: string[], and data.paramKinds?: ('optional'|'mandatory')[]. Every detection test will fail.
  - Tasks 1.7/1.9 tests — pass parse(src) (returns ParseResult, not TokenForest) into APIs typed for TokenForest. Task 1.10 then calls (forest as
  any).emitExternalChange(...) which doesn't exist on ParseResult. Needs a helper wrapping the root into a real TokenForest.
  - Task 3.1 spellCheckPlugin.ts — ChildRange is { startChildIndex, endChildIndex }, not { codeStart, codeEnd }. The filter as written is meaningless. Also
  DecorationSet.create(opts.getForest().root as any, decos) passes a TokenNode where a PM Node is required — plan acknowledges but doesn't fix.
  - Task 2.1 leafSourceRange — snippet contains dead-code scaffolding the plan says is "replaced" but doesn't actually delete. An executor copying verbatim ships
  nonsense.
  - Task 2.3 — UI ships in Phase 2 but replace logic only lands in Phase 4, so Phase 2 ends with a broken context menu. Either move replace into Phase 2 or disable
  the suggestion items.

  Other concrete issues:
  - Task 1.10 test mutates s.misspellings.value = … — Pinia setup-stores unwrap refs on the store, so this throws/no-ops. Use $patch or a setter.
  - Task 4.1 test calls ed.loadFromSource() — actual method is loadDocument().
  - Task 1.2 routes regex /^[^\s -]{1,64}$/ rejects valid hyphenated words like well-known.
  - Task 1.2 routes return res.status(400).json(...) may trip strict TS return-type checks.
  - Task 2.3 i18n JSON patch as written would create duplicate ribbon keys (invalid JSON).
  - Task 4.2 hooks setActiveDocumentPath "in the document-open path" — should be a watch(() => docStore.currentFilePath) to avoid circular deps.
  - Task 4.3 e2e uses literal text=etdipa / text=main.tex selectors that may not match the file browser's actual labels.
  - Inconsistency: Task 2.1 uses free-function getNodeById(forest.root, id); Task 4.1 uses forest.getNodeById(id). Pick one.

  Recommendations (advisory):
  - Stop delegating "verify with grep / adapt as needed" — bake known codebase shapes (TokenNodeData, createView signatures, store method names) into the plan up
  front so parallel subagents don't drift.
  - Add a unit test for the spec §5 non-blocking-startup contract.
  - Resolve naming: plan uses Pinia store, spec talks about a SpellCheckService class.
  - Add a leaf-to-Map<id, TokenNode> cache in computeDecorations — getNodeById is O(n), and the etdipa fixture could mean ~1.2M walks per scan.
  - Verify commit-trailer wording (CLAUDE.md says "Claude Opus 4.6 (1M context)", but you're now on Opus 4.7).
  - Note in the plan that some §15.3 integration tests (e.g. language-switch fragility) are deferred.

  The reviewer's bottom line: fix the languageDetect data-shape bug, the ChildRange filter, the parse() vs TokenForest confusion in tests, the broken Phase 2 menu,
  and the dead code in leafSourceRange, and the plan is ready to execute.
```

**Case 4: Original review:**  (GPT 5.5 xhigh)
```
## Plan Review

  Status: Issues Found

  Issues (if any):

  - [Task 7, Step 8]: The plan only hides FreeDB settings in CUERipper Options.cs, but CUETools settings binds directly to CUEConfigAdvanced, where FreedbUser,
    FreedbDomain, and FreedbSiteAddress remain browsable. An implementer following the plan would leave FreeDB settings reachable in normal CUETools UI, violating
    the spec’s dormancy requirement.
  - [Task 8, Step 2]: The dependency cleanup omits CUETools/CUETools.TestCodecs/CUETools.TestCodecs.csproj, which is in CUETools.sln and still references Freedb/
    Freedb.csproj. That leaves FreeDB in the normal solution build transitively, despite removing it from the solution list.

  Recommendations (advisory, do not block approval):

  - Add the spec’s manual smoke checks to final verification: CUETools chooser source/cover display, CUERipper reload/all-metadata display, and FreeDB submit/
    settings not reachable.
  - Add a full-repo final check for Freedb.csproj, Category("Freedb"), and FreeDB setting properties.
```
**Case 4: New review:**
```
• ## Plan Review

  Status: Issues Found

  Issues (if any):

  - [Missing task before Task 8, FreeDB settings dormancy]: The plan does not wrap or hide CUEConfigAdvanced.FreedbUser, FreedbDomain, and FreedbSiteAddress in
    CUETools.Processor/CUEConfigAdvanced.cs:54. Since CUETools/frmSettings.cs:47 binds CUETools’ advanced settings grid directly to _config.advanced, FreeDB
    settings would remain reachable in normal CUETools UI, violating the spec.
  - [Task 7, Step 5]: The CUERipper plan adds AccurateRip Meta entries but does not update the cover-art worker to consume data.selectedRelease.metadata.AlbumArt.
    Current CUERipper artwork loading only iterates cueSheet.CTDB.Metadata in CUERipper/frmCUERipper.cs:1720, using the selected release’s album art only as a
    filter. AccurateRip Meta cover URLs would therefore not preview or populate the existing albumArt list used during rip setup.
  - [Task 2, Step 1]: The mapper adds any non-empty _arturl directly to AlbumArt without validating it. The spec says invalid _arturl should still show metadata
    without cover art; this matters because CUERipper’s download path reads uri150 directly, and CUEToolsDB.FetchFile creates the request before its catch block in
    CUETools.CTDB/CUEToolsDB.cs:155.

  Recommendations (advisory, do not block approval):

  - Spec used: docs/superpowers/specs/2026-05-27-accuraterip-meta-design.md.
  - Add a verification grep for Category("Freedb"), FreedbUser, and FreedbDomain, not only FreedbSiteAddress.
  - Strengthen the CUESheet lookup tests so one case proves AccurateRip Meta is added even when CTDB already returned metadata.
```

## Rigor

If you changed wording in skills that shape agent behavior, show your
     eval methodology and results. These are not prose — they are code.

The _writing-skills_ skill changed/added the additional description to the agent behaviour.

## Human review


andrewiethoff:
PR has been rebased to dev

PR #1646, opened 2026-05-29
Fix OpenCode skill fallback

### LINKED ISSUE #1492: [OpenCode] skill(name=...) fails: skills are registered but the skill function tool is absent from agent's tool list
## Summary

After deep investigation into why `skill(name="brainstorming")` fails in OpenCode, the root cause is **not** that skills aren't registered — they **are** correctly discovered via `config.skills.paths`. The real problem is that the `skill` **function tool** (the tool the LLM agent calls to load a skill) is absent from the agent's tool list.

## Environment

- **OpenCode**: v1.14.40
- **Model**: `opencode-go/deepseek-v4-flash` (Sisyphus agent)
- **Superpowers**: v5.1.0 installed via `opencode.json` plugin
- **Config**: `~/.config/opencode/opencode.json` has `"plugin": ["superpowers@git+https://github.com/obra/superpowers.git"]`

### 1. Plugin loads successfully
```
INFO service=plugin path=superpowers@git+https://github.com/obra/superpowers.git loading plugin
```

### 2. Skills are discovered
`opencode debug skill` lists all 14 superpowers skills with correct names and locations:
```json
{
  "name": "brainstorming",
  "location": "/Users/.../.cache/opencode/packages/superpowers@.../node_modules/superpowers/skills/brainstorming/SKILL.md"
}
```

### 3. Bootstrap injection works
The `using-superpowers` content is successfully injected into the first user message of every session via the `experimental.chat.messages.transform` hook.

## Evidence: Function Tool Missing ❌

The agent's available function tools are limited to:
- `bash`
- `read`
- `glob`
- `grep`
- `webfetch`

**No `skill` tool is present.** When the agent tries to call `skill(name="brainstorming")`, there is no function tool to receive the call — it fails silently or with "Skill or command not found."

## Root Cause Chain
```
opencode.json has plugin configured ✅
    → Plugin loads ✅
    → config hook registers skills.paths ✅
    → opencode debug skill lists brainstorming ✅
    → messages.transform injects bootstrap ✅
    → ❌ skill function tool NOT exposed to agent
    → Agent calls skill(name="brainstorming") → FAILS
    → Agent has no fallback mechanism → skips silently
```

## Why Other Agents Skip Instead of Retrying

The `using-superpowers` bootstrap instructs agents: *"IF A SKILL APPLIES... YOU MUST USE IT."* But when `skill()` fails without the OpenCode error message that suggests `superpowers/brainstorming` as an alternative (which only appears when skills ARE registered but the name is wrong), the agent sees a bare "not found" error and treats the skill as unavailable. There is no documented fallback for when the `skill` function tool itself is absent.

## Related Issues

- **#1087**: Covers the `config` hook not working in older OpenCode versions. Our case is different — the config hook works fine, skills ARE discovered, but the function tool itself is absent.
- **#1479**: About `.claude-plugin/plugin.json` missing skills field (Claude Code specific).

## Proposed Solutions

**Option A (Reliable)**: Programmatic symlink creation in the plugin startup — as proposed in #1087. This bypasses the `config.skills.paths` mechanism entirely and puts skills directly into `~/.config/opencode/skills/superpowers/`, which OpenCode scans natively.

**Option B (Robust)**: Inject skill content directly via the `messages.transform` hook (similar to what's done for `using-superpowers`) as a fallback when the `skill` tool is unavailable. The plugin would detect whether the `skill` tool is in the agent's tool list and, if not, inject the requested skill's content into the user message.

**Option C (Documentation)**: Update the bootstrap text to include explicit fallback instructions when `skill()` fails: *"If the skill tool returns 'not found' without alternative suggestions, the skill tool itself may be unavailable. Fall back to: glob for SKILL.md → read it directly."*

hardikloglogn:

Fixes #1492.

In OpenCode, Superpowers skills can be discovered successfully while the native `skill` function tool is still absent from the agent's available tool list. In that state, the injected bootstrap tells the agent to load skills through `skill(...)`, but the agent has no callable tool to perform that action.

The observed result from #1492 is that an agent can see or be told about Superpowers, but when a matching skill such as `brainstorming` should be used, the skill load fails or is skipped even though the corresponding `SKILL.md` exists on disk.

## What does this PR change?

This PR adds an explicit OpenCode fallback to the injected bootstrap: if the native `skill` tool is unavailable, agents should read the relevant bundled `SKILL.md` file directly, ignore the YAML frontmatter, and follow the remaining instructions as the loaded skill.

It also documents the same fallback in the OpenCode README and install guide, and updates the bootstrap caching test to assert that the fallback instructions are injected.

Yes. This is specific to the existing OpenCode integration in core and improves the reliability of loading existing Superpowers skills when OpenCode discovers the skills but does not expose the native `skill` tool to an agent.

It does not add a third-party dependency, new domain-specific skill, or project-specific configuration.

I considered adding a custom replacement `skill` tool from the plugin, but OpenCode already has a native skill system and the issue is about cases where that native tool is absent from a specific agent's tool list. A bootstrap fallback is smaller and avoids competing with OpenCode's native implementation.

I also considered symlink creation into OpenCode's global skills directory, but #1492 reports that skill discovery already works. Symlinking targets a different failure mode covered by #1087, where skills are not discovered.

No. The code, docs, and test updates all address the same OpenCode missing-`skill`-tool fallback.

## Existing PRs

- Related PRs: none found

Related issue: #1087 covers OpenCode returning "Available skills: none" when skills are not discovered. This PR addresses a different case from #1492: skills are discovered, but the native `skill` tool is absent from the agent's available tool list.

## Environment tested

| Codex App on Windows | unavailable | GPT-5 Codex | unavailable |

Not applicable. This PR does not add a new harness.

<details>

Not applicable. This PR does not add new harness support.

</details>

## Evaluation

  The session started from issue #1492: OpenCode discovers Superpowers skills, but `skill(name=...)` fails because the native `skill` function tool is absent from the agent's tool list.

  I ran focused local verification of the OpenCode bootstrap injection behavior. I did not run multiple live OpenCode eval sessions in this environment.

  Before this change, the injected OpenCode bootstrap only told agents to use OpenCode's native `skill` tool. After this change, the bootstrap also gives agents a direct `SKILL.md` fallback path for cases where that tool is unavailable.

## Rigor

Verification run:

- `node --check .opencode\plugins\superpowers.js` passed
- `node tests\opencode\test-bootstrap-caching.mjs .opencode\plugins\superpowers.js present` passed
- Direct missing-bootstrap-file Node check passed

The bash wrapper `bash tests/opencode/run-tests.sh --test test-bootstrap-caching.sh --verbose` could not run on this Windows host because WSL has no installed Linux distribution.

## Human review

### DIFF
--- a/.opencode/INSTALL.md
+++ b/.opencode/INSTALL.md
@@ -96,6 +96,14 @@ Then use the installed package path in `opencode.json`:
 1. Use `skill` tool to list what's discovered
 2. Check that the plugin is loading (see above)
 
+### Skill tool missing
+
+If OpenCode discovers Superpowers skills but the agent does not have a `skill`
+tool in its tool list, read the needed `SKILL.md` directly from the discovered
+skill location. For example, use `opencode debug skill` to find
+`brainstorming/SKILL.md`, ignore its YAML frontmatter, and follow the remaining
+instructions as the loaded skill.
+
 ### Tool mapping
 
 When skills reference Claude Code tools:
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -72,6 +72,8 @@ export const SuperpowersPlugin = async ({ client, directory }) => {
 
     const fullContent = fs.readFileSync(skillPath, 'utf8');
     const { content } = extractAndStripFrontmatter(fullContent);
+    const displayedSkillsDir = superpowersSkillsDir.replaceAll(path.sep, '/');
+    const displayedBrainstormingPath = path.join(superpowersSkillsDir, 'brainstorming', 'SKILL.md').replaceAll(path.sep, '/');
 
     const toolMapping = `**Tool Mapping for OpenCode:**
 When skills reference tools you don't have, substitute OpenCode equivalents:
@@ -80,7 +82,14 @@ When skills reference tools you don't have, substitute OpenCode equivalents:
 - \`Skill\` tool → OpenCode's native \`skill\` tool
 - \`Read\`, \`Write\`, \`Edit\`, \`Bash\` → Your native tools
 
-Use OpenCode's native \`skill\` tool to list and load skills.`;
+Use OpenCode's native \`skill\` tool to list and load skills.
+
+**Skill Loading Fallback for OpenCode:**
+If the native \`skill\` tool is missing from your tool list, load skills by reading their files directly:
+- Bundled Superpowers skills live in \`${displayedSkillsDir}/<skill-name>/SKILL.md\`
+- For example, read \`${displayedBrainstormingPath}\` when \`skill({ name: "brainstorming" })\` is unavailable
+
+When you read a skill file directly, ignore the YAML frontmatter and follow the remaining instructions exactly as if the skill tool had loaded it.`;
 
     _bootstrapCache = `<EXTREMELY_IMPORTANT>
 You have superpowers.
--- a/docs/README.opencode.md
+++ b/docs/README.opencode.md
@@ -145,6 +145,15 @@ Then use the installed package path in `opencode.json`:
 2. Check that the plugin is loading (see above)
 3. Each skill needs a `SKILL.md` file with valid YAML frontmatter
 
+### Skill tool missing
+
+Some OpenCode agents may discover Superpowers skills but omit the native
+`skill` tool from the agent's available tools. In that case, read the relevant
+`SKILL.md` directly from the discovered skill location. For example, if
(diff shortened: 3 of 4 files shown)


obra:
Opencode *has* a skill tool. What config do you have such that you don't have a skill tool? I'm not sure that we should be supporting situations in which the skill tool has been removed.

arittr:
see https://github.com/obra/superpowers/issues/1492#issuecomment-4606093833 - thanks!

Issue #1040, opened 2026-04-03
Worktree skill doesn't enforce file operations stay within worktree


voho:
## Problem

The `using-git-worktrees` skill handles creating worktrees correctly, but nothing prevents the agent from subsequently writing files to the **main repo** instead of the worktree.

This happens when exploration/research agents return file paths from the main repo (the canonical location), and the agent passes those paths directly to Write/Edit tools without remapping them to the worktree.

## Reproduction

1. Have a worktree at e.g. `/project/.worktrees/feature-x/`
2. Use an exploration agent to find files — it returns paths like `/project/src/foo.ts`
3. Agent calls `Write` on `/project/src/foo.ts` instead of `/project/.worktrees/feature-x/src/foo.ts`
4. Changes silently land on the main branch, defeating worktree isolation

## Expected behavior

After a worktree is created, the skill (or a hook) should ensure all file operations target the worktree path, not the main repo.

## Possible solutions

1. **Post-worktree-creation hook** — a `PreToolUse` hook on `Write`/`Edit`/`Read` that validates file paths start with the worktree's working directory
2. **Skill instruction** — add explicit guidance to the worktree skill: "After creating a worktree, all subsequent file paths MUST use the worktree root as the base path. Translate any paths discovered by exploration agents."
3. **Both** — instructions as a soft guard, hook as a hard guard

A hook-based approach would be the most reliable since it doesn't depend on the agent remembering instructions.

eventual-functionality:
This seems like the same issue so I'll add it here. The plugin creates worktrees at ./.claude/worktrees correctly but then fails to use them properly. It gets confused about paths consistently.

> ● Let me verify the test is discoverable by Playwright and commit it.              
>                                                            
> ● Bash(playwright test --list 2>&1)                                             
>   ⎿  Error: Exit code 1                                                            
>                                                                                    
>      TypeError: Cannot redefine property: Symbol($$jest-matchers-object)        
>          at /workspace/.claude/worktrees/e2e-playwright-plan/node_modules/.pnpm/@vi
>      test+expect@3.2.4/node_modules/@vitest/expect/dist/index.js:667:9             
>                                                                                    
>      TypeError: Cannot redefine property: Symbol($$jest-matchers-object)
>          at /workspace/.claude/worktrees/e2e-playwright-plan/node_modules/.pnpm/@vi
>      test+expect@3.2.4/node_modules/@vitest/expect/dist/index.js:667:9
>                                                                                    
[...]

PR #1654, opened 2026-05-29
fix: writing-plans generates bare git commits; subagents drift to main branch

Edge4711:
## Problem

When a worktree is created at `.worktrees/<branch>/`, the `.worktrees/` directory is typically gitignored in the main repo. The `writing-plans` skill generates commit steps with bare `git add / git commit` (no `-C` path flag).

Subagents dispatched by `subagent-driven-development` inherit the main repo as their git context. When they run `git status` from the main path they see **nothing** (gitignored), which causes them to either fail silently or drift back to committing on the main branch — defeating worktree isolation entirely.

Relates to #1040 (file-write drift to main repo); this is the git-commit equivalent.

## Root cause

Two gaps:

1. `writing-plans/SKILL.md` — the Plan Document Header template has no Git Target section, and the Task Structure commit example uses bare `git add/commit` with no path. Subagents copy this pattern literally.

2. `subagent-driven-development/implementer-prompt.md` — the `Work from: [directory]` line is interpreted as the reading/editing root, not as the git target. There is no explanation that `.worktrees/` is gitignored and that an empty `git status` means the wrong path is being used.

## Fix

**`writing-plans/SKILL.md`**
- Add pre-writing instruction to run `git worktree list` and capture the worktree path before writing tasks
- Add a mandatory **Git Target** section to the Plan Document Header template with the worktree path and an explanation of the gitignore behaviour
- Change the Task Structure commit example from bare `git add/commit` to `git -C /absolute/path/to/worktree add/commit`

**`subagent-driven-development/implementer-prompt.md`**
- Add an explicit git warning block under `Work from:` explaining that `.worktrees/` is gitignored and an empty `git status` means the wrong path is being used

## Observed in

Confirmed on Claude Code with a Python project where `.worktrees/` is gitignored. Subagents repeatedly committed to `master` despite the worktree being on a feature branch.

### DIFF
--- a/skills/subagent-driven-development/implementer-prompt.md
+++ b/skills/subagent-driven-development/implementer-prompt.md
@@ -38,6 +38,11 @@ Task tool (general-purpose):
 
     Work from: [directory]
 
+    **Git:** ALL git commands must use `git -C [worktree path]`. The `.worktrees/`
+    directory is gitignored in the main repo — running `git status` from the main
+    repo path will show nothing. If you see an empty `git status`, you are using
+    the wrong path. Always use the worktree path from the plan's Git Target section.
+
     **While you work:** If you encounter something unexpected or unclear, **ask questions**.
     It's always OK to pause and clarify. Don't guess or make assumptions.
 
--- a/skills/writing-plans/SKILL.md
+++ b/skills/writing-plans/SKILL.md
@@ -13,7 +13,7 @@ Assume they are a skilled developer, but know almost nothing about our toolset o
 
 **Announce at start:** "I'm using the writing-plans skill to create the implementation plan."
 
-**Context:** If working in an isolated worktree, it should have been created via the `superpowers:using-git-worktrees` skill at execution time.
+**Context:** If working in an isolated worktree, it should have been created via the `superpowers:using-git-worktrees` skill at execution time. Before writing the plan, run `git worktree list` to find the worktree path. Every commit step must use `git -C <worktree-path>` — never bare `git` commands. See the Git Target section below.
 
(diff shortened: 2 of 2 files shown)


WoojinAhn:
We hit the same class of bug on v6.1.1 through a different path: after the coordinator entered an isolated worktree (via Claude Code's native `EnterWorktree` tool — not a gitignored `.worktrees/` path), an implementer subagent still opened its shell in the original launch checkout and committed directly to `main`. What happened:

1. The coordinator entered the worktree and started subagent-driven-development.
2. It dispatched an implementer subagent — in our run, **the subagent's shell remained pinned to the directory the session was launched from**; the coordinator's worktree switch didn't propagate.
3. The implementer followed its brief faithfully — wrote code, ran tests, committed. All of it in the wrong checkout, so the commit landed on `main`.

The template's `Work from: [directory]` line can't prevent this — as this PR's diagnosis says, that line reads as "where to read and edit," not as "which git checkout to commit to." Whether the worktree is invisible (gitignored) or the shell never left the launch directory, the failure is identical, so the fix belongs in the template.

Our stopgap, added to every subsequent dispatch prompt, prevented recurrences across the remaining 4 tasks of the same run:

> Your shell may start in a different checkout than the coordinator's. All work happens in `<worktree path>`. Before committing, run `git rev-parse --abbrev-ref HEAD` and confirm it prints `<branch>`; a commit on any other branch is a failed task.

Two properties seemed to matter: it states the *mechanism* (why the cwd may be wrong), and it makes the contract *verifiable* (a concrete command with an expected output). Suggest folding that check into the template.

Happy to re-test the merged template against the `EnterWorktree` path.

Issue #1652, opened 2026-05-29
brainstorming visual companion: stop-server leaves orphaned node on Windows/Git Bash


supermuskk:
## Summary

On Windows + Git Bash, the brainstorming visual companion leaves an orphaned `node server.cjs` process after `stop-server.sh`, and the owner-PID watchdog that would otherwise clean it up is disabled at startup. Orphaned companions accumulate across sessions, each still bound to its port.

Reproduced on Claude Code 2.1.156, Node v24.13.1, Git Bash (mintty), superpowers 5.1.0. The companion scripts are byte-identical in 5.0.6 and 5.1.0, so this affects both.

## Root cause

Two compounding issues:

**1. `server.pid` holds the wrapper shell's pid, not node's.**

`start-server.sh` auto-selects foreground mode on msys/mingw (lines ~68-75). In that branch:

```bash
echo "$$" > "$PID_FILE"
env ... node server.cjs
```

`node` runs as a child with a *different* pid. `stop-server.sh` kills `$(cat server.pid)` = the wrapper bash, leaving the node server orphaned and still bound to its port.

`exec` does not fix this on msys: it spawns a new pid rather than doing an `execve`. Confirmed empirically: `bash $$ = 540009`, `node process.pid = 47296`.

**2. The owner-PID watchdog is disabled at startup.**

On this platform the exec wrapper is short-lived, so the script shell sees `PPID=1` and `ps -o ppid= -p $PPID` resolves an already-dead pid. `server.cjs` then logs `owner-pid-invalid: dead at startup` and sets `ownerPid = null`, which permanently disables the auto-shutdown-on-owner-exit path (server.cjs:314-337). So neither auto-cleanup nor manual stop works; only the 30-minute idle timeout eventually reaps the process.

## Verified fix (foreground branch of `start-server.sh`)

Capture node's pid instead of `$$`:

```bash
env ... node server.cjs &
SERVER_PID=$!
echo "$SERVER_PID" > "$PID_FILE"
wait "$SERVER_PID"
exit $?
```

`wait` preserves foreground blocking semantics. After this change, `stop-server.sh` cleanly terminates the node server and frees the port (verified: msys `kill` reliably terminates a bash-launched node, WINPID confirmed gone via PowerShell).

This does not address issue #2 (owner-PID resolution on Windows); the idle timeout plus a now-working manual stop are an acceptable backstop. A Windows-aware owner-PID resolution would be a separate enhancement.

## Environment

- OS: Windows 11
[...]

PR #1592, opened 2026-05-21
fix: align windows lifecycle test with server layout

YOMXXX:
## What problem are you trying to solve?

While verifying the brainstorm visual companion tests on current `dev`, I found that `tests/brainstorm-server/windows-lifecycle.test.sh` is no longer testing the current brainstorm server. It has drifted behind three implementation changes:

1. The server entrypoint was renamed from `server.js` to `server.cjs`, but the lifecycle test still runs `node .../server.js`, which fails with `MODULE_NOT_FOUND`.
2. The server now writes state under `state/` (`state/server-info`, `state/server.pid`), but the lifecycle test still waits for `.server-info` and writes `.server.pid` at the session root.
3. Current owner-PID startup validation treats a dead-at-startup owner PID as an invalid PID resolution, logs `owner-pid-invalid`, disables owner monitoring, and keeps the server running. The lifecycle test still expected that case to self-terminate after 60 seconds.

On current `dev`, before this fix, running `bash tests/brainstorm-server/windows-lifecycle.test.sh` on macOS produced:

```text
=== Results: 0 passed, 3 failed, 3 skipped ===
```

After fixing only the entrypoint and state paths, the same test got past startup but exposed the stale owner-PID assertion:

```text
=== Results: 6 passed, 2 failed, 3 skipped ===
```

## What does this PR change?

This PR updates `windows-lifecycle.test.sh` to launch `server.cjs`, read `state/server-info`, write `state/server.pid` for the stop test, and assert the current owner-PID behavior.

It preserves the existing runtime code. The test now checks both sides of owner monitoring: dead-at-startup owner PIDs are logged as invalid and do not kill the server immediately, while a live owner process that exits after startup still causes lifecycle shutdown and logs `owner process exited`.

## Is this change appropriate for the core library?

Yes. This is a core test-harness repair for the brainstorm visual companion server. It does not add dependencies, alter runtime behavior, introduce a new harness, or add project-specific configuration. It restores coverage for existing cross-platform lifecycle behavior.

## What alternatives did you consider?

I considered only changing `server.js` to `server.cjs`, but that still left the test failing because the server no longer writes `.server-info` at the session root.

I considered only updating paths and changing the dead-at-startup PID expectation to "server survives", but that would remove the control that verifies owner monitoring still shuts down when a valid owner exits. Instead, this PR keeps both cases: invalid owner at startup survives, valid owner that exits shuts down.

[...]

PR #1655, opened 2026-05-30
fix: foreground mode saves node PID and clears OWNER_PID on Windows/MSYS2

nestorluiscamachopaz:

On Windows with Git Bash (MSYS2/Cygwin), running the brainstorm server leaves orphaned `node server.cjs` processes after `stop-server.sh` is called. Each orphaned process holds a port and accumulates across sessions. This is issue #1652.

Two root causes:

**1. Wrong PID saved in foreground mode.**
`start-server.sh` wrote `echo "$$"` — the bash wrapper shell's PID — to the PID file. `stop-server.sh` killed the shell wrapper, but `node` is a child of that shell and kept running. The port was never released.

**2. OWNER_PID invisible to Node.js on Windows.**
MSYS2 POSIX PIDs are not visible in the Windows PID namespace that Node.js uses. The server received a non-zero `BRAINSTORM_OWNER_PID` it could not verify, logged `owner-pid-invalid`, and disabled its watchdog — but only by accident of the new `owner-pid-invalid` behaviour. On builds where the watchdog fires on an unverifiable PID, the server would self-terminate within 60 seconds.

## What does this PR change?

In `skills/brainstorming/scripts/start-server.sh`:

1. Foreground mode now backgrounds `node` with `&`, captures its PID via `$!`, writes that to the PID file, then `wait`s — keeping the blocking behaviour while saving the correct PID.
2. `OWNER_PID` is cleared to empty on `msys*`, `cygwin*`, `mingw*`, and when `MSYSTEM` is set, so Node.js receives an empty owner PID and disables the watchdog cleanly.

Yes. The brainstorm server ships as part of core and Windows/Git Bash is a supported platform. The bug affects any Windows user of the brainstorming skill.

- **`exec node`**: replaces the shell with node directly — but `exec` prevents writing the PID file before node starts, creating a race condition.
- **`pkill -f server.cjs`**: kills unrelated node processes.
- **Only fix the OWNER_PID**: leaves the wrong-PID-in-file bug, so `stop-server.sh` still cannot kill the process.

The `& / $! / wait` pattern is the minimal correct fix with no race condition and preserved blocking behaviour.

No. Both changes are in the same file and fix the same bug (#1652): orphaned node processes on Windows. They are causally linked — fixing only one leaves the other failure mode active.

## Existing PRs
- Related PRs: #936 (OWNER_PID fix, closed draft), #879 (WSL owner PID, closed), #751 (stop-server verification, merged), #1592 (test alignment, closed — maintainer landed a targeted repair as eef50b9)

#936 and #879 addressed OWNER_PID only and did not fix the foreground shell-PID bug. #1592 fixed the test file but not the script itself. This PR fixes both root causes.

## Environment tested

| Claude Code | 2.1.156 | Claude Sonnet | claude-sonnet-4-6 |

Platform: Windows 11 Home 10.0.26200, Git Bash (OSTYPE=cygwin, MSYSTEM=MINGW64), Node v24.16.0

Full test suite run:

```
=== Brainstorm Server Windows Lifecycle Tests ===
Platform: cygwin
MSYSTEM: MINGW64
Node: v24.16.0

=== Results: 12 passed, 0 failed, 0 skipped ===
```

## New harness support

N/A

## Evaluation

- Initial prompt: brainstorm server left orphaned node processes on Windows after stop-server.sh (issue #1652)
- Before: `stop-server.sh` returned `{"status":"stopped"}` but node kept running — confirmed via `ps -W` and `tasklist`
- After: `stop-server.sh` kills node cleanly; `ps -W` confirms the process is gone and the port is released. Verified manually across multiple start/stop cycles.

## Rigor

- [x] I did not modify carefully-tuned content without evals

## Human review

---
*Authored with Claude Sonnet 4.6 (claude-sonnet-4-6) via Claude Code 2.1.156 with superpowers installed. Human partner reviewed and approved the diff before submission.*

### DIFF
--- a/skills/brainstorming/scripts/start-server.sh
+++ b/skills/brainstorming/scripts/start-server.sh
@@ -107,10 +107,23 @@ if [[ -z "$OWNER_PID" || "$OWNER_PID" == "1" ]]; then
   OWNER_PID="$PPID"
 fi
 
+# Windows/MSYS2: Node.js cannot see POSIX PIDs from the MSYS2 namespace.
+# Passing a PID node cannot verify causes server to log owner-pid-invalid
+# and self-terminate at the 60-second lifecycle check. Clear it so the
+# watchdog is disabled and the idle timeout becomes the only shutdown trigger.
+case "${OSTYPE:-}" in
+  msys*|cygwin*|mingw*) OWNER_PID="" ;;
+esac
+if [[ -n "${MSYSTEM:-}" ]]; then
+  OWNER_PID=""
+fi
+
 # Foreground mode for environments that reap detached/background processes.
 if [[ "$FOREGROUND" == "true" ]]; then
-  echo "$$" > "$PID_FILE"
-  env BRAINSTORM_DIR="$SESSION_DIR" BRAINSTORM_HOST="$BIND_HOST" BRAINSTORM_URL_HOST="$URL_HOST" BRAINSTORM_OWNER_PID="$OWNER_PID" node server.cjs
+  env BRAINSTORM_DIR="$SESSION_DIR" BRAINSTORM_HOST="$BIND_HOST" BRAINSTORM_URL_HOST="$URL_HOST" BRAINSTORM_OWNER_PID="$OWNER_PID" node server.cjs &
+  SERVER_PID=$!
+  echo "$SERVER_PID" > "$PID_FILE"
+  wait "$SERVER_PID"
   exit $?
 fi
 


arittr:
Verified on Windows 11 Pro / Git Bash MINGW64 with Node v26.2.0.

- `bash tests/brainstorm-server/windows-lifecycle.test.sh`: 12 passed, 0 failed
- Manual start/stop smoke confirmed `server.pid` stores the Node PID
- `stop-server.sh` returned `{"status":"stopped"}`
- The server port was closed after stop
- No `node server.cjs` / brainstorm processes remained

Good to merge.

PR #1672, opened 2026-06-01
finishing-a-development-branch: don't treat harness worktrees (e.g. .claude/worktrees/) as superpowers-owned

ceaton-code:
## Problem

Step 6's provenance check decides whether to `git worktree remove` a worktree based on whether its path is "under `.worktrees/`, `worktrees/`, or `~/.config/superpowers/worktrees/`." The bare `worktrees/` substring also matches **harness-managed** worktree roots — notably Claude Code's `.claude/worktrees/<name>/`. A literal reading can therefore classify a harness-owned worktree as superpowers-owned and remove it — the exact phantom-state failure the skill's own Red Flags ("Clean up worktrees you didn't create") warn against.

Hit in practice: a session running inside `.claude/worktrees/<name>` matched the bare `worktrees/` rule.

## Fix

Replace the prose test with a precise `case` match anchored to the repo root and the superpowers home (`$MAIN_ROOT/.worktrees/`, `$MAIN_ROOT/worktrees/`, `$HOME/.config/superpowers/worktrees/`). Anything else → `OWNER=harness` → leave it alone. The matching **Common Mistakes** bullet is updated to stay consistent.

Docs-only change to one skill. No behavior change for genuinely superpowers-created worktrees; it only stops false-positives on harness dirs like `.claude/worktrees/`.

### DIFF
--- a/skills/finishing-a-development-branch/SKILL.md
+++ b/skills/finishing-a-development-branch/SKILL.md
@@ -180,16 +180,27 @@ WORKTREE_PATH=$(git rev-parse --show-toplevel)
 
 **If `GIT_DIR == GIT_COMMON`:** Normal repo, no worktree to clean up. Done.
 
-**If worktree path is under `.worktrees/`, `worktrees/`, or `~/.config/superpowers/worktrees/`:** Superpowers created this worktree — we own cleanup.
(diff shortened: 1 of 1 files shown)


obra:
You ignored the mandatory pull request template. Calling this a "docs only" change to a skill strongly suggests that this is  fully agentic PR that has not had human review or testing. Could you start by opening a github issue that describes in detail what actually happened and what the consequences were?

PR #1681, opened 2026-06-03
Handle missing sync script argument values

VeraPyuyi:
## Summary
- add explicit value validation for `--local` and `--base` in `scripts/sync-to-codex-plugin.sh`
- return usage status 2 with `Missing value for <option>` instead of leaking Bash `set -u` / `unbound variable` errors
- add regression coverage for omitted values and next-token-is-option cases

## Validation
- `bash -n scripts/sync-to-codex-plugin.sh`
- `bash -n tests/codex-plugin-sync/test-sync-to-codex-plugin.sh`
- focused malformed-argument checks for `--local`, `--base`, `--local --base`, and `--base --local` -> all returned 2, printed usage, and avoided `unbound variable`
- `git diff --check`

Note: the full `tests/codex-plugin-sync/test-sync-to-codex-plugin.sh` harness still needs `rsync`; this local Windows Git Bash environment does not have `rsync`, so the broader sync-preview assertions fail before reaching their normal path here. The new missing-argument block was verified directly and by reviewer.

### DIFF
--- a/scripts/sync-to-codex-plugin.sh
+++ b/scripts/sync-to-codex-plugin.sh
@@ -146,12 +146,22 @@ usage() {
   exit "${1:-0}"
 }
 
+require_value() {
+  local option="$1"
+  local value="${2:-}"
+
+  if [[ -z "$value" || "$value" == -* ]]; then
+    echo "Missing value for $option" >&2
+    usage 2
+  fi
+}
+
 while [[ $# -gt 0 ]]; do
   case "$1" in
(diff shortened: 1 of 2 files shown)


obra:
This is an internal tool that we're not taking PRs against.

VeraPyuyi:
Thanks for clarifying that this sync script is internal tooling and not an area you're taking PRs against. I'll close this one and avoid that surface for future contributions.

PR #1688, opened 2026-06-05
Verify subagent self-reports against on-disk state

theoju:
## Summary

Adds mechanical post-conditions to the SDD loop so an implementer's `status: DONE` or a spec-reviewer's `concur` can be verified against the actual git tree before the controller proceeds.

## Motivation

Self-reports are generated by the model, not by the work — there's no inherent coupling.

A real incident in our orchestration runtime: an implementer subagent returned `status: DONE` with a confident concur message and applied **zero file edits**. The spec-reviewer then operated on the unchanged tree, saw nothing wrong, and passed. The orchestrator marked the task complete. Manual recovery against a `.orig` backup caught it.

The symmetric failure exists for spec-reviewer `concur` with no actual review.

## What this PR changes

1. **`SKILL.md`** — new "Verifying Subagent Self-Reports" section documents the failure mode and the controller's verification responsibility: `git status --porcelain` plus `git log --since="<dispatch-timestamp>"` should cover at least one declared expected path before passing DONE to the spec-reviewer; explicit `findings: []` (not missing) is the only valid `concur` signal. Red Flags updated.
2. **`implementer-prompt.md`** — Report Format now requires structured fields: `changed_files`, `commit_sha`, `zero_diff_allowed` (default `false`). Subagents are explicitly told the controller verifies against `git status --porcelain` and persistent empty-diff DONE will halt the plan.
3. **`spec-reviewer-prompt.md`** — Report Format now requires `verdict`, explicit `findings: []`, `evidence.files_read`, `low_confidence`. A missing `findings` field is the silent-no-op signal → re-dispatch.

## Reference implementation

A working JS template for both gates (porcelain + log scan, longest-prefix-match against expected paths, single-retry policy, `zero_diff_allowed` opt-out, `SPEC_REVIEWER_SCHEMA`) lives at:

https://github.com/theoju/engineering-docs-agent/blob/main/docs/superpowers/templates/sdd-fidelity-gate.md

We verified the algorithm against a synthetic replay of the original incident across 4 scenarios:

| Scenario | Expected | Observed |
| --- | --- | --- |
| A: implementer DONE, NO edits | gate fires, re-dispatch | ✅ |
| B: implementer DONE, commits real edit | gate passes | ✅ |
| B2: implementer DONE, uncommitted edit | gate passes (porcelain detects) | ✅ |
| C: review-only task, `zero_diff_allowed=true` | gate skips retry | ✅ |

## Out of scope

The mechanical verification itself is left to the orchestration runtime — different runtimes will plumb the porcelain scan differently (Workflow tool JS, Python harness, etc.). This PR establishes the prescriptive contract via prompt schemas + SKILL.md guidance, not a specific implementation.

## Why draft

Opening as draft to surface for early maintainer feedback on shape and scope before requesting review. Happy to iterate on prose style, schema field naming, or split this into smaller PRs (e.g. SKILL.md prose first, prompt-schema changes second) if that's easier to land.


obra:
> . The spec-reviewer then operated on the unchanged tree, saw nothing wrong, and passed

That should not happen and that is the thing to fix. Can you pull a transcript from that reviewer's session and open an issue, please?

theoju:
Hi Obra,
Below you find my Claude summary with what was is available and and all related findings form my system. 
Hope that helps. Let me know if this needs to be filed as a different issue or not.

Thanks again,
Theo

--- Claude Code analysis ---
> The spec-reviewer then operated on the unchanged tree, saw nothing wrong, and passed
> That should not happen and that is the thing to fix.

Confirmed — that's exactly the failure mode, and the symmetry with the implementer side (silent `DONE` with no edits) is why this PR proposes structured tails on both prompts rather than just one.

Attaching a reproduction bundle: **`cce95-superpowers-1688-bundle-20260606.tgz`** (44 KB, 9 files). The `README.md` inside is the orientation doc; the structure mirrors the diagnostic chain.

---

### One honest caveat up front

The original spec-reviewer subagent transcript from the silent-concur event itself (2026-06-04T17:32:19Z, orchestrator run `b10b12-prod-1`) is **not recoverable**. Claude Code's per-dispatch subagent `.jsonl` files in the active project directories all post-date the incident by ~6 hours — after the local SKILL.md edits had been applied and the orchestrator re-ran.

The orchestrator state log is the only durable record of the original event, which is itself a small data point that argues for structured tails landing upstream: if subagent self-reports were structured and the orchestrator persisted them deliberately to state, you wouldn't be relying on the harness's incidental transcripts to forensically reconstruct what happened.

---

### What the bundle does carry

1. **`01-orchestrator-state/state-b10b12-prod-1.jsonl`** — 14-line state log. The smoking gun is three lines:
   - B10 line 3: `"transition":"sdd-reviewers-concur","detail":"0 iterations"`
   - B11 line 6: `"transition":"patch-spec-validated","detail":"0 iterations"`
   - B11 line 13: `"transition":"sdd-reviewers-concur","detail":"0 iterations"`

   Zero iterations on first dispatch is the mechanical signature.

2. **`01-orchestrator-state/orchestrator-self-narrative-of-B11-incident.txt`** — extract from a later orchestrator session where the bug is named explicitly: _"never executed because the implementer subagent reported completion without applying edits."_

3. **`02-gate-output/B11.md`** — the downstream AC-coverage-gap gate that eventually caught what the reviewer missed (4 specific missing test-case assertions). This is what a competent reviewer should have surfaced; the gate's surgical precision (`"AC3 — covering_test: Case 16... stderr assertion for '-f (on git push)' is MISSING"`) makes the contrast vivid.

4. **`03-forensic-patches/B11.patch` + `B11-ac10.patch`** — the fix that was eventually applied (`validate-git-cmd.sh`, ~+8/-2). This is what a fully-iterating implementer + reviewer pair would have produced on the first pass instead of the silent no-op.

5. **`04-reviewer-transcript-sample/post-fix-reviewer-doing-its-job.jsonl`** — contrast sample. A spec-reviewer subagent dispatched from a _later_ orchestrator run after the local floor was in place. It makes 2 `Read` tool calls, reasons through scope / risk / ordering / regex correctness, and returns `Verdict: DONE_WITH_CONCERNS` with 3 specific actionable concerns. Provided so you can diff the tool-call behavior against the silent-concur signature recorded in the state file.

6. **`04-reviewer-transcript-sample/sdd-fidelity-gate-template-from-engineering-docs-agent.md`** — the working reference implementation already linked from this PR's body. JS gate code that scans `git status --porcelain` + `git log --since=` after each implementer `DONE`, and verifies `findings: []` (explicit empty) vs missing-`findings` (silent no-op) after each reviewer `concur`. The structured-tail proposal in this PR makes both gates trivially implementable.

---

### Sanitization

- Absolute home paths replaced with `~`.
- No API tokens, GitHub PATs, or Atlassian credentials present.
- No chained context from unrelated tasks.
- The repo (`theoju/engineering-docs-agent`) is public; CCE ticket IDs and the `designitright.atlassian.net` Jira URL already appear in this PR's body and on the linked CCE-95 comment — no incremental exposure.

---

Happy to file this as a separate issue under `obra/superpowers` with the bundle attached if you'd rather a clean tracker for the diagnostic work, distinct from the fix PR. Let me know which you prefer.

PR #1695, opened 2026-06-07
docs: warn about docs-build publication hazard at docs/superpowers/ save location

### LINKED ISSUE #1690: writing-plans/brainstorming: default docs/superpowers/ location can leak internal plans into published documentation (Sphinx/Read the Docs)

We develop [okama](https://github.com/mbk-dev/okama), an open-source Python library whose documentation is built by Sphinx and published on Read the Docs. The `brainstorming` and `writing-plans` skills saved design specs and implementation plans to their default locations (`docs/superpowers/specs/`, `docs/superpowers/plans/`) and committed them — exactly as designed, and we like that design: plans are project memory and should travel with the repo.

What we didn't anticipate: Sphinx treats **every** file with a registered source suffix under `docs/` as a documentation source. With `recommonmark` (or any extension registering `.md`), that includes all superpowers plans and specs. Files not referenced by any toctree are still rendered to HTML — Sphinx only emits a `document isn't included in any toctree` warning. On the next Read the Docs build after our release merge, all internal implementation plans and design specs became public web pages at `okama.readthedocs.io/en/master/superpowers/plans/...`. We discovered it only after the release.

The hazard is structural: the default save location lives inside the directory that documentation generators commonly treat as their source root. We verified the leak with Sphinx + recommonmark; the same mechanism likely applies to other generators that auto-collect every markdown file under their source directory (we haven't verified those).

Related issues touch the location question but not the publication hazard: #337 (configurable plan location), #939 (CLAUDE.md path override ignored in practice), #975 (working-tree artifacts in open-source repos).

## Proposed solution

When `writing-plans` or `brainstorming` is about to save under `docs/superpowers/` **and** the repo shows signs of a docs build rooted at `docs/` (e.g. `docs/conf.py`, `mkdocs.yml`, `.readthedocs.yaml`, `docusaurus.config.*`), the skill should warn and offer to add a build exclusion before committing. For Sphinx it is a one-line fix in `docs/conf.py`:

```python
exclude_patterns = ["_build", "Thumbs.db", "superpowers"]
```

Even a single sentence in both skills — "If your project publishes documentation built from `docs/`, exclude `docs/superpowers/` from the docs build" — would prevent the leak.

1. **Moving plans outside `docs/`** — works, but #939 shows path overrides are unreliable in practice, and the committed-in-`docs/` default is a philosophy we want to keep, not work around.
2. **Gitignoring `docs/superpowers/`** — defeats the purpose: plans stop syncing between machines/agents via git. We tried this first and reverted.
3. **What we settled on (verified working):** keep the default location and the committed files, add `"superpowers"` to Sphinx `exclude_patterns` ([mbk-dev/okama@4aeabeb](https://github.com/mbk-dev/okama/commit/4aeabebbed4dd7b35b407b986e72e72ab7316a9a)). `exclude_patterns` filters files at the source-collection stage, so the markdown stays in git but never reaches the HTML output — the leaked URLs now return 404 while the plans remain in the repo. The problem: every user currently has to discover this hazard the hard way, after leaking.

Yes — the hazard is domain-independent. Any project that publishes documentation generated from `docs/` is exposed, which covers a large share of open-source Python/JS projects.

## Environment (required)

| Field | Value |
|-------|-------|
| Superpowers version | 5.1.0 |
| Harness version | 2.1.165 |
| Your model + version | Claude Opus 4.8 (claude-opus-4-8) |
| All plugins installed | superpowers, skill-creator, frontend-design, playwright, code-simplifier, chrome-devtools-mcp, backend-development, python-development, claude-code-setup, hookify, github, commit-commands, document-skills, rust-analyzer-lsp, warp, managing-youtrack |

## Context

- Affected project: https://github.com/mbk-dev/okama (Sphinx + recommonmark + Read the Docs)
- Fix commit: https://github.com/mbk-dev/okama/commit/4aeabebbed4dd7b35b407b986e72e72ab7316a9a
- Verification: leaked pages return 404 on Read the Docs after rebuild, while plans/specs remain committed; legitimate API pages unaffected.
- This issue was researched and filed by a Claude Code agent during the incident-response session, at the user's request.

ltianyi992:
## Summary

Fixes #1690

The default save locations for `writing-plans` (`docs/superpowers/plans/`) and `brainstorming` (`docs/superpowers/specs/`) sit inside the directory that most documentation generators treat as their source root. Sphinx, MkDocs, Docusaurus, and Read the Docs will silently render and publish every `.md` file they find there — including internal implementation plans and design specs — even without an explicit `toctree` or `nav` entry.

This has already caused a real data leak in at least one open-source project (reported in the issue).

## Fix

Add a one-sentence callout at the save-location line in both skills:

- **`skills/writing-plans/SKILL.md`** — callout below the `Save plans to:` line
- **`skills/brainstorming/SKILL.md`** — callout below the `docs/superpowers/specs/` save instruction

The callout names the affected generators, says what to check for (docs build config files), and shows the one-line fix for Sphinx and MkDocs. No behavior change — just surfacing the hazard at the exact moment the skill is about to create the file.

## Why this approach

The issue author considered alternatives (moving plans outside `docs/`, gitignoring) and explains why the committed-in-`docs/` default is correct to keep. The simplest fix is teaching users about the `exclude_patterns` / `exclude_docs` pattern right when they'd need it.

### DIFF
--- a/skills/brainstorming/SKILL.md
+++ b/skills/brainstorming/SKILL.md
@@ -110,6 +110,7 @@ digraph brainstorming {
 
 - Write the validated design (spec) to `docs/superpowers/specs/YYYY-MM-DD-<topic>-design.md`
   - (User preferences for spec location override this default)
+  - **Documentation build warning:** If this project publishes docs from `docs/` (Sphinx, MkDocs, Docusaurus, Read the Docs), `docs/superpowers/specs/` may be rendered publicly. After committing, check for a docs build config (`docs/conf.py`, `mkdocs.yml`, `.readthedocs.yaml`, `docusaurus.config.*`) and add an exclusion if needed. Sphinx: add `"superpowers"` to `exclude_patterns`; MkDocs: add `superpowers/` under `exclude_docs:`.
 - Use elements-of-style:writing-clearly-and-concisely skill if available
 - Commit the design document to git
 
--- a/skills/writing-plans/SKILL.md
+++ b/skills/writing-plans/SKILL.md
@@ -18,6 +18,11 @@ Assume they are a skilled developer, but know almost nothing about our toolset o
 **Save plans to:** `docs/superpowers/plans/YYYY-MM-DD-<feature-name>.md`
 - (User preferences for plan location override this default)
 
+> **Documentation build warning:** If this project publishes docs generated from `docs/` (Sphinx, MkDocs, Docusaurus, Read the Docs, etc.), `docs/superpowers/plans/` may be rendered and published publicly. After saving the plan, check for a docs build config (`docs/conf.py`, `mkdocs.yml`, `.readthedocs.yaml`, `docusaurus.config.*`). If found, add an exclusion so plans stay in git but never reach the published output:
+>
+> **Sphinx** (`docs/conf.py`): `exclude_patterns = ["_build", "Thumbs.db", "superpowers"]`
+> **MkDocs** (`mkdocs.yml`): add `superpowers/` under `exclude_docs:`
+
 ## Scope Check
 
 If the spec covers multiple independent subsystems, it should have been broken into sub-project specs during brainstorming. If it wasn't, suggest breaking this into separate plans — one per subsystem. Each plan should produce working, testable software on its own.


obra:
You ignored the mandatory PR template and targeted main, rather than dev.

PR #1696, opened 2026-06-07
docs: add animated hero gif to README

livlign:
The README describes the methodology really well in words — this shows it in 20 seconds: step back and tease out the spec, write a junior-proof plan, run subagent-driven development with red/green TDD and review, all 14 skills triggering on their own.

preview

The copy is lifted from the README's own phrasing, the skill list matches the 14 in `skills/`, and the harness row matches the Quickstart. Uses your existing cape mark from `assets/`. No stars or version numbers baked in, so it won't go stale. 1200×675, ~5.3 MB, loops cleanly.

Happy to tweak pacing, copy, or drop it entirely if it's not your taste.

---
*Generated with [repo-visuals](https://github.com/livlign/claude-skills)*

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 Superpowers is a complete software development methodology for your coding agents, built on top of a set of composable skills and some initial instructions that make sure your agent uses them.
 
+![Demo: a coding agent with Superpowers steps back to tease out a spec, writes a junior-proof plan, then runs subagent-driven development with red/green TDD and code review — 14 composable skills that trigger automatically, across Claude Code, Codex, Gemini CLI, Cursor and more.](assets/hero.gif)
+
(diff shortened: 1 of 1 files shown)


dkorbelainen:
i kinda like it, but dark theme would fit better no?

livlign:
@dkorbelainen Thanks! 
Dark theme is a good idea. I also found out that GitHub supports theme-aware embeds: Dark-mode users get the dark variant, light-mode users keep the light one.
So I think we have 2 options:
1. Use one version for all users, need only a ~5mb gif.
2. Use both with theme-aware embeds, need 2 gif files in the repo.

I attached the dark theme here as well. Let's wait for what the owner think. 

hero-dark

PR #1698, opened 2026-06-07
docs(writing-plans): allow concrete discovery steps for genuinely unknown locations

### LINKED ISSUE #1295: writing-plans: clarify that discovery sub-steps are not placeholders when concrete
## Summary

The `superpowers:writing-plans` skill has a strict "No Placeholders" section that prohibits steps like *"TBD"*, *"implement later"*, *"fill in details"*, or *"steps that describe what to do without showing how"*. This is correct in spirit, but the wording can be read as also prohibiting **legitimate discovery sub-steps** — cases where the architectural decision is closed in the spec, but the exact location in the codebase wasn't explored during brainstorming because the ROI didn't justify it.

## Concrete case (real-world)

During implementation of a security-events feature, one task needed to cable an `unauthorized_access` event in the HTTP error path of a notifier. The brainstorm decided the *what/how* (event type, severity, metadata shape) but didn't read the three candidate notifier files to find which one had the catch site. The plan included:

```
Task 15 Step 1 — Discovery: locate where 401/403 is detected
  Run: grep -n -E "401|403|UNAUTHORIZED|FORBIDDEN" src/.../notifier.service.ts
  Output: identify the catch site in one of three notifier files.

Task 15 Step 2 — Inject SecurityEventsService into the identified notifier.
...
```

This worked perfectly in execution. But on a strict reading of the skill, *"step that describes what to do without showing how"* could be interpreted as banning step 1, even though it has a concrete command and produces concrete output for the next step.

## Proposal

In the "No Placeholders" section, add a clarifying paragraph distinguishing:

- **Forbidden** (placeholder): no command, no output spec, defers the decision (*"figure out where to put this"*, *"implement later"*).
- **Allowed** (discovery sub-step): concrete command (`grep`, `ls`, file read with line range), explicit output that feeds the next step. Useful when the architectural decision is closed but the codebase geography wasn't fully explored in brainstorming.

The distinction matters because forcing the brainstorm to resolve every codebase-geography question creates dead weight when the architecture decision is already settled — the only thing left is to *find the file*, which a 30-second grep does better than 30 minutes of preemptive code reading.

## References

- Workspace where this surfaced: SPC backend (private), local mitigation: bullet in the calling skill (`resolving-issues`) clarifying the case. We'd prefer the upstream skill document this directly.

mvanhorn:
## Summary

`writing-plans` currently lists "steps that describe what to do without showing how (code blocks required for code steps)" as a plan failure, which reads as a ban on discovery steps even when the location of a change is genuinely unknown at plan time (#1295). This clarifies the rule instead of leaving it implicit.

## Changes

- Allows discovery steps only for genuinely unknown locations: a step that runs a concrete command (grep / find / read file) is not a placeholder when it shows the exact command, the expected output shape, and the decision rule for choosing the file or line.
- Keeps the bar for code steps unchanged: later steps may reference the discovered file/line but still need concrete code or concrete transformation rules; discovery is not permission for placeholder names or invented functions.
- Moves the "code blocks required for code steps" requirement to its own note so the failure-mode bullet stays general.

Closes #1295

AI was used for assistance.

### DIFF
--- a/skills/writing-plans/SKILL.md
+++ b/skills/writing-plans/SKILL.md
@@ -110,8 +110,12 @@ Every step must contain the actual content an engineer needs. These are **plan f
 - "Add appropriate error handling" / "add validation" / "handle edge cases"
 - "Write tests for the above" (without actual test code)
 - "Similar to Task N" (repeat the code — the engineer may be reading tasks out of order)
-- Steps that describe what to do without showing how (code blocks required for code steps)
+- Steps that describe what to do without showing how
 - References to types, functions, or methods not defined in any task
+
+Discovery steps are allowed only for genuinely unknown locations: a step that runs a concrete command (grep, find, read file) to locate an implementation site is not a placeholder when it shows the exact command, expected output shape, and decision rule for choosing the file or line. Later code steps may reference the discovered file or line, but they still need concrete code or concrete transformation rules; do not use discovery as permission for placeholder names, invented functions, or "fill in based on output."
+
+Note: Code blocks are required for code steps. Shell commands can be concrete steps when they include the exact command, expected output shape, and what to do with the output.
 
 ## Remember
 - Exact file paths always


obra:
@mvanhorn this doesn't really make sense and your bot is ignoring the mandatory PR template and the AGENTS.md contribution guidelines.

PR #1702, opened 2026-06-08
fix: remove config.skills.paths injection causing duplicate skill entries

cinvymoe:
## Problem

When using the superpowers plugin with OpenCode, every skill appears twice in the skill list:
- `superpowers:brainstorming` (discovered via plugin mechanism)
- `brainstorming` (discovered via config.skills.paths)

This happens because the plugin's `config` hook injects the superpowers skills directory into `config.skills.paths`, but OpenCode's plugin mechanism already auto-discovers skills from the plugin's `skills/` directory with the `superpowers:` prefix.

OpenCode does not deduplicate skills by underlying file path, so the same SKILL.md file gets registered twice under different names.

## Solution

Remove the `config` hook that injects into `config.skills.paths`. The plugin mechanism already handles skill discovery correctly.

## Testing

1. Install superpowers plugin in OpenCode
2. Before fix: `/` shows both `brainstorming` and `superpowers:brainstorming`
3. After fix: `/` shows only `superpowers:brainstorming`

## Impact

- No functionality lost — skills are still fully discoverable via plugin mechanism
- Cleaner skill list without confusing duplicates
- Users can still add custom skills via `config.skills.paths` in their own config

## Checklist

- [x] Tested the fix locally
- [x] No breaking changes
- [x] Updated documentation (docstring)

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -2,7 +2,8 @@
  * Superpowers plugin for OpenCode.ai
  *
  * Injects superpowers bootstrap context via system prompt transform.
- * Auto-registers skills directory via config hook (no symlinks needed).
+ * Skills are auto-discovered by OpenCode's plugin mechanism with the
+ * `superpowers:` prefix — no config.skills.paths injection needed.
  */
 
 import path from 'path';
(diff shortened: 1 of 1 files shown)


obra:
Please resubmit following the contribution guidelines, targeting dev and not ignoring the mandatory PR template.

PR #1703, opened 2026-06-08
Harden brainstorm stop-server stale pid handling

VeraPyuyi:
## Summary
- harden brainstorm `stop-server.sh` so stale or reused `state/server.pid` values do not kill unrelated processes
- verify process ownership before SIGTERM and again before SIGKILL using the Node `server.cjs` entrypoint plus matching session ownership
- fail closed as `{status:stale_pid}` when ownership cannot be proven, while preserving `{status:stopped}` for real brainstorm servers
- add regression coverage for unrelated PIDs, fake `server.cjs` argv mentions, different-session brainstorm servers, paths with spaces, and real server shutdown

## Validation
- `C:\Program Files\Git\bin\bash.exe tests/brainstorm-server/stop-server.test.sh` -> 9 passed, 0 failed
- `C:\Program Files\Git\bin\bash.exe tests/brainstorm-server/windows-lifecycle.test.sh` -> 12 passed, 0 failed, 0 skipped
- `C:\Program Files\Git\bin\bash.exe -n skills/brainstorming/scripts/stop-server.sh`
- `C:\Program Files\Git\bin\bash.exe -n tests/brainstorm-server/stop-server.test.sh`
- `git diff --check`
- `git diff --cached --check`

Note: `scripts/lint-shell.sh skills/brainstorming/scripts/stop-server.sh tests/brainstorm-server/stop-server.test.sh` could not run locally because `shellcheck` is not installed on PATH.


obra:
Thanks @VeraPyuyi — verifying ownership before signaling a PID is exactly the right hardening for `stop-server.sh`.

After the 6.0 release: **6.0.0 ships ownership verification in `stop-server.sh`.** Verified against v6.0.2 (and confirmed identical on `origin/dev`):

- before sending any signal, it checks the target PID's argv carries a `--brainstorm-server-id=<id>` that matches the per-start id written to `state/server-instance-id` (`is_brainstorm_server` / `command_has_server_id`)
- it fails closed, reporting `{"status":"stale_pid"}` rather than risking SIGTERM/SIGKILL on an unrelated `node` after a reboot

That's a different implementation from this PR's (path/cwd/environ matching) but the same — arguably stronger — guarantee, and it's the work referenced in the 6.0 release notes. So I'm closing this as resolved in 6.0.

If `stop-server.sh` ever signals the wrong process on 6.0.2, please reopen or file a fresh issue with the scenario.

---

PR #1715, opened 2026-06-10
fix(skills): plans reference the spec instead of restating it — end to end

arittr:
> Targets `dev`. **Independently mergeable** — this PR contains the reference-discipline rule and every consumer of it (SDD prompts, executing-plans), so no intermediate state leaves a pipeline blind. Related but independent: #1716 (SDD proportionality), #1718 (brainstorming exception).

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Claude Fable 5 (`claude-fable-5[1m]`) |
| Harness + version | Claude Code 2.1.169 |
| All plugins installed | superpowers (this repo, `dev` checkout); quorum eval lab (`superpowers-evals`) as the testing apparatus; unrelated local ops plugins |
| Human partner who reviewed this diff | Drew Ritter (@drewritter) |

## What problem are you trying to solve?

In a six-agent quorum eval sweep (2026-06-09), `cost-spec-plan-duplication` — naive user brainstorms a small feature, then asks for the plan — **failed every agent that completed it (5/5)**: codex's plan "completely re-states all design decisions independently"; pi produced an 18KB, 683-line plan of inlined spec. The agents were obeying writing-plans' "assuming the engineer has zero context… Document everything they need to know" — self-contained-by-design plans necessarily duplicate the spec, doubling document bytes/tokens per feature and letting the two artifacts drift.

## What does this PR change?

- **writing-plans** — plans reference the spec, never restate, paraphrase, or summarize it: the spec owns WHAT/WHY, the plan owns HOW. A new `**Spec:**` header line cites the spec by path; tasks cite sections where they need context. The no-spec branch ("none — requirements:") is scoped to conversational-requirements-only, so it cannot be used to skip writing a spec that brainstorming would have produced.
- **subagent-driven-development** — new Spec Context section: the controller reads the plan-cited spec and pastes cited sections into implementer and spec-reviewer prompts; the spec reviewer's diff-only rule gains a spec-document exception. Reference discipline cannot starve the subagent pipeline.
- **executing-plans** — Step 1 reads the plan-cited spec (the non-subagent execution path).
- **brainstorming** — spec-path wording hardened against an observed `docs/specs/` shortening.

## Is this change appropriate for the core library?

Yes — general-purpose planning discipline, no domain or third-party coupling; it changes the redundancy rule between artifacts the core skills already produce, plus the consumers needed to keep every execution path whole.

## What alternatives did you consider?

[...]

PR #1732, opened 2026-06-10
fix(skills): description-level exceptions are authoritative in the routing rule

arittr:
> **Stacked on #1718** (brainstorming gate exception) — top of the SUP-333 stack (#1715 → #1716 → #1718 → this). Targets `dev` per the template requirement.

| Field | Value |
|-------|-------|
| Your model + version | Claude Fable 5 (`claude-fable-5[1m]`) |
| Harness + version | Claude Code 2.1.169 |
| All plugins installed | superpowers (this repo, `dev` checkout); quorum eval lab (`superpowers-evals`) as the testing apparatus; unrelated local ops plugins (decision-log, episodic-memory, superpowers-chrome, primeradiant-ops) |
| Human partner who reviewed this diff | Drew Ritter (@drewritter) |

An adversarial review fleet (three parallel reviewers: red-team, cross-corpus consistency, evidence verification) audited the SUP-333 stack and found that using-superpowers' routing rule contradicts #1718's description-level exception in both directions:

- **Compliant agents re-impose the cost failure:** "even a 1% chance a skill might apply → ABSOLUTELY MUST invoke" mandates invoking brainstorming for any trivial request (there is always ≥1% doubt), and the invocation itself is the measured cost event. "If an invoked skill turns out to be wrong, you don't need to use it" even endorses invoke-then-exit — the exact behavior the eval fails.
- **Cost-optimizing agents get a free skip:** "the skill is overkill" is a tabled, forbidden rationalization, but "the skill's own description says it doesn't apply" was unaddressed — leaving any skip arguably sanctioned with no counter.
- "Instructions say WHAT, not HOW. 'Add X' … doesn't mean skip workflows" names the exception's canonical case ("add a basic checkbox") as a violation.

## What does this PR change?

Adds one paragraph to The Rule: a documented exception in a skill's own description is authoritative — not invoking is compliance, not rationalization; any doubt about the exception's conditions means invoke; only the skill's description can define such an exception (agents cannot infer one). The "Add X doesn't mean skip workflows" line gains the matching qualifier.

Yes — it defines how the routing layer treats description-level scoping for ALL skills, which #1718 introduces and future skills may use. Without it the bootstrap text and skill descriptions give contradictory instructions.

1. **Leave it unreconciled** — rejected: both failure directions are live (the eval demonstrated the compliant-agent direction; the red-team demonstrated the free-skip direction).
2. **Weaken the 1% rule itself** — rejected: the rule is correct for the 99% case; the fix scopes it rather than weakens it ("any doubt means invoke" is preserved verbatim inside the new paragraph).
3. **Put the reconciliation in brainstorming instead** — rejected: the contradiction lives in the routing layer's text; future description-level exceptions would hit it again.

No — one rule, two coordinated touches in one file (the rule paragraph and the User Instructions line that contradicted it).

## Existing PRs
- Related PRs: #1715, #1716, #1718 (the stack this completes); none found touching the routing rule's exception semantics.

## Environment tested

| Claude Code (agent under test) | 2.1.169 | Claude Opus | claude-opus-4-8 |
| opencode (agent under test) | 1.16.2 | GPT | openai/gpt-5.5 |

N/A — no harness changes.

## Evaluation

- **Initial prompt:** quorum eval scenarios with scripted naive users (no skills named by the driver).
- **Eval sessions after the change:** the full 5-run verification battery ran on the assembled stack including this change: `cost-checkbox-over-trigger`/claude **pass** (the exception fires through the routing layer — no brainstorming invocation), `triggering-writing-plans`/claude **pass** (the 1% rule still triggers skills that DO apply), `cost-spec-plan-duplication`/claude **pass** (brainstorming still gates real features), `cost-trivial-task-review-fanout`/opencode **pass**, `sdd-rejects-extra-features`/claude **pass** (run IDs in the stack PRs).
- **Before/after:** before, the routing text and the description exception contradicted; after, both eval directions hold simultaneously — skip fires only for the documented exception, triggers fire everywhere else.

## Rigor

The Red Flags table is untouched; the new paragraph closes the "description exempts me" gap the table could not cover, with the doubt-means-invoke backstop preserved. Adversarial findings (D1/D2) and both exploit directions are documented above; the 5-run battery is the post-change evidence.

## Human review

---

## Round 3: staff-review refinements + evidence

**Refinements in this PR's follow-up commit:** the `skill_flow` digraph now routes through a "Skill's own description exempts this request?" diamond (no/any-doubt → invoke) — this stack's own evidence says agents follow flowcharts literally, and the chart previously contradicted the rule; the `<EXTREMELY-IMPORTANT>` block gains a one-line deferral to The Rule (previously it read unconditional, contradicting the rule in the same always-loaded file); writing-skills now distinguishes negative triggering conditions (scope — allowed, and required at the description per this rule) from workflow summaries (still forbidden), so a future editor applying its checklist does not strip the exception and silently regress the cost evals.

**Final-text evidence:** the exception routes correctly where supported — `cost-checkbox-over-trigger` skip: claude 3/3, codex ✓, antigravity ✓ (kimi does not pick up description exceptions; unchanged from baseline). The 1% rule still triggers skills that apply: `triggering-writing-plans`/claude **3/3 pass**; ×codex fail — **byte-for-byte its pre-existing documented signature** (loads sibling skills, skips the mandated one; predates this stack, tracked separately).

**Merge guidance:** merge together with #1718 (see its note).

---

## Round 4: stated-scan mandate

The exception-skip path now requires a visible artifact: one line naming the exception and the empty tripwire scan, written before the first action ("If you did not write the scan line, you did not scan — invoke the skill instead"). The flowchart's exempt edge routes through the scan statement; the red-flags table counters "too trivial to scan". Rationale: the measured boundary leaks were silent non-consultation — externalizing the scan forces it to happen, at a cost of one sentence on the trivial path. Measured effect recorded on #1718 (boundary cells 1/3 → 2/3 with the trivial path intact at 2/2).


arittr:
This change is part of the following stack:

- #1715
    - #1716
        - #1718
            - #1732 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

arittr:
Consolidated into #1718 per independent-mergeability restructure: the routing-rule change and the brainstorming description exception are one behavioral mechanism (the stated-scan protocol spans both files), and shipping them separately left an active contradiction window between merges. #1718 now carries both, plus the full eval evidence.

PR #1734, opened 2026-06-11
fix: use blocked todo status at brainstorming approval gates

chenhaipeng:
## Problem

Automated todo-continuation systems (e.g. OMO's `todo-continuation-enforcer`) inject continuation nudges when they detect incomplete todos (`status: in_progress`). When the brainstorming skill reaches an approval gate and waits for user confirmation, the todo remains `in_progress` — triggering a nudge that the agent interprets as user approval, bypassing the gate.

## Root Cause

Two systems with conflicting assumptions:
- **brainstorming skill**: pauses at approval gates, waiting for user input
- **OMO enforcer**: detects `in_progress` todos and injects system messages to continue work

The injected message is indistinguishable from real user input at the agent level.

## Fix

Add a rule to the brainstorming skill's HARD-GATE and User Review Gate sections: when waiting for user approval, mark the corresponding todo as `blocked` instead of `in_progress`.

OMO's `getIncompleteCount()` already filters out `blocked` status, so no code changes are needed on the OMO side. The `todowrite` tool accepts `blocked` as a valid string status value.

## Changes

- **HARD-GATE section** (L12-16): Added blocked-status rule that applies to ALL approval gates
- **User Review Gate section** (L128-133): Added explicit `blocked` → `in_progress` transition instruction

## Testing

Verified that `todowrite` accepts `status: \"blocked\"` and that OMO's `getIncompleteCount()` in `src/hooks/todo-continuation-enforcer/todo.ts` already excludes `blocked` todos from the incomplete count.

### DIFF
--- a/skills/brainstorming/SKILL.md
+++ b/skills/brainstorming/SKILL.md
@@ -11,6 +11,8 @@ Start by understanding the current project context, then ask questions one at a 
 
 <HARD-GATE>
 Do NOT invoke any implementation skill, write any code, scaffold any project, or take any implementation action until you have presented a design and the user has approved it. This applies to EVERY project regardless of perceived simplicity.
+
+**Todo status at approval gates:** When you are waiting for user approval (design review, spec review, or any user decision), mark the corresponding todo as `blocked` — not `in_progress`. This prevents automated todo-continuation systems from injecting messages that bypass the user approval gate. Switch back to `in_progress` only after the user responds.
 </HARD-GATE>
 
 ## Anti-Pattern: "This Is Too Simple To Need A Design"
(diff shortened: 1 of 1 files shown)


obra:
I think that an 'automated enforcer' that forces continuation any time there is an in progress task is...wrong. This change does not feel like a good fit for superpowers right now.

chenhaipeng:
> I think that an 'automated enforcer' that forces continuation any time there is an in progress task is...wrong. This change does not feel like a good fit for superpowers right now.

I think so too. Right now it's mainly happening with opencode's oh-my-openagnet. The root cause is openagent's mandatory auto-completion mechanism. Overall, there's some minor compatibility issues

PR #1739, opened 2026-06-11
Add quick-ship skill for small/lightweight/fast projects

matthewvilaysack:
## Summary

- Adds a new `quick-ship` skill designed for small, lightweight, one-sitting projects that don't need the full brainstorming → writing-plans → subagent-driven-development ceremony.
- Includes a clear decision gate (4 criteria) so agents know when to use this vs. `brainstorming`.
- Tagged with `small-project`, `lightweight`, `fast`, `quick` in the frontmatter.

## What it does

6-step workflow: restate goal in one sentence → identify blast radius → build → test → commit → ship. Each step has guard rails that redirect to `brainstorming` if the scope turns out to be larger than expected.

## Why it's useful

The existing Superpowers methodology is excellent for non-trivial work, but it's heavier than needed for a one-liner flag addition, a small script, or a quick config tweak. This skill fills that gap without abandoning the quality discipline (test step is non-optional, anti-patterns section covers scope creep).


obra:
Hi,

This isn't a great fit for Superpowers, but you should absolutely publish your own plugin!

PR #1786, opened 2026-06-17
docs: Make SDD Model Selection guidance harness-aware for OpenCode-family agents

### LINKED ISSUE #1776: subagent-driven-development: New model guidance cannot apply on Opencode-based platforms

Possibly related to #1631 but seems distinct enough to warrant a separate issue.

## Environment (required)

| Field | Value |
|-------|-------|
| Superpowers version | 6.0.0 |
| Harness (Claude Code, Cursor, etc.) | Kilo Code (OpenCode fork) |
| Harness version | Latest main |
| Your model + version | Various (different model per subagent) |
| All plugins installed | Only superpowers |
| OS + shell | Linux + zsh |

Interaction of Superpowers + platform capabilities

## What happened?

Version 6.0.0 adds guidance to `subagent-driven-development` informing the agent on how to select a model for the task. However, this workflow is not aligned with the capabilities of OpenCode and its forks. Agents cannot dynamically specify a model to use when invoking a subagent in OpenCode. Instead, the user has a configured list of subagents, which may have different models assigned to them, and the parent agent selects an appropriate subagent for the current task based on the subagent descriptions. As a result, the new guidance on "Model Selection" is more likely to confuse agents on OpenCode harnesses rather than to provide benefit.

mvanhorn:
## Summary
Superpowers v6.0.0 added a 'Model Selection' section to the subagent-driven-development skill that instructs the agent to choose and explicitly specify a model when dispatching each subagent (cheap for mechanical tasks, capable for design). This guidance assumes the harness lets the orchestrator pick a model at dispatch time, which is true for Claude Code's Task tool.

Fixes #1776

### DIFF
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -88,6 +88,21 @@ digraph process {
 - `./spec-reviewer-prompt.md` - Dispatch spec compliance reviewer subagent
 - `./code-quality-reviewer-prompt.md` - Dispatch code quality reviewer subagent
 
+## Model Selection
+
+**Portable principle:** Use the least-powerful model that can reliably handle the subagent's role.
+
+| Role tier | Use for |
+|-----------|---------|
+| Cheap / fast | Mechanical implementation tasks, small docs edits, rote formatting, simple spec checks |
+| Standard | Most implementation tasks and ordinary reviewer passes |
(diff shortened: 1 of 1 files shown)


mvanhorn:
Closing this one. It was built on an older base (the branch sits on Release v4.3.1) and adds a brand-new Model Selection section rather than editing the existing guidance, so it would conflict with or duplicate what's on main. I'll re-approach against current main if the issue still stands. Sorry for the noise.

PR #1787, opened 2026-06-17
fix: Stop in-place mutation of first user message so opencode session titles ignore bootstrap

### LINKED ISSUE #1784: Bootstrap injection pollutes auto-generated session titles (opencode)

## Environment (required)

| Field | Value |
|-------|-------|
| Superpowers version | 5.0.7 |
| Harness (Claude Code, Cursor, etc.) | opencode |
| Harness version | 1.17.7 |
| Your model + version | `claude-haiku-4-5` (title `small_model`); main model `claude-opus-4-8` |
| All plugins installed | superpowers, @franlol/opencode-md-table-formatter |
| OS + shell | macOS 26.5.1 (arm64), zsh |

## What happened?

Auto-generated opencode session titles are derived from the Superpowers bootstrap blob instead of the user's first prompt — producing titles like "Superpowers activated", "Superpowers detected", or other unrelated text pulled from the injected content.

Root cause: the `experimental.chat.messages.transform` hook does `firstUser.parts.unshift({ ...ref, text: bootstrap })` (`superpowers.js` ~line 108-109), mutating the first user message's `parts` array **in place**. In opencode that array is shared in-memory with the session history that the built-in title agent reads (`ensureTitle` → `a.history`). The title model therefore sees the `<EXTREMELY_IMPORTANT>` bootstrap first and titles the session from it. Stored history on disk stays clean — only the title generation call is affected.

## Steps to reproduce
1. Install Superpowers in opencode.
2. Start a new session and send any first prompt (e.g. "fix the failing test").
3. Observe the auto-generated session title.

## Expected behavior

Title reflects the user's first prompt (e.g. "Fix failing test").

## Actual behavior

Title reflects the injected bootstrap (e.g. "Superpowers activated") or other unrelated injected content.

Stored DB history for an affected session is clean (first user part = the real prompt), yet the session title is "Superpowers detected" — confirming the taint comes only from the in-memory `parts` mutation reaching the title call, not from persisted content.

Suggested fix — clone the first message instead of mutating its shared `parts` array:

```js
const idx = output.messages.indexOf(firstUser);
output.messages[idx] = {
  ...firstUser,
  parts: [{ ...firstUser.parts[0], type: 'text', text: bootstrap }, ...firstUser.parts],
};
```

This still delivers the bootstrap to the model but leaves the history object untouched, so titles derive from the real first prompt.

mvanhorn:
## Summary
The opencode plugin's 'experimental.chat.messages.transform' hook injects the Superpowers bootstrap blob by calling firstUser.parts.unshift({...ref, text: bootstrap}) (superpowers.js ~line 135-136), mutating the first user message's parts array in place. In opencode that parts array is shared in memory with the session history the built-in title agent reads (ensureTitle -> a.history), so the title model sees the <EXTREMELY_IMPORTANT> bootstrap first and generates titles like 'Superpowers activated' instead of titles derived from the user's real first prompt.

## Changes
In the transform hook in .opencode/plugins/superpowers.js, stop mutating the shared parts array in place. Instead replace firstUser.parts with a new array (e.g.

Fixes #1784

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -1,7 +1,7 @@
 /**
  * Superpowers plugin for OpenCode.ai
  *
- * Injects superpowers bootstrap context via system prompt transform.
+ * Injects superpowers bootstrap context via message transform.
  * Skills are discovered via OpenCode's native skill tool from symlinked directory.
  */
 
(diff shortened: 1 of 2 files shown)


mvanhorn:
Closing this. The branch is based on Release v4.3.1 and switches the bootstrap injection back to a messages.transform approach, which would conflict with the current superpowers.js and risks reintroducing the #226 agent-reset behavior the newer code fixed. I'd rather not ship that. Will re-do against current main if still needed.

PR #1801, opened 2026-06-18
fix(using-git-worktrees): assign $LOCATION and $BRANCH_NAME in Step 1 before use

### LINKED ISSUE #1797: using-git-worktrees: $LOCATION and $BRANCH_NAME used in Step 1 code block without prior assignment
`skills/using-git-worktrees/SKILL.md` Step 1 includes a "Create the Worktree" code block that uses `$LOCATION` and `$BRANCH_NAME` shell variables without either documenting them as agent-supplied placeholders or showing assignment lines:

```bash
# Determine path based on chosen location
path="$LOCATION/$BRANCH_NAME"

git worktree add "$path" -b "$BRANCH_NAME"
cd "$path"
```

Neither variable is assigned earlier in the SKILL.md, and there is no preceding line of the form `LOCATION=...` / `BRANCH_NAME=...`. An agent reading the workflow and executing the block verbatim ends up running `git worktree add "/" -b ""` (both vars empty), which fails with an empty-branch-name error.

The earlier prose (lines 60–76) does discuss how to pick a *location* and the implicit branch name, but the variables in the code block are presented as if already in scope.

### Suggested fix

Prepend explicit assignment lines so the block is runnable:

```bash
# Use the LOCATION you chose above (e.g., ".worktrees" or a project-specific path)
LOCATION="<chosen-location>"
# Use BRANCH_NAME based on the task (e.g., "feat-add-login")
BRANCH_NAME="<chosen-branch-name>"

# Determine path based on chosen location
path="$LOCATION/$BRANCH_NAME"

git worktree add "$path" -b "$BRANCH_NAME"
cd "$path"
```

Or, if the intent is for the agent to fill the variables before pasting the block, document that explicitly above the code fence (e.g., "**Agent fills:** `LOCATION` = …, `BRANCH_NAME` = …").

specterslient95-lgtm:
Fixes #1797.

`skills/using-git-worktrees/SKILL.md` Step 1 uses `$LOCATION` and `$BRANCH_NAME` in a Bash code block without assigning them, so an agent that executes the block verbatim ends up running `git worktree add "/" -b ""`.

This PR prepends two explicit assignment lines with descriptive placeholder values so the block is runnable and the agent has a clear cue about what to fill in.

### DIFF
--- a/skills/using-git-worktrees/SKILL.md
+++ b/skills/using-git-worktrees/SKILL.md
@@ -90,6 +90,11 @@ git check-ignore -q .worktrees 2>/dev/null || git check-ignore -q worktrees 2>/d
 #### Create the Worktree
 
 ```bash
+# Use the LOCATION you chose above (e.g. ".worktrees" or a project-specific path)
+LOCATION="<chosen-location>"
+# Use BRANCH_NAME based on the task (e.g. "feat-add-login")
+BRANCH_NAME="<chosen-branch-name>"
+
 # Determine path based on chosen location
 path="$LOCATION/$BRANCH_NAME"
 


obra:
Hi @specterslient95-lgtm — I'm Claude, an AI agent (Fable 5, running in Claude Code), posting from Jesse (@obra)'s account at his direction. Jesse had me run a skeptical triage of all 340 open superpowers items — every factual claim tested against the current `dev` tree, then adversarially re-checked by a second, independent agent — and he reviewed the verdicts and directed this batch of closures. This one is closing because the submission itself didn't hold up under those checks — specifics below, with the exact tests run:

Thanks for the detailed report, but this doesn't hold up under testing. I gave a fresh Sonnet agent the Step 1b text with no native worktree tool available (forcing the git fallback), across 3 independent sessions, and asked it to set up a workspace for a real task. In all 3, the agent derived concrete values from the preceding directory-selection guidance and the task itself (e.g. LOCATION=.worktrees, BRANCH_NAME=add-login) and ran the fully-substituted `git worktree add` command — it never emitted the raw `$LOCATION`/`$BRANCH_NAME` verbatim. Skill bash blocks are read and adapted by a reasoning agent, not sourced as a literal script; Step 0 of this same skill relies on the same expectation.

This exact fix was independently submitted as #1820 and already closed for this reason. Even under the "agent executes literally" premise, the fix wouldn't prevent failure — `LOCATION="<chosen-location>"` is still a placeholder, so verbatim execution would just silently succeed with a garbage path/branch instead of failing loudly.

Separately, this PR skips the entire PR template (no model/harness disclosure, no duplicate search, no evaluation) and targets `main` instead of `dev` — either alone is grounds to close without review per CONTRIBUTING.

Closing as works-as-designed / non-reproducing.

---

PR #1826, opened 2026-06-22
fix(tests): resolve shellcheck warnings in Claude Code test helpers

creazyfrog:

| Field | Value |
|-------|-------|
| Your model + version | Claude Sonnet 4.6 |
| Harness + version | Claude Code CLI |
| All plugins installed | superpowers |
| Human partner who reviewed this diff | rohitcse.gec@gmail.com |

Running `scripts/lint-shell.sh --all` (the project's own linter) exits with code 1 due to shellcheck warnings in three test files. Two of these warnings reflect real bugs:

- **SC2320** in `test-subagent-driven-development-integration.sh` line 170: `$?` in the failure handler referred to the preceding `echo ""` (exit 0), not to the failed `claude` command. The error message always printed `exit code: 0`, making test failures hard to diagnose.
- **SC2155** in `test-helpers.sh`: five `local var=$(cmd)` patterns where the `local` builtin's exit code (always 0) masks a failure from `mktemp` or `grep`. Under `set -e` this silently continues with an empty variable instead of halting.
- **SC2064** in `test-subagent-driven-development-integration.sh` line 40: `trap "..." EXIT` with double-quotes expands `$TEST_PROJECT` immediately. A wrapper function makes the intent explicit.
- **SC2088** in `test-worktree-path-policy.sh`: four grep patterns intentionally check for the literal tilde string `~/.config/...` in Markdown skill files — not the expanded `$HOME` path. Added `# shellcheck disable` comments to document this.

## What does this PR change?

- `tests/claude-code/test-helpers.sh`: declare five `local` variables on their own line, then assign separately
- `tests/claude-code/test-subagent-driven-development-integration.sh`: capture `$?` into `exit_code` before `echo`, replace inline trap with `_cleanup()` wrapper
- `tests/claude-code/test-worktree-path-policy.sh`: add `# shellcheck disable=SC2088` with explanation on four lines where `~` must remain unexpanded

After this PR, `scripts/lint-shell.sh --all` exits 0 with no warnings.

Adding `# shellcheck disable=SC2155` comments inline. Chose separate-declaration instead because it is the idiomatic fix, makes the code correct under `set -e`, and produces no extra noise.

No — all three files are test helpers in the same `tests/claude-code/` directory, all changes are shellcheck correctness fixes, and the goal is a single clean `lint-shell.sh --all` run.

## Existing PRs
- Related PRs: none found

## Environment tested

| Claude Code CLI | latest | Claude | Sonnet 4.6 |

## Evaluation

`scripts/lint-shell.sh --all` before: exits 1 with 11 warnings across 3 files.
`scripts/lint-shell.sh --all` after: exits 0, `Linting 39 shell files`, no output.

## Rigor

- [x] I did not modify carefully-tuned content

## Human review

### DIFF
--- a/tests/claude-code/test-helpers.sh
+++ b/tests/claude-code/test-helpers.sh
@@ -7,7 +7,8 @@ run_claude() {
     local prompt="$1"
     local timeout="${2:-60}"
     local allowed_tools="${3:-}"
-    local output_file=$(mktemp)
+    local output_file
+    output_file=$(mktemp)
 
     # Build command as an argv array so timeout wraps claude directly.
     local cmd=(claude -p "$prompt")
@@ -74,7 +75,8 @@ assert_count() {
     local expected="$3"
     local test_name="${4:-test}"
 
-    local actual=$(echo "$output" | grep -c "$pattern" || echo "0")
+    local actual
+    actual=$(echo "$output" | grep -c "$pattern" || echo "0")
 
     if [ "$actual" -eq "$expected" ]; then
         echo "  [PASS] $test_name (found $actual instances)"
@@ -98,8 +100,9 @@ assert_order() {
     local test_name="${4:-test}"
 
     # Get line numbers where patterns appear
-    local line_a=$(echo "$output" | grep -n "$pattern_a" | head -1 | cut -d: -f1)
-    local line_b=$(echo "$output" | grep -n "$pattern_b" | head -1 | cut -d: -f1)
+    local line_a line_b
+    line_a=$(echo "$output" | grep -n "$pattern_a" | head -1 | cut -d: -f1)
+    line_b=$(echo "$output" | grep -n "$pattern_b" | head -1 | cut -d: -f1)
 
     if [ -z "$line_a" ]; then
         echo "  [FAIL] $test_name: pattern A not found: $pattern_a"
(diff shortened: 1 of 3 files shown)


jnMetaCode:
Since this PR is already touching the `assert_count` line, one small thing that survives the split:

```bash
actual=$(echo "$output" | grep -c "$pattern" || echo "0")
```

`grep -c` prints `0` **and** exits 1 when there are no matches, so `|| echo "0"` fires too and `actual` becomes the two-line string `"0\n0"`. The `-eq` below then errors instead of comparing:

```
$ assert_count "hello world" "forbidden" 0 "expect zero"
test-helpers.sh: line 81: [: 0
0: integer expression expected
  [FAIL] expect zero
```

**Scoping this honestly:** `assert_count` has no callers in the tree today, so this is latent rather than breaking anything now. The same idiom at `task_count` / `todo_count` in `test-subagent-driven-development-integration.sh` *is* live, but those use `-ge` and the zero case is supposed to fail anyway — the verdict stays correct there, only the message garbles (`Only 0\n0 subagent(s) dispatched`) plus a stray stderr line.

So: not urgent, just a trap for whoever first calls `assert_count` with an expected count of 0.

If you want it folded in while you're here:

```bash
local actual
actual=$(echo "$output" | grep -c "$pattern"; true)
actual=${actual:-0}
```

`; true` swallows only the exit status without appending a second line. Entirely your call whether it belongs in a shellcheck-scoped PR — happy to send it separately if you'd rather keep this one focused.

PR #1852, opened 2026-06-26
subagent-driven-development: concrete opencode-go model roster

kthorn:
## What

Replaces the abstract tiering in `subagent-driven-development`'s **Model Selection** section ("use a fast/cheap model" / "standard model" / "most capable available") with a **concrete opencode-go roster keyed to coding strength + cost**, so a controller dispatching a subagent has real model IDs to reach for instead of guessing.

## Why

The prior section gave the *rule* (least powerful model that can handle each role; turn count beats token price; always specify the model explicitly) but **no roster** — "use the standard model" is an instruction a controller cannot execute, because no model is named "standard." Controllers are left to anchor on whatever model ID they happen to remember and then fall off a cliff for the other tiers (see the PR description on the paired kthorn-skills change for a worked example of this failure).

Concrete model IDs turn the abstract rubric into a lookup table.

## The roster

Coding strength (higher is better, per the project's benchmarks); `req/5h` is the opencode-go gateway's usage-budget proxy (more = cheaper/more abundant).

| Tier | Role | Model | Strength | req/5h |
|------|------|-------|----------|--------|
| Mechanical | Implementer (transcription + testing, complete code in plan, 1–2 files) | `opencode-go/deepseek-v4-flash` + `thinking: high` | 56 (max eff) | 31,650 |
| Integration / judgment | Implementer (multi-file, prose spec) | `opencode-go/mimo-v2.5-pro` → fallback `opencode-go/kimi-k2.7-code` | 60 / 61 | 3,250 / 1,350 |
| Architecture / hard | Implementer (design judgment, broad codebase) | `opencode-go/glm-5.2` + `thinking: high` | 69 (max eff) | 880 |
| Per-task review | Task reviewer (spec + quality gate, every task) | `opencode-go/mimo-v2.5-pro` | 60 | 3,250 |
| Final whole-branch review | Final reviewer (one per branch) | `opencode-go/glm-5.2` + `thinking: high` | 69 (max eff) | 880 |

**Effort mechanism:** `deepseek-v4-flash`, `glm-5.2`, `deepseek-v4-pro`, and `kimi-k2.7-code` are all `reasoning: true` with `thinkingFormat: deepseek`, so a `thinking: high` level on the dispatch (or the agent's `thinking` field) routes to max effort — verified empirically against the gateway.

## What's preserved

The two load-bearing rules from the prior version are kept verbatim in spirit:
- **Always specify the model explicitly** — an omitted model inherits the session's expensive default.
- **Turn count beats token price** — the floor for prose-spec implementers and reviewers is mid-tier (MiMo Pro), not Flash; Flash only wins when the work is genuinely transcription-plus-testing.

## What's project-local (not in this PR)

The roster is opencode-go-specific. Projects on other gateways should declare their own roster in their AGENTS.md; the skill's note points there for projects that need a different roster. The paired project-local edits (in repos outside this one) move the per-task reviewer off `glm-5.2` (scarcest+strongest, now reserved for architecture and final review) onto `mimo-v2.5-pro`.

## Notes

- `kimi-k2.6`, `glm-5.1`, `qwen3.6-plus`, `qwen3.7-max`, `qwen3.7-plus`, `minimax-m2.7`, `minimax-m2.5`, and non-pro `mimo-v2.5` were benchmarked weaker than `deepseek-v4-flash` and excluded from the usable roster.
- `minimax-m3` is very-narrowly dominated by `mimo-v2.5-pro` and dropped for now; can be reinstated if mid-tier availability becomes a concern.

### DIFF
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -98,36 +98,33 @@ conflicts that only emerge from implementation.
 
 ## Model Selection
 
-Use the least powerful model that can handle each role to conserve cost and increase speed.
+Use the least powerful model that can handle each role to conserve cost and increase speed. The roster below is concrete to the **opencode-go** gateway (`https://opencode.ai/zen/go/v1`, key in `~/.pi/agent/auth.json`); coding-strength scores are higher-is-better, `req/5h` is the gateway's usage-budget proxy (more = cheaper/more abundant). Effort is controlled by the `thinking` level on the dispatch or the agent's `thinking` field; `deepseek-v4-flash`, `glm-5.2`, `deepseek-v4-pro`, and `kimi-k2.7-code` are all `reasoning: true` with `thinkingFormat: deepseek` and honor `thinking: high` (max effort). Projects may declare additional providers/models in their AGENTS.md — match the model's strength to the task's difficulty.
 
-**Mechanical implementation tasks** (isolated functions, clear specs, 1-2 files): use a fast, cheap model. Most implementation tasks are mechanical when the plan is well-specified.
+| Tier | Role | Model | Strength | req/5h | When |
+|------|------|-------|----------|--------|------|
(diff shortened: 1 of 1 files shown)


kthorn:
Closing — this is opencode-go-gateway-specific roster guidance; keeping it on the fork (kthorn/superpowers) rather than upstreaming to obra/superpowers.

PR #1864, opened 2026-06-29
fix(hooks): discover per-user Git Bash and skip WSL stub on Windows

### LINKED ISSUE #1863: run-hook.cmd bash discovery misses per-user Git install (%LOCALAPPDATA%\Programs\Git), falls through to WSL bash stub
**Env:** Windows 11, Claude Code 2.1.195, superpowers 6.0.3, Git for Windows installed **per-user** at `%LOCALAPPDATA%\Programs\Git`.

After fixing the PowerShell call-operator problem (#1751), the SessionStart hook still fails to inject context. `hooks/run-hook.cmd` probes only:

```
C:\Program Files\Git\bin\bash.exe
C:\Program Files (x86)\Git\bin\bash.exe
```

then falls back to `where bash`. On a per-user Git install neither standard path exists, and `where bash` resolves first to the Windows `WindowsApps\bash.exe` **WSL launcher stub**. With no WSL distro installed it errors and injects nothing:

```
WSL (8 - Relay) ERROR: CreateProcessCommon:800: execvpe(/bin/bash) failed: No such file or directory
```

The working Git Bash sits unprobed at `%LOCALAPPDATA%\Programs\Git\bin\bash.exe`. Verified: invoking `hooks/session-start` directly under that bash produces the correct `SessionStart` context JSON and exits 0 — so the script is fine; only the launch/discovery path is broken.

This is distinct from #912 (Cursor-specific, about routing through the wrapper at all, closed by #1054); here the wrapper *is* used but its bash discovery is incomplete.

## Suggested fix

In `run-hook.cmd`, also probe the per-user Git install path, and avoid selecting a `bash.exe` that resolves under `WindowsApps` (the WSL stub) — prefer a real Git Bash:

```
%LOCALAPPDATA%\Programs\Git\bin\bash.exe
%LOCALAPPDATA%\Programs\Git\usr\bin\bash.exe
```

citizen204:
## Summary

- `hooks/run-hook.cmd`: adds `%LOCALAPPDATA%\Programs\Git\bin\bash.exe` to the bash discovery probe list, covering per-user Git for Windows installs that are invisible to the existing system-path checks.
- Replaces the plain `where bash | bash` fallback with a `for /f` loop that iterates every candidate returned by `where bash` and skips any path under `WindowsApps\` — the WSL launcher stub that errors with "No such file or directory" when no WSL distro is installed.

The two changes together ensure a user with a per-user Git install and no WSL distro gets a working `SessionStart` hook instead of a silent failure or a confusing relay error.

Fixes #1863

## Changes

- `hooks/run-hook.cmd`: add per-user Git path probe; filter WSL stub in PATH fallback

### DIFF
--- a/hooks/run-hook.cmd
+++ b/hooks/run-hook.cmd
@@ -27,11 +27,19 @@ if exist "C:\Program Files (x86)\Git\bin\bash.exe" (
     exit /b %ERRORLEVEL%
 )
 
-REM Try bash on PATH (e.g. user-installed Git Bash, MSYS2, Cygwin)
-where bash >nul 2>nul
-if %ERRORLEVEL% equ 0 (
-    bash "%HOOK_DIR%%~1" %2 %3 %4 %5 %6 %7 %8 %9
+REM Try per-user Git for Windows install (%LOCALAPPDATA%\Programs\Git)
+if exist "%LOCALAPPDATA%\Programs\Git\bin\bash.exe" (
+    "%LOCALAPPDATA%\Programs\Git\bin\bash.exe" "%HOOK_DIR%%~1" %2 %3 %4 %5 %6 %7 %8 %9
     exit /b %ERRORLEVEL%
+)
+
+REM Try bash on PATH, skipping WindowsApps\bash.exe (the WSL stub)
+for /f "usebackq delims=" %%B in (`where bash 2^>nul`) do (
+    echo(%%B | findstr /i "WindowsApps" >nul 2>nul
+    if errorlevel 1 (
+        "%%B" "%HOOK_DIR%%~1" %2 %3 %4 %5 %6 %7 %8 %9
+        exit /b %ERRORLEVEL%
+    )
 )
 
 REM No bash found - exit silently rather than error


hooks/run-hook.cmd
```diff
@@ -27,13 +27,21 @@ if exist "C:\Program Files (x86)\Git\bin\bash.exe" (
...
-if %ERRORLEVEL% equ 0 (
-    bash "%HOOK_DIR%%~1" %2 %3 %4 %5 %6 %7 %8 %9
+REM Try per-user Git for Windows install (%LOCALAPPDATA%\Programs\Git)
+if exist "%LOCALAPPDATA%\Programs\Git\bin\bash.exe" (
+    "%LOCALAPPDATA%\Programs\Git\bin\bash.exe" "%HOOK_DIR%%~1" %2 %3 %4 %5 %6 %7 %8 %9
     exit /b %ERRORLEVEL%
 )
 
+REM Try bash on PATH, skipping WindowsApps\bash.exe (the WSL stub)
+for /f "usebackq delims=" %%B in (`where bash 2^>nul`) do (
```
tspearconquest, line 37: This shouldn't need a `where` at all.

`for %%I in (bash.exe) do (` should be enough to pick up any %PATH% bash. It works for me anyway and shaves some milliseconds from startup (against the Copilot 10s timeout in my case) by not forking a `where` process.

PR #1882, opened 2026-07-01
fix: narrow using-superpowers trigger scope

kht33668944-tech:
## Summary
- Narrow using-superpowers discovery so it applies to concrete work where a specialized skill may apply.
- Replace the broad 1% trigger with clear-match or explicit-skill-request guidance.
- Add a PowerShell regression check for the trigger wording and update harness porting docs.

## Validation
- powershell -NoProfile -ExecutionPolicy Bypass -File tests\\using-superpowers\\test-trigger-scope.ps1
- git diff --check

## Note
Direct push to obra/superpowers main was attempted but rejected with 403 for this GitHub account, so this PR is opened from a fork.

### DIFF
--- a/docs/porting-to-a-new-harness.md
+++ b/docs/porting-to-a-new-harness.md
@@ -703,8 +703,10 @@ Then:
     unrecognized.
   - **Otherwise lean on the installed `using-superpowers` skill itself.** If the
     harness surfaces each installed skill's name + description at session start,
-    the `using-superpowers` description ("Use when starting any conversation…")
-    can prompt the model to load it — installing the skill *is* the bootstrap.
+    the `using-superpowers` description ("Use when a user asks to implement,
+    debug, review, plan, research, automate, or modify files and a specialized
+    skill may apply") can prompt the model to load it — installing the skill *is*
+    the bootstrap.
     Softer (no guaranteed wrapper; it carries triggering but not the tool mapping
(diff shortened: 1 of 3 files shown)


obra:
Hi @kht33668944-tech — I'm Claude, an AI agent (Fable 5, running in Claude Code), posting from Jesse (@obra)'s account at his direction. Jesse had me run a skeptical triage of all 340 open superpowers items — every factual claim tested against the current `dev` tree, then adversarially re-checked by a second, independent agent — and he reviewed the verdicts and directed this batch of closures. This one is closing because the submission itself didn't hold up under those checks — specifics below, with the exact tests run:

Thanks for the interest, but this isn't reviewable as submitted. It targets `main` instead of `dev`, and the PR template is skipped entirely — including the mandatory model/harness/plugin disclosure (the `codex/` branch name suggests Codex authorship, which needs to be stated, not omitted). More importantly, this deletes the `using-superpowers` Red Flags table and "1% chance" language, added specifically to harden the skill against rationalization (commits f6ee98a, 2d7408d) — per CLAUDE.md that tuned content needs real adversarial eval evidence to change, not a script that just checks the file contains the exact strings this same diff inserts. If you've hit a concrete session where the current wording caused a bad over-trigger, describe it, retarget `dev`, complete the template, and bring before/after eval results and we're glad to look again.

---

PR #1890, opened 2026-07-02
fix(brainstorming): triggerable description, resolvable paths, orphan prompt, exclusions

Deepsea1:
# fix(brainstorming): triggerable description, resolvable paths, orphan prompt, exclusions

## Problem

Four structural defects in `skills/brainstorming/`, found during a structured skill-design audit:

1. **The description is the skill's only activation signal, and the current one over-fires with zero exclusions.** `"You MUST use this before any creative work..."` is second-person policy aimed at the agent *after* loading, not a triggering condition the harness can match *before* loading. As written it claims every creative task unconditionally — including bug fixes, plan execution, and research questions that other superpowers skills (systematic-debugging, executing-plans) explicitly own, and including work that already has an approved design.

2. **Plugin-root-relative paths don't resolve from normal working directories.** `SKILL.md` points to `skills/brainstorming/visual-companion.md`, and `visual-companion.md` invokes `scripts/start-server.sh`, `scripts/stop-server.sh`, and references `scripts/frame-template.html` / `scripts/helper.js`. These paths are relative to the plugin root, but the agent's cwd during a brainstorming session is the user's project — so following them as written fails (or worse, silently reads/executes the wrong file if the user's repo happens to have a `scripts/` dir).

3. **`spec-document-reviewer-prompt.md` is orphaned.** When the subagent spec-review loop was replaced by the inline Spec Self-Review, nothing in SKILL.md referenced the prompt file anymore — it ships in the skill directory with no way to be discovered.

4. **No "When NOT to Use" guidance in the body**, compounding the over-firing from (1).

## Fix

- **Description rewritten** as third-person triggering conditions with the same trigger breadth plus exclusions: *"Use when starting creative work — new features, components, added functionality, or behavior changes — that lacks an approved design or spec. Explores user intent, requirements and design before implementation. Not for bug fixes (use systematic-debugging), executing an existing plan (use executing-plans), or pure research questions."*
- **`{baseDir}` used for all skill-internal paths** so they resolve regardless of cwd: `{baseDir}/visual-companion.md` in SKILL.md; `{baseDir}/scripts/start-server.sh`, `{baseDir}/scripts/stop-server.sh`, `{baseDir}/scripts/frame-template.html`, `{baseDir}/scripts/helper.js` in visual-companion.md.
- **Orphan resolved by re-linking, not deleting:** the Spec Self-Review section now offers an optional deep-review path — dispatch a fresh-eyes subagent using the template in `{baseDir}/spec-document-reviewer-prompt.md`. Kept rather than deleted because `skills/writing-plans/` still ships its sibling `plan-document-reviewer-prompt.md`, so retaining the file is the consistent choice.
- **Added a "When NOT to Use" section** to the SKILL.md body naming the alternatives (systematic-debugging for bugs, executing-plans for approved plans, plain answers for research questions).

## Scope

Two files, docs-only, no behavior/script changes:

```
 skills/brainstorming/SKILL.md            | 14 ++++++++++++--
 skills/brainstorming/visual-companion.md | 16 ++++++++--------
 2 files changed, 20 insertions(+), 10 deletions(-)
```

These findings came out of a structured skill-design audit (description-as-trigger, path resolvability, orphan-file, and exclusion checks applied across a skill library).

### DIFF
--- a/skills/brainstorming/SKILL.md
+++ b/skills/brainstorming/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: brainstorming
-description: "You MUST use this before any creative work - creating features, building components, adding functionality, or modifying behavior. Explores user intent, requirements and design before implementation."
+description: "Use when starting creative work — new features, components, added functionality, or behavior changes — that lacks an approved design or spec. Explores user intent, requirements and design before implementation. Not for bug fixes (use systematic-debugging), executing an existing plan (use executing-plans), or pure research questions."
 ---
 
 # Brainstorming Ideas Into Designs
@@ -16,6 +16,14 @@ Do NOT invoke any implementation skill, write any code, scaffold any project, or
 ## Anti-Pattern: "This Is Too Simple To Need A Design"
 
 Every project goes through this process. A todo list, a single-function utility, a config change — all of them. "Simple" projects are where unexamined assumptions cause the most wasted work. The design can be short (a few sentences for truly simple projects), but you MUST present it and get approval.
+
+## When NOT to Use
+
+Brainstorming is for creative work that lacks an approved design. Skip it when:
+
+- **Fixing a bug or investigating a failure** — use the systematic-debugging skill instead
+- **Executing an already-approved plan or spec** — use the executing-plans skill instead
+- **Answering a pure research or explanation question** — no design is being produced, so just answer it
 
 ## Checklist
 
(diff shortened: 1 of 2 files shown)


obra:
Closing. This is the output of a skill-design audit rather than a problem someone hit in a session, and each hunk has a concrete defect: the description rewrite narrows a deliberately broad trigger (the exact class of change our guidelines say requires extensive eval evidence — none is provided); `{baseDir}` is not a substitution Claude Code performs in skill markdown, so it would appear literally and break the very invocations it claims to fix; and re-linking spec-document-reviewer-prompt.md reverses a deliberate removal (e6221a4, made with eval data). It also targets main instead of dev. Per CLAUDE.md, speculative audit findings without a motivating session aren't accepted.

PR #1891, opened 2026-07-02
brainstorming: require design-doc claims to be verified or labeled assumptions

eternalcloud:
## What

Adds a **"Ground every claim in evidence"** contract to the brainstorming skill's "Working in existing codebases" guidance, plus a fifth **Evidence check** item in the Spec Self-Review (and the matching word in checklist item 7).

The contract: every claim a design doc makes about the *current* system must be either **verified** (code read this session, file cited) or an explicitly **labeled assumption** ("Assumed — confirm during implementation") — never plausible inference stated as fact. It also names the five places unverified claims predictably cluster: negative space (what the system does NOT do), integration seams (existing readers/writers of a resource being extended, especially partial writers and replace-vs-merge), runtime wiring (routes/commands inferred from file names), "follows the existing pattern" (without confirming the pattern covers the new case), and forgiving helpers reused to back strict guarantees.

## Why (RED baseline — real production failure)

A design spec written under this skill accumulated 5 blocking defects caught only by a 6-round external cross-model review, all with one mechanism: current-behavior claims written from plausible inference rather than evidence. Concretely: a route asserted from a filename that didn't exist; "existing validation rejects malformed payloads" when it accepted empty selectors; new fields designed onto a resource whose partial-payload client save would have silently clobbered them; "follow the existing migration pattern" where the pattern (create-if-not-exists) couldn't add columns; and a forgiving lookup helper proposed to back a strict verification contract. Each landed in one of the five cluster spots the new guidance names.

## Testing (per writing-skills TDD)

- Failure classified as wrong-shaped output (compliant spec, missing required property), so the fix is a positive contract + structural checklist slot, not prohibitions — per "Match the Form to the Failure".
- Micro-tested the wording: same spec-writing task on the same repo, 1 no-guidance control + 2 guided reps. Both guided reps converged on cite-or-label behavior and explicitly flagged the one unverifiable claim as "Assumed — confirm during implementation"; the control asserted that same negative-space claim (schema uniqueness) as unverified fact. No degradation in the guided outputs.

### DIFF
--- a/skills/brainstorming/SKILL.md
+++ b/skills/brainstorming/SKILL.md
@@ -27,7 +27,7 @@ You MUST create a task for each of these items and complete them in order:
 4. **Propose 2-3 approaches** — with trade-offs and your recommendation
 5. **Present design** — in sections scaled to their complexity, get user approval after each section
 6. **Write design doc** — save to `docs/superpowers/specs/YYYY-MM-DD-<topic>-design.md` and commit
-7. **Spec self-review** — quick inline check for placeholders, contradictions, ambiguity, scope (see below)
+7. **Spec self-review** — quick inline check for placeholders, contradictions, ambiguity, scope, evidence (see below)
 8. **User reviews written spec** — ask user to review the spec file before proceeding
 9. **Transition to implementation** — invoke writing-plans skill to create implementation plan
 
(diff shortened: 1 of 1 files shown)


obra:
Carrying over an idea from #2066 (now closed as speculative) that's adjacent to this PR's verified-claims requirement: beyond verifying claims the design *makes*, a repo-grounded self-review could also check for what the design *misses* — naming collisions with existing code, blast radius of touched interfaces, and whether the stated baseline behavior matches what the repo actually does. If this PR's eval work is being extended, those three checks may be worth including in the same pass rather than as a follow-up.

PR #1939, opened 2026-07-06
fix: use relative Claude hook command

### LINKED ISSUE #1918: Windows: SessionStart:startup hook fails when plugin path contains a parenthesis '(' (regression variant of #51, repro on 6.1.1)
## Summary

`SessionStart:startup` hook fails on Windows when the user profile path contains a `(` (parenthesis):

```
⎿ SessionStart:startup hook error: 'C:\Users\UserName' is not recognized as an internal or external command, operable program or batch file.
```

The error string is truncated exactly at the `(` — cmd.exe treats `(` as a special grouping character, so the path `C:\Users\UserName(External)\...\run-hook.cmd` is split and `C:\Users\UserName` is parsed as the command.

Non-blocking, so the session still starts — but superpowers' SessionStart context injection is silently skipped every session.

## Environment

- Superpowers: **6.1.1**
- OS: Windows 11 Pro 10.0.26100
- Home dir: `C:\Users\UserName(External)\` (contains a parenthesis)

## Relation to #51

#51 covered a **space** in the path and was closed as fixed by the v3.6+ polyglot `run-hook.cmd` wrapper. This is a **distinct trigger** — the parenthesis — and it still reproduces on 6.1.1. A path with no space but with `(` breaks.

## Root cause

`hooks/hooks.json`:

```json
{
}
```

Although `${CLAUDE_PLUGIN_ROOT}` is quoted in the JSON, once it is expanded and handed to cmd.exe the `(` in the path is still interpreted as a special character, breaking the command before `run-hook.cmd` is ever reached. The polyglot wrapper's internal quoting is correct — the failure is at the outer invocation layer, before the wrapper runs.

## Repro

1. Install superpowers under a Windows user whose profile path contains `(` (e.g. `C:\Users\UserName(External)\`).
2. Start Claude Code.
3. Observe the `SessionStart:startup` hook error truncated at the `(`.

## Suggested fix

Make the hook invocation robust to cmd.exe metacharacters (`(`, `)`, `&`, `^`, spaces) in `CLAUDE_PLUGIN_ROOT`. Options: escape/quote-harden the expanded path before it reaches cmd.exe, or invoke the wrapper in a way that isn't parsed by cmd.exe's command tokenizer.

luochen211:
## Summary
- switch the Claude Code SessionStart hook command to `./hooks/run-hook.cmd session-start`
- add regression coverage that the Claude hook command remains relative and dispatches from the plugin root
- update the Windows hook docs and harness porting guidance

## Why
When `${CLAUDE_PLUGIN_ROOT}` is expanded into the outer hook command on Windows, `cmd.exe` can parse metacharacters in the install path before `run-hook.cmd` starts. A user profile path containing `(` breaks the command at the outer invocation layer, so SessionStart context injection is skipped.

Using the same relative dispatcher style as the Cursor hook keeps the expanded install path out of the outer command. The wrapper still derives its own directory and runs the extensionless hook script.

Fixes #1918

## Testing
- `bash tests/hooks/test-session-start.sh`
- `bash tests/codex/test-marketplace-manifest.sh`
- `bash tests/kimi/test-plugin-manifest.sh`
- `bash tests/shell-lint/test-lint-shell.sh`
- `bash -n hooks/session-start hooks/run-hook.cmd tests/hooks/test-session-start.sh`
- `node -e 'JSON.parse(require("fs").readFileSync("hooks/hooks.json", "utf8")); JSON.parse(require("fs").readFileSync("hooks/hooks-cursor.json", "utf8"));'`\n- `git diff --check`

### DIFF
--- a/docs/porting-to-a-new-harness.md
+++ b/docs/porting-to-a-new-harness.md
@@ -392,12 +392,14 @@ Claude Code shape is universal. Compare `hooks/hooks.json` and
 Claude Code uses. Match your `hooks-<harness>.json` to whichever existing file is
 closest, not to a single canonical template.
 
-The hook **command string references a harness-provided plugin-root variable**,
-and its name differs per harness: `hooks.json` uses `${CLAUDE_PLUGIN_ROOT}`,
-`hooks-cursor.json` uses a relative path. Use
-whatever your harness exports. (The `session-start` script re-derives the root
-itself via `dirname`, so the script body doesn't depend on this — but the
-command in the manifest does.)
+The hook **command string is interpreted by the harness's shell**. Prefer a
+relative command when the harness resolves it from the plugin root, as
+`hooks.json` and `hooks-cursor.json` do with `./hooks/run-hook.cmd`; this keeps
+Windows `cmd.exe` from parsing metacharacters in the expanded install path.
+When a harness requires an explicit plugin-root variable, use whatever that
+harness exports. (The `session-start` script re-derives the root itself via
+`dirname`, so the script body doesn't depend on this — but the command in the
+manifest does.)
 
 **Discovering the harness's contract.** The three facts above — env var, JSON
 field/nesting, matcher strings — are the harness's contract, not Superpowers',
--- a/docs/windows/polyglot-hooks.md
+++ b/docs/windows/polyglot-hooks.md
@@ -41,7 +41,7 @@ hooks/
         "hooks": [
           {
             "type": "command",
-            "command": "\"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start",
+            "command": "./hooks/run-hook.cmd session-start",
             "async": false
           }
         ]
@@ -51,7 +51,9 @@ hooks/
 }
 ```
 
-The path is quoted because `${CLAUDE_PLUGIN_ROOT}` may contain spaces.
+The command is relative to the plugin root so the outer shell never sees the
+expanded plugin install path. On Windows, that avoids `cmd.exe` metacharacter
+parsing for user profile paths that contain characters such as parentheses.
 
 ## How `run-hook.cmd` Works at a High Level
 
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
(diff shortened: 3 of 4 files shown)


obra:
Hi @luochen211 — I'm Claude, an AI agent (Fable 5, running in Claude Code), posting from Jesse (@obra)'s account at his direction. Jesse had me run a skeptical triage of all 340 open superpowers items — every factual claim tested against the current `dev` tree, then adversarially re-checked by a second, independent agent — and he reviewed the verdicts and directed this batch of closures. This one is closing because the submission itself didn't hold up under those checks — specifics below, with the exact tests run:

Thanks for digging into #1918 — the cmd.exe root cause is right: once a special character like `(` appears inside the quoted path, cmd's /C parser falls back to stripping only the first/last quote on the whole line, leaving `(` to be parsed as a grouping operator. That's a real, unaddressed bug.

This specific fix can't be merged, though. Claude Code hook commands run with cwd at the session's current/project directory, not the plugin's install directory — the official docs say so directly ("do not rely on relative paths from cwd... use ${CLAUDE_PROJECT_DIR} or ${CLAUDE_PLUGIN_ROOT}"). I reproduced it: `./hooks/run-hook.cmd session-start` fails with "No such file or directory" from any directory other than the plugin's own checkout, while the current `${CLAUDE_PLUGIN_ROOT}`-anchored command succeeds from that same spot. Merging this would silently disable SessionStart context injection for every user on every OS — not just the Windows-parenthesis case it targets. The new test doesn't catch this because it `cd`s into the plugin root before invoking the relative command, which isn't how Claude Code actually runs it.

Please also complete the PR template — model/harness/plugin disclosure, the human-review checkbox, and an evaluation section are all required and currently missing. A fix that keeps `${CLAUDE_PLUGIN_ROOT}` but switches to exec form (args) instead of a quoted shell-form string looks like the right direction — worth trying next.

---

luochen211:
Thanks for the detailed triage. That makes sense.

I agree this should not rely on the hook cwd being the plugin root. I’ll rework the fix to preserve the `${CLAUDE_PLUGIN_ROOT}`-anchored path while avoiding the cmd.exe shell-form quoting issue, likely by moving the hook command toward an exec/args form if the plugin manifest supports it.

I’ll also add a regression test that invokes the hook from a non-plugin cwd, so it matches how Claude Code actually runs hooks, and I’ll complete the required PR template sections before submitting the next version.

PR #1944, opened 2026-07-06
[codex] Add agent workflow orchestration metadata

tanurus:
﻿## Summary

Adds Phase 1 agent workflow orchestration guidance to Superpowers:

- requires task metadata in implementation plans
- adds controller validation and deterministic routing rules
- adds write-scope enforcement guidance
- keeps Claude/Gemini as external reviewers only
- adds a PowerShell metadata validator and regression test

## Validation

- `tests/phase1-orchestration.test.ps1`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,11 +195,13 @@ The Pi package loads the Superpowers skills and a small extension that injects t
 
 4. **subagent-driven-development** or **executing-plans** - Activates with plan. Dispatches fresh subagent per task with two-stage review (spec compliance, then code quality), or executes in batches with human checkpoints.
 
-5. **test-driven-development** - Activates during implementation. Enforces RED-GREEN-REFACTOR: write failing test, watch it fail, write minimal code, watch it pass, commit. Deletes code written before tests.
+5. **cross-model-review** or **external-model-workers** - Activates when you want opt-in external Claude/Gemini review or task execution. External review stays read-only; external workers run in isolated worktrees with write-scope enforcement.
 
-6. **requesting-code-review** - Activates between tasks. Reviews against plan, reports issues by severity. Critical issues block progress.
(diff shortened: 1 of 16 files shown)


obra:
Hi @tanurus — I'm Claude, an AI agent (Fable 5, running in Claude Code), posting from Jesse (@obra)'s account at his direction. Jesse had me run a skeptical triage of all 340 open superpowers items — every factual claim tested against the current `dev` tree, then adversarially re-checked by a second, independent agent — and he reviewed the verdicts and directed this batch of closures. This one is closing as a **venue call** — a decision about where the work belongs, not a judgment of its quality:

Thanks for the work, but this can't land in core, mostly for a venue reason.

The heart of the PR — the new `cross-model-review` and `external-model-workers` skills — wraps the external `claude` and `gemini` CLIs as subprocess workers (`run-worker-with-model.ps1` pipes to `claude -p ...` and throws "gemini CLI is not installed or not on PATH"). Superpowers is zero-dependency by design; anything requiring an external tool belongs in its own plugin — this is CLAUDE.md's "Third-party dependencies" case. It also re-adds `using-superpowers/references/gemini-tools.md`, which we removed in 711d895 after Google EOLed the Gemini CLI on 2026-06-18, so part of this reverts a deliberate decision.

It also isn't reviewable as-is: it targets `main` (should be `dev`), it's a draft with no template (no model/harness/plugin disclosure, no problem statement, no eval evidence), it's PowerShell in a bash/node/python repo, and the new "Deterministic Routing" text in subagent-driven-development permits parallel implementer dispatch while that same file's Red Flags still say "Never: Dispatch multiple implementation subagents in parallel."

The metadata-driven, write-scope-aware routing idea is real and overlaps open work — #1117, #970, #956. If you want to pursue it, split that piece out, target `dev`, complete the template with disclosure, and bring eval evidence per the skill-change policy; ship the external-CLI workers as a standalone plugin.

---
*If you think this call is wrong, reply here — Jesse reads these, and closures can be revisited.*

Issue #1780, opened 2026-06-17
Use a working-tree SDD directory (.superpowers/sdd) to avoid .git write approvals and improve SDD UX


MadMaximusJB:
## What problem does this solve?

When running Subagent-Driven Development (SDD), the skill creates and manages short-lived workflow artifacts (task briefs, implementer reports, review packages, and a progress ledger). Those artifacts are currently written under the repository's `.git/sdd` path (via `git rev-parse --git-path sdd`). Many agent harnesses (Claude Code and similar) treat writes under `.git/` as protected and prompt for explicit user approval on every write. This produces repeated interactive approval prompts that block automated subagent workflows and degrade the UX.

## Proposed solution

Move the SDD workspace out of the repository's `.git` namespace and into a deterministic top-level working-tree path (proposed: `.superpowers/sdd`). Concretely:

1. Prefer `.superpowers/sdd` as the canonical directory for SDD artifacts.
2. Change scripts and code that use `git rev-parse --git-path sdd` to prefer `.superpowers/sdd` and fall back to `.git/sdd` for compatibility. Example snippet to use in scripts:

```bash
if [ -d .superpowers/sdd ]; then
  dir=.superpowers/sdd
elif git rev-parse --git-path sdd >/dev/null 2>&1 && [ -d "$(git rev-parse --git-path sdd)" ]; then
  dir=$(git rev-parse --git-path sdd)
else
  dir=.superpowers/sdd
fi
mkdir -p "$dir"
```

3. Update `skills/subagent-driven-development/SKILL.md` and relevant docs to reference the new path and show example commands.
4. Add `.superpowers/` to the repository `.gitignore` to avoid accidental commits.
5. Optionally include a migration helper that moves existing `.git/sdd` content to `.superpowers/sdd` on first run (logged) and preserves file permissions where reasonable.
6. Update tests (including `tests/claude-code/*`) to accept the new path or both locations during transition.
7. Add a short release note/changelog entry describing the change and upgrade path.

## What alternatives did you consider?

- Keep using `.git/sdd` and ask harnesses to whitelist it: requires per-harness changes and is out-of-repo, fragile, and inconsistent across environments.
- Add a pre-creation step in controllers that pre-creates all task report files: mitigates prompts in some cases but is a fragile, partial workaround and leaves artifacts hidden under `.git/`.
- Use a configurable environment variable or user config to override the path: valid, but adds complexity and opt-in burden; a repo-default working-tree path avoids per-user configuration.

[...]

PR #1905, opened 2026-07-03
Scope SDD task briefs by plan file

tianma-if:
# Scope SDD task briefs by plan file

Base: `obra/superpowers:dev`  
Head: `msh01/superpowers:codex/scope-sdd-task-briefs-by-plan`

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | GPT-5 |
| Harness + version | Codex desktop app (exact app version not exposed in this session) |
| All plugins installed | Codex app tools exposed in this session, including GitHub, OpenAI bundled browser/chrome/computer-use, OpenAI primary runtime document/pdf/spreadsheet/presentation/template tools, Cloudflare, HeyGen, Hugging Face, Neon Postgres, Notion, Vercel, and local custom HyperFrames/media skills |
| Human partner who reviewed this diff | msh01 |

## What problem are you trying to solve?

Issue #1888 reports that SDD scratch artifacts under `.superpowers/sdd/` silently collide across efforts. One concrete failure is that `task-brief PLAN 1` writes the default path `.superpowers/sdd/task-1-brief.md`, so starting Task 1 for a different plan in the same worktree overwrites the previous Task 1 brief.

I reproduced that narrow failure with a regression test: two different plan files in the same repo both wrote the same default brief path before this change.

## What does this PR change?

Namespaces the default `task-brief` output by plan filename slug:

```text
.superpowers/sdd/<plan-slug>/task-N-brief.md
```

Explicit `OUTFILE` behavior is unchanged.

## Is this change appropriate for the core library?

Yes. It affects core SDD scratch artifact handling and prevents a silent overwrite/data-loss mode for any project using SDD.

## What alternatives did you consider?

I considered solving all of #1888 at once with cleanup and stale-ledger behavior, but that would be a larger behavioral change. This PR takes the narrow, test-backed slice that prevents default task brief collisions across plan files.

## Does this PR contain multiple unrelated changes?

No. The script, test, and SDD wording all describe the same plan-scoped default brief path.

## Existing PRs

- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: none found for #1888 or plan-scoped SDD task brief paths.

## Environment tested

| Harness | Harness version | Model | Model version/ID |
|---------|-----------------|-------|------------------|
| Codex desktop app | exact app version not exposed | GPT-5 | GPT-5 |

## New harness support (required if this PR adds a new harness)

Not applicable.

## Evaluation

[...]

Issue #1888, opened 2026-07-01
subagent-driven-development: .superpowers/sdd scratch artifacts are never pruned and silently collide across efforts


glen-84:
## What problem does this solve?

The `subagent-driven-development` skill writes per-task scratch artifacts to `.superpowers/sdd/` — `task-N-brief.md`, `task-N-report.md`, and a `review-*.diff` (or `task-N-review.diff`) per task, plus the `progress.md` ledger. Nothing ever prunes them. There is no cleanup step in the skill itself nor in `finishing-a-development-branch`, so the directory grows without bound and mixes artifacts from unrelated, already-merged efforts.

Two concrete failure modes:

1. **Silent clobber across efforts.** Artifact names are keyed only by task number (`task-1-report.md`, `task-1-brief.md`, …), not by effort. Starting a new plan reuses `task-1-*`, so the implementer for the new effort's Task 1 overwrites the *previous* effort's `task-1-report.md` with no warning. The old effort's record is destroyed.

2. **Unbounded accumulation.** Non-colliding artifacts from prior efforts (`task-2-*` … `task-12-*`, every `review-*.diff`, one-off files like `task-consolidation-report.md`) linger forever. Observed in a real workspace: `.superpowers/sdd/` held ~40 stale files from a completed-and-merged effort when a new, unrelated effort began. `progress.md` was likewise stale — it still described the old effort until manually overwritten, which is a real hazard given the ledger is documented as the post-compaction recovery map ("trust the ledger and `git log`").

The self-ignoring `.gitignore` (`*`) keeps this out of `git status`/commits, so the mess is invisible until you `ls` the directory — but it is still confusing state that can mislead a resumed or compacted session (a stale ledger is read as authoritative), and it makes the clobber-on-collision data-loss silent.

## Proposed solution

A few options, not mutually exclusive:

1. **Effort-scoped subdirectory.** Namespace artifacts per plan/branch, e.g. `.superpowers/sdd/<plan-slug>/…`, so a new effort can't clobber a prior one and each effort's ledger + reports stay grouped. `sdd-workspace` already centralizes the path resolution, so this is a single-point change.

[...]

PR #1952, opened 2026-07-09
fix: session-scope the SDD workspace so concurrent sessions on one checkout stop colliding

### LINKED ISSUE #1942: SDD workspace (.superpowers/sdd) collides across concurrent sessions sharing a checkout — regression from #1780 (6.1.0)
## Summary

The v6.1.0 change that moved SDD scratch out of `.git/` (#1780) introduced a **cross-session collision**: `task-brief`, `review-package`, and the `progress.md` ledger now write to `<working-tree-root>/.superpowers/sdd/` keyed only by task number, and **that directory is shared by every Claude session running against the same checkout**. Concurrent sessions silently overwrite each other's files.

**Observed symptom:** running `subagent-driven-development` in one session while other sessions ran their own SDD plans against the same repo checkout, implementer subagents received **another plan's task text** — e.g. my plan's `task-3-brief.md` contained a different plan's "Deploy the bridge" task, `task-8-brief.md` contained an unrelated "Auth guard" task. A controller's `progress.md` can likewise reflect another session's state.

## Root cause

- **≤6.0.x:** `task-brief` wrote to `git rev-parse --git-path sdd` = `<git-dir>/sdd`. Each worktree has its own `.git`, so this was **per-worktree isolated**. The script's own header said: *"unique per repo instance, so concurrent sessions cannot collide."*
- **6.1.0 (#1780):** the new `sdd-workspace` helper resolves the dir via `git rev-parse --show-toplevel` = `<working-tree-root>/.superpowers/sdd`. This correctly fixes the ".git/ is a protected path, subagents can't write reports there" problem — but the **working-tree root is shared by all sessions on that checkout**, not per-session. The RELEASE-NOTES say it's "resolved per worktree by a shared `sdd-workspace` helper" — but per *working-tree* is not per *session*, and the keys (`task-N`, `progress.md`) are not session-scoped.

The frequency scales with concurrent-session count, so heavy parallel users hit it constantly while single-session users never see it.

## Impact

Silent, data-race-class: a subagent can build the **wrong task** if it trusts the brief (many controllers dispatch "read your brief at &lt;path&gt;"). The failure is invisible unless the controller inlines requirements and the implementer notices the mismatch.

## Suggested fix

Session-scope the workspace while keeping it out of `.git/`. Options:

1. Key the dir by a session id: `<working-tree-root>/.superpowers/sdd/<session-id>/` (e.g. from a `CLAUDE_SESSION_ID`-style env var, or a per-invocation uuid the controller threads through).
2. Or have `task-brief`/`review-package` stamp each output with its source plan path + task header, and refuse to silently overwrite a file whose stamp doesn't match — so a cross-session clobber is detectable rather than silent.
3. At minimum, document the collision so controllers pass an explicit per-session `OUTFILE` (the scripts already accept a 3rd positional arg).

## Local mitigation we deployed (in case useful to others)

A PreToolUse hook that denies bare `task-brief`/`review-package` calls (the ones that default to the shared dir) and requires an explicit per-session `OUTFILE` or a worktree-scoped cwd, plus guidance to keep dispatch prompts self-contained and have implementers verify the brief's `## Task N` header before trusting it.

**Environment:** superpowers 6.1.0, Claude Code, macOS, multiple concurrent sessions against one repo checkout.

mvanhorn:
## Summary

Session-scope the SDD workspace so concurrent sessions on one checkout stop colliding.

## Why this matters

The v6.1.0 change (#1780) that moved SDD scratch out of `.git/` introduced a cross-session data race. The `sdd-workspace` helper resolves the artifact dir via `git rev-parse --show-toplevel` -> `<working-tree-root>/.superpowers/sdd`, and `task-brief`, `review-package`, and the `progress.md` ledger all write into it keyed only by task number (`task-N-brief.md`, `progress.md`). That directory is shared by every agent session running against the same checkout, so concurrent sessions silently overwrite each other's files: an implementer subagent can receive another plan's task text (the reporter saw `task-3-brief.md` holding a different plan's "Deploy the bridge" task) and a controller's `progr

## Testing

Changes are scoped to the files named in the fix; added or updated tests where the project has a suite, and matched existing conventions.

AI was used for assistance.

Fixes #1942


nedharvey1197:
Reporter of #1942 here — thanks for picking this up, @mvanhorn.

Flagging that this is still blocking real work downstream. We run multiple concurrent Claude Code sessions against one checkout, and the shared `.superpowers/sdd/` dir keyed only by task number is an active hazard: our observed symptom was an implementer subagent receiving **another plan's task text** (a "Task 8" briefed as an unrelated auth task). Silent — the brief looks well-formed, it just belongs to a different plan.

We are currently mitigating locally with a PreToolUse hook that denies bare `task-brief` / `review-package` calls (forcing an explicit per-session OUTFILE), plus a discipline of keeping every dispatch prompt self-contained so a clobbered brief cannot corrupt output. Those work, but they are compensating controls for something the skill should guarantee.

Worth noting the ecosystem has converged on this independently — #1888 (same collision, different angle), #1986 (cross-task subagent reuse), and #1905 (scope briefs by plan file, closed today). That suggests the underlying issue is being hit by more than one user.

Anything we can do to help move this along — testing against our multi-session setup, splitting it into a smaller reviewable change, or narrowing scope to just the session-scoping fix? Happy to help; mostly wanting to make sure it does not stall silently.

mvanhorn:
Thanks for the detailed report - really useful to know it's actively biting a concurrent-session setup, and good to see the convergence with #1888/#1986/#1905. Narrowing this to just the session-scoping fix sounds like the right way to get it moving, so I'll split that out as a focused change. And yes, testing against your multi-session setup once it's ready would be great - I'll ping you when the scoped version is up.

mvanhorn:
Actually, scratch the 'split it out' - I went back through it and this PR is already scoped to just the session-scoping fix. The sdd-workspace script now keys the workspace dir on a sanitized session id (SUPERPOWERS_SDD_SESSION, falling back to CLAUDE_CODE_SESSION_ID, then to the shared dir), so concurrent sessions on one checkout get separate brief/report/diff/progress dirs instead of clobbering each other. The other three files are just the task-brief consumer, the SKILL doc, and tests - nothing broad bundled in. It's mergeable as-is, so no restructuring needed. Testing it against your multi-session setup would be the ideal confirmation; the knob is SUPERPOWERS_SDD_SESSION if you want to force distinct ids.

mvanhorn:
Rebased onto 6.2.0 in 1553cb5, and the rebase turned up something you should know: upstream added plan-scoped SDD workspaces in 6.2.0, which already resolves collisions between different plans to separate `.superpowers/sdd/<plan>` directories. That covers most of what this PR was for.

What's left is narrower: two concurrent sessions executing the same plan in the same working tree still share a directory.

Rather than nest a session dir under your plan dir and end up with two stacked mechanisms, `sdd-workspace` now computes one flat key — `<plan>--<sanitized-session>+<session-hex>` when `SUPERPOWERS_SDD_SESSION` or `CLAUDE_CODE_SESSION_ID` is set, and plain `<plan>` when neither is. The hex suffix keeps byte identity after sanitization so `a/b` and `a?b` can't collapse onto the same workspace.

If you'd rather not carry the same-plan case at all, this is a reasonable thing to close — 6.2.0 got the bigger half of it and I'd rather say that than quietly leave a redundant PR open.

PR #1964, opened 2026-07-11
fix(brainstorm): launch BRAINSTORM_OPEN_CMD without a shell

### LINKED ISSUE #1957: Harden the BRAINSTORM_OPEN launcher: use execFile instead of parsing the env var into argv

I want my org to allow superpowers plugin, but our (skill-based) security audit picked the following signals that block the adoption.

```
Verdict
Severity tier: High

Recommendation: Not usable as-is

Level 1 and Level 2 agree at High, driven by a single confirmed finding: scripts/server.cjs shell-execs an environment variable (BRAINSTORM_OPEN_CMD) when auto-opening the user's browser. The human-judgment pass (Step 3) confirms this is a real code pattern, not a false positive — but finds substantial mitigating context (opt-in only, loopback-only, operator-trust-level input) that a reviewer should weigh.

Safety recommendations
Must fix before use (per the automated High tier):

[H-1] scripts/server.cjs:540 — cp.exec() shell-interprets BRAINSTORM_OPEN_CMD + a JSON-stringified URL. A reviewer who accepts the mitigating context in [H-1]'s Reviewer notes (opt-in, loopback-only, operator-trust-level env var) may reasonably choose to accept this as shipped rather than block on it — but the recommended fix is still worth doing:
Should fix (hardening, low effort):

Replace the BRAINSTORM_OPEN_CMD branch's cp.exec(cmd + ' ' + url) with cp.execFile over a shell-parsed argv (split the operator's command string on whitespace, or require it be pre-split), matching the pattern already used two lines below for the built-in platform launchers (xdg-open, rundll32.exe, open). This removes the shell-injection shape structurally rather than relying on the trust assumption about who can set the env var.
No other hardening items identified — the rest of the server (auth token, timing-safe compare, Origin check, CSP headers, path-traversal guards on /files/, PID-identity verification in stop-server.sh, chmod 600 on session files) reflects deliberate, above-average security engineering for a local dev tool.
```

## Proposed solution

As stated by the audit agent, one possible solution is to change how cp.exec is used

> Replace the BRAINSTORM_OPEN_CMD branch's cp.exec(cmd + ' ' + url) with cp.execFile over a shell-parsed argv (split the operator's command string on whitespace, or require it be pre-split), matching the pattern already used two lines below for the built-in platform launchers (xdg-open, rundll32.exe, open)

Create a custom fork of the skill, but that would defeat the purpose of the marketplace.

Yes, other organizations might find the same audit result, blocking adoption.

## Environment (required)

| Field | Value |
|-------|-------|
| Harness version | 2.1.197 (Claude Code) |
| Your model + version | Sonnet 5 |
| All plugins installed | No |

aznikline:
## What

Hardens the `BRAINSTORM_OPEN_CMD` browser-launch path in `skills/brainstorming/scripts/server.cjs` so it no longer goes through a shell. Closes #1957.

## Why

The security audit in #1957 flagged that `maybeOpenBrowser()` ran the operator-provided `BRAINSTORM_OPEN_CMD` via `cp.exec(cmd + ' ' + JSON.stringify(url))`, i.e. through `sh -c`. The env var is opt-in, loopback-only, trusted operator input, so the practical risk is bounded — but the shell-exec shape is what the audit (and downstream org security reviews) block adoption on. The issue's proposed resolution was to match the platform launchers two lines below, which already use `execFile` with the URL as a single argv element.

## How

- Replaced the `cp.exec` branch with `cp.execFile(bin, args.concat(url), …)` — no shell, the URL is a single argv element.
- Added a quote-aware tokenizer `shellArgv()` that splits the operator's command string into an argv honoring single/double quotes (so a command like `node "/path/with space/capture.cjs" "/marker/path"` still resolves to the right tokens, matching how the existing test fixture builds `BRAINSTORM_OPEN_CMD`).
- A command with **unbalanced quotes** returns `null` and is rejected — it is never handed to a shell as a fallback. The previous `try/catch` best-effort semantics are preserved on the actual spawn.

This mirrors the existing `browserLauncherForPlatform` pattern (bin + args + `execFile`), so the two launcher paths now share the same no-shell shape.

## Tests

- Existing `lifecycle.test.js` tests for `BRAINSTORM_OPEN_CMD` still pass — their fixture (`node "script" "marker"`) tokenizes cleanly to `[node, script, marker]`, with the URL appended as the final element.
- Added a regression test: *`BRAINSTORM_OPEN_CMD` launches without a shell and keeps the URL as one argv element*. It captures the launched `argv` and asserts the server URL (which carries a `?key=` query) arrives as exactly one token with no shell-splitting — the direct proof the injection shape is gone.

## Behavior change

Operators who set `BRAINSTORM_OPEN_CMD` to a command relying on shell features (pipes, `&&`, env expansion, globbing) will need to wrap it in a script — the command is now tokenized and `execFile`'d, not shell-interpreted. This matches how the platform launchers already behave and is the intended hardening. I called this out in the env var's comment.

## Checklist

- [x] Change is scoped to the flagged line and its helper
- [x] No new shell surface introduced
- [x] Existing tests green; added a regression test for the no-shell contract

### DIFF
--- a/skills/brainstorming/scripts/server.cjs
+++ b/skills/brainstorming/scripts/server.cjs
@@ -301,6 +301,33 @@ function browserLauncherForPlatform(url, {
   return null;
 }
 
+// Quote-aware tokenizer for the trusted BRAINSTORM_OPEN_CMD operator input.
+// Splits a single- or double-quoted command string into an argv (no shell, no
+// metacharacter expansion), so execFile can launch it without a `sh -c` layer.
+// Returns null on unbalanced quotes so a malformed command is never handed to
+// a shell as a fallback.
+function shellArgv(input) {
+  const argv = [];
+  let cur = '';
+  let quote = null;
+  for (let i = 0; i < input.length; i++) {
+    const ch = input[i];
+    if (quote) {
+      if (ch === quote) { quote = null; continue; }
+      cur += ch;
+    } else if (ch === '"' || ch === "'") {
+      quote = ch;
+    } else if (ch === ' ' || ch === '\t' || ch === '\n') {
+      if (cur.length) { argv.push(cur); cur = ''; }
+    } else {
+      cur += ch;
+    }
+  }
+  if (quote) return null; // unbalanced quotes — refuse to guess
+  if (cur.length) argv.push(cur);
+  return argv.length ? argv : null;
+}
+
 function isRegularFileInsideContentDir(filePath) {
   let stat, realContentDir, realFilePath;
   try {
@@ -535,9 +562,17 @@ function maybeOpenBrowser() {
   if (clients.size > 0) return; // the user already opened it
   const url = companionUrl(); // must carry the key or the gate 403s it
   const cp = require('child_process');
-  // Operator-provided launcher: run as given (this env var is trusted operator input).
+  // Operator-provided launcher: this env var is trusted operator input (opt-in,
+  // loopback-only), but we still avoid the shell entirely. Split the command into
+  // an argv with a quote-aware tokenizer and pass the URL as a single argv element
+  // via execFile, mirroring the platform launchers below — so a URL containing
+  // shell metacharacters can never reach a shell. A command that fails to tokenize
+  // (unbalanced quotes) is rejected rather than handed to a shell.
   if (process.env.BRAINSTORM_OPEN_CMD) {
-    try { cp.exec(process.env.BRAINSTORM_OPEN_CMD + ' ' + JSON.stringify(url), () => {}); } catch (e) { /* best effort */ }
+    const argv = shellArgv(process.env.BRAINSTORM_OPEN_CMD);
+    if (argv) {
+      try { cp.execFile(argv[0], argv.slice(1).concat(url), () => {}); } catch (e) { /* best effort */ }
+    }
     return;
   }
   // Platform launchers: pass the URL as an argv element via execFile (no shell),
--- a/skills/brainstorming/visual-companion.md
+++ b/skills/brainstorming/visual-companion.md
@@ -102,6 +102,22 @@ scripts/start-server.sh \
 
 Use `--url-host` to control what hostname is printed in the returned URL JSON.
 
+### Custom browser launcher: `BRAINSTORM_OPEN_CMD`
+
+The `--open` flag auto-launches the user's default browser using the platform's native launcher. For environments where that doesn't apply — a container with no desktop session, a headless dev box reached via a tunnel, or a non-standard browser binary — set `BRAINSTORM_OPEN_CMD` to the command that should open a URL:
+
+```bash
+BRAINSTORM_OPEN_CMD='firefox' scripts/start-server.sh --project-dir /path --open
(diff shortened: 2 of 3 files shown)


muunkky:
Independently took a run at this before I noticed #1964 — and landed on the same core fix (quote-aware tokenizer → `execFile`, no shell). Nice to see it converge; good sign it's the right shape.

Two bits from my version that might be worth folding into #1964, if useful:

- **Operator docs** — a `BRAINSTORM_OPEN_CMD` block in `visual-companion.md` (accepted shape, trust posture, the no-shell note). The issue title mentions docs, and #1964 is code-only: https://github.com/muunkky/superpowers/blob/showcase/1957/skills/brainstorming/visual-companion.md
- **A direct injection test** — set `BRAINSTORM_OPEN_CMD` to `node "<capture>" "<marker>" ; touch "<pwned>"` and assert `<pwned>` is never created — i.e. prove the metacharacters don't execute (complements your URL-integrity assertion): https://github.com/muunkky/superpowers/blob/showcase/1957/tests/brainstorm-server/lifecycle.test.js

Also minor: #1964 targets `main`; but I pushed mine to `dev` as per contribution rules.

Full write-up on my fork if it's worth a skim — PRD → design → ADR: [PRD](https://github.com/muunkky/superpowers/blob/showcase/1957/docs/prds/PRD-001-brainstorm-companion-launcher-injection-shape.md) · [design](https://github.com/muunkky/superpowers/blob/showcase/1957/docs/designs/DD-001-brainstorm-launcher-shell-free-argv.md) · [ADR](https://github.com/muunkky/superpowers/blob/showcase/1957/docs/adr/ADR-001-brainstorm-launcher-shell-free-argv.md). Happy to send the docs + test as a follow-up to #1964 rather than open a competing PR.

*Disclosure (per contribution rules): produced with an autonomous development harness — gitban (muunkky.github.io/gitban-site) — on Claude Code 2.1.207 / Claude Opus 4.8; reasoned from the issue + direct code inspection.*

aznikline:
@muunkky — really nice writeup, and a strong signal to see two independent passes converge on the same tokenizer → `execFile` shape. 

I folded both of your bits straight into #1964 (appreciate you offering them as a follow-up rather than competing):

- **Operator docs** — added a `BRAINSTORM_OPEN_CMD` section to `visual-companion.md` covering the accepted shapes (bare command / command+args / quoted argv), the trust posture (env var = invoking user's authority), and the no-shell note (the launcher tokenizes and runs without a shell, so `;`/`&` in the value are literal). The issue title does call out docs, and #1964 was code-only — good catch.
- **Direct injection test** — added alongside the URL-integrity regression: `BRAINSTORM_OPEN_CMD` set to a capture command with a `; touch <pwned>` tail, asserting `<pwned>` is never created. Complements the existing assertion that the URL's `&`/`?` stay a single argv element.

Also retargeted the PR base `main` → `dev` per the contribution rules (good flag — I'd missed that). Branch is rebased onto `dev`, `mergeable`/`CLEAN`. Full brainstorm-server suite green (136 tests: 125 node + 11 bash, including both the URL-integrity and the new injection case).

@obra — ready for review when you have a moment. Closes #1957.

PR #1967, opened 2026-07-11
systematic-debugging: check for self-ingested output in agent/LLM pipelines

CucukovskiNikola:
In multi-turn agent systems the model's own prior output — a reply, a status
line, or a cached summary from an earlier turn — can loop back into a
downstream stage's input and fabricate "impossible" state the user never
supplied.

This adds a new Phase-1 step (step 6, "Check for Self-Ingested Output") that
directs the debugger to trace what each stage actually READS, not just what
the user sent, and to check the provenance of the exact string the failing
stage matched on. A matching Red Flags entry catches the "symptom points at
feature X, so the bug is in X" reflex before it skips that check.

Both additions slot into the existing Phase-1 data-flow section and Red Flags
list; no other content changes.

### DIFF
--- a/skills/systematic-debugging/SKILL.md
+++ b/skills/systematic-debugging/SKILL.md
@@ -119,6 +119,27 @@ You MUST complete each phase before proceeding to the next.
    - Keep tracing up until you find the source
    - Fix at source, not at symptom
 
+6. **Check for Self-Ingested Output (agent/LLM pipelines)**
+
+   **WHEN debugging a multi-turn agent or LLM pipeline that reaches "impossible" state:**
+
+   The system's OWN prior output can loop back as input. A generated reply, a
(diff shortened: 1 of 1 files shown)


obra:
Thanks for the submission. Closing this one: the change modifies tuned skill content without any real-world evidence — no session, error, or issue where self-ingested output actually bit someone — which makes it a speculative fix under this repo's contribution rules. It also targets `main` (PRs go to `dev`) and skips the PR template entirely.

If you actually hit this failure mode in a real debugging session, a resubmission on `dev` with the full template, the concrete session evidence, and eval results showing the added step improves outcomes would get a proper look — the underlying pattern is interesting, but skill text only changes here with evidence behind it.

— Claude Fable 5 (AI assistant), Claude Code 2.1.218, doing PR triage for this repo

PR #1970, opened 2026-07-11
Fix dead references to pruned claude-code-tools.md/copilot-tools.md (#1969)

rasibintang:
> Targets `dev`.

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Claude Fable 5 (`claude-fable-5`) |
| Harness + version | Claude Code, VS Code extension (Agent SDK) |
| All plugins installed | superpowers 6.1.1 (claude-plugins-official marketplace) |
| Human partner who reviewed this diff | @rasibintang |

## What problem are you trying to solve?

#1969: commit `e7ddc25` deliberately deleted `references/claude-code-tools.md` and `references/copilot-tools.md`, but two places still point at them as live files. `skills/writing-skills/SKILL.md`'s Overview (line 12) links both dead files as where to find your runtime's skills directory — every agent that loads writing-skills reads a pointer to nowhere, and the Claude Code personal-skills path is no longer documented on the path that sentence gives. `docs/porting-to-a-new-harness.md`'s Appendix A ("use this as the live index") lists the deleted files as the tool-mapping entries for Claude Code, Cursor, and Copilot CLI.

Found by a mechanical audit of every relative markdown link target under `skills/` (same defect class as #1962/#1511: deliberate removal, dangling references left behind).

## What does this PR change?

Four lines. The writing-skills sentence now inlines the Claude Code path (`~/.claude/skills/`) and links only the surviving reference files (`codex-tools.md`, `gemini-tools.md`); the porting-guide table marks the three affected harnesses' adapter files as pruned instead of linking them.

## Is this change appropriate for the core library?

Yes — it fixes broken pointers in a core skill and a core contributor doc. No new content, dependencies, or third-party anything.

## What alternatives did you consider?

1. **Restore the two deleted files** — rejected: `e7ddc25` removed them deliberately ("nothing left that wasn't generic"); restoring would revert a maintainer decision.
2. **Delete the sentence entirely** — rejected: its payload (where personal skills live) is real guidance; only the pointers were stale. Kept the sentence, fixed the pointers.
3. Leaving RELEASE-NOTES/spec mentions untouched was deliberate — they are historical records and correct as written.

## Does this PR contain multiple unrelated changes?

No — one defect (dangling references left by `e7ddc25`), fixed at both sites where it appears.

## Existing PRs
- [x] I have reviewed all open AND closed PRs for duplicates or prior art
[...]

Issue #1969, opened 2026-07-11
writing-skills and porting guide still link claude-code-tools.md / copilot-tools.md, deleted in e7ddc25


rasibintang:
- [x] I searched existing issues and this has not been proposed before

## What happened

Commit `e7ddc25` ("Prune per-harness tool-mapping boilerplate") deliberately deleted `references/claude-code-tools.md` and `references/copilot-tools.md` — the commit message says they "had nothing left that wasn't generic". Two places still point at the deleted files as if they were live:

1. **`skills/writing-skills/SKILL.md` (Overview, line 12)** links both dead files as the place to find your runtime's skills directory:
   > see [claude-code-tools.md](../using-superpowers/references/claude-code-tools.md), [codex-tools.md](…), [copilot-tools.md](../using-superpowers/references/copilot-tools.md), or [gemini-tools.md](…) for the path on your runtime

   `skills/using-superpowers/references/` on current `dev` contains only `antigravity-tools.md`, `codex-tools.md`, `gemini-tools.md`, and `pi-tools.md`. Every agent that loads writing-skills is pointed at two files that don't exist — and the sentence's actual payload (where Claude Code personal skills live) is now documented nowhere on the path it gives.

2. **`docs/porting-to-a-new-harness.md` (Appendix A table)** lists `references/claude-code-tools.md` as the live tool-mapping file for Claude Code and Cursor, and `references/copilot-tools.md` for Copilot CLI. The table's own caption says "use this as the live index" — new-harness porters following it hit missing files.

`RELEASE-NOTES.md` and the 2026-05-05 spec also mention the files, but those are historical records and correct as written.

## How found

Mechanical audit: extracted every relative markdown link target under `skills/` and checked existence (same defect class as #1962). These two were the only genuine dead references after filtering quoted examples.

## Proposed fix

Point the writing-skills sentence only at surviving reference files and inline the Claude Code path (`~/.claude/skills/`); mark the pruned adapters as pruned in the porting table. I have a 4-line-diff branch ready and will open a PR.

---
*Authoring disclosure: investigated and written by Claude Fable 5 (`claude-fable-5`) on Claude Code (VS Code extension), superpowers 6.1.1 installed; human partner @rasibintang directed and reviewed.*

obra:
[...]

PR #1902, opened 2026-07-03
Document Antigravity tool mappings for skill loading

tianma-if:
# Document Antigravity tool mappings for skill loading

Base: `obra/superpowers:dev`  
Head: `msh01/superpowers:codex/document-antigravity-tool-mappings`

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | GPT-5 |
| Harness + version | Codex desktop app (exact app version not exposed in this session) |
| All plugins installed | Codex app tools exposed in this session, including GitHub, OpenAI bundled browser/chrome/computer-use, OpenAI primary runtime document/pdf/spreadsheet/presentation/template tools, Cloudflare, HeyGen, Hugging Face, Neon Postgres, Notion, Vercel, and local custom HyperFrames/media skills |
| Human partner who reviewed this diff | msh01 |

## What problem are you trying to solve?

`bash tests/antigravity/run-tests.sh` fails on `upstream/dev` because `skills/using-superpowers/references/antigravity-tools.md` does not document the Antigravity skill-load/tool mapping expected by the test:

```text
FAIL: mapping does not document view_file as the file/skill-read tool
```

The test expects the platform reference to describe how Antigravity loads skill files without a `Skill` tool and to list the core Antigravity tools used by Superpowers workflows.

## What does this PR change?

Adds the missing Antigravity mapping rows for `view_file`, `write_to_file`, `replace_file_content`, `run_command`, `grep_search`, and `invoke_subagent`, plus short sections for skill loading and subagent support.

## Is this change appropriate for the core library?

Yes. This updates an existing core platform reference for a supported harness and makes the existing Antigravity regression test pass.

## What alternatives did you consider?

I considered changing the test, but the test describes the intended Antigravity behavior: no `Skill` tool, read `SKILL.md` with `view_file` and `IsSkillFile`. Updating the reference is the direct fix.

## Does this PR contain multiple unrelated changes?

No. It only updates the Antigravity tool mapping reference.

## Existing PRs

- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: #1657 added Antigravity CLI support. Older closed attempts include #535, #682, and #192. This PR does not add Antigravity support; it fills missing mapping details in the existing reference so the current test passes.

## Environment tested

| Harness | Harness version | Model | Model version/ID |
|---------|-----------------|-------|------------------|
| Codex desktop app | exact app version not exposed | GPT-5 | GPT-5 |

## New harness support (required if this PR adds a new harness)

[...]

PR #1847, opened 2026-06-25
Prune per-harness tool-mapping boilerplate

obra:
> Targets `dev`.

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Claude Opus 4.8 (`claude-opus-4-8[1m]`) |
| Harness + version | Claude Code 2.1.186 |
| All plugins installed | superpowers (this repo, local dev build), elements-of-style, episodic-memory, superpowers-developing-for-claude-code, superpowers-lab, superpowers-chrome, claude-session-driver, plugin-dev, frontend-design, agent-sdk-dev, mcp-server-dev, code-simplifier, claude-code-setup, github-triage, linear, context7, release-radar, primeradiant-ops, summarize-meetings, worldview-synthesis, gopls/rust-analyzer/swift LSPs |
| Human partner who reviewed this diff | Jesse Vincent (@obra), maintainer — reviewed the complete diff before submission |

## What problem are you trying to solve?

The per-harness reference files under `using-superpowers/references/` had grown
large action→tool mapping tables plus skill-loading explainers that restate
guidance modern agents already follow on their own. That's tokens spent on
boilerplate. Two files (claude-code-tools.md, copilot-tools.md) had nothing left
that wasn't generic once the boilerplate was stripped.

## What does this PR change?

Trims `antigravity-tools.md`, `codex-tools.md`, and `pi-tools.md` down to the
harness-specific notes that still carry weight (subagent dispatch, task tracking,
instructions-file paths). Deletes `claude-code-tools.md` and `copilot-tools.md`.

These reference files are loaded on demand when a harness needs them — they are
not part of the always-injected bootstrap, so this does not change baseline
skill-triggering behavior.

## Is this change appropriate for the core library?

Yes — they're the core per-harness adaptation docs for the using-superpowers
skill, used by all users on those harnesses.

## What alternatives did you consider?

Leaving the tables intact. Rejected: they re-explain tool equivalents agents
already infer, and the maintenance/token cost isn't justified for content that
doesn't change behavior.

## Does this PR contain multiple unrelated changes?

No. One problem: prune per-harness reference boilerplate. One of a four-PR series
splitting a local working branch into independent PRs against `dev`. The
sibling "Compress the using-superpowers bootstrap" PR owns the `SKILL.md` changes
(including its Platform-Adaptation harness list); this PR owns only the
`references/*.md` files. The four touch disjoint files and merge in any order.

## Existing PRs
[...]

PR #1987, opened 2026-07-13
test: realign antigravity + pi mapping assertions with pruned references

gaurav0107:
> **Targets the `dev` branch.**

| Field | Value |
|-------|-------|
| Your model + version | Claude Opus 4.8 (1M context) — model ID `claude-opus-4-8[1m]` |
| Harness + version | Claude Code 2.1.143 |
| All plugins installed |  `superhuman@superhuman`, `superpowers@claude-plugins-official` |
| Human partner who reviewed this diff | Gaurav Dubey (@gaurav0107) |

The plugin-infrastructure test suite fails on `dev` (and on `main`, 6.1.1). Two per-harness mapping tests assert content that was intentionally deleted:

```
$ git checkout dev
$ bash tests/antigravity/test-antigravity-tools.sh
FAIL: mapping does not document view_file as the file/skill-read tool   # exit 1

$ node tests/pi/test-pi-extension.mjs
✖ pi tools reference documents pi-specific mappings
  AssertionError: The input did not match the regular expression /write/   # exit 1 (1 of 6 cases)
```

Root cause: commit `e7ddc25` ("Prune per-harness tool-mapping boilerplate", 2026-06-24) deliberately removed the skill-loading explainers and the generic action→tool tables from `antigravity-tools.md` and `pi-tools.md`, trimming each file "to the harness-specific notes that still carry weight (subagent dispatch, task tracking…)". That commit did **not** touch `tests/`, so the two content-assertion tests still grep for tokens the prune removed — `view_file`, `IsSkillFile`, `run_command`, `grep_search` (antigravity) and `read`/`write`/`edit`/`bash` (pi). With no CI in the repo, the breakage went unnoticed.

## What does this PR change?

Updates the two orphaned tests to assert only the harness-specific mappings `e7ddc25` kept (subagent dispatch and task tracking) instead of the boilerplate it removed. No reference, skill, or product content is touched — only the stale test assertions (and their now-inaccurate header comment / PASS string).

Yes. These are the repo's own plugin-infrastructure tests under `tests/`. No new dependency, no skill-behavior change, no third-party integration — it makes the existing suite green again on `dev`.

1. **Restore the pruned content to the reference files** so the tests pass unchanged — rejected: it directly contradicts `e7ddc25`'s stated intent, and the restore-direction attempt **PR #1902** ("Document Antigravity tool mappings for skill loading") was already **closed**. Reintroducing the boilerplate would re-break what the maintainer deliberately trimmed.
2. **Delete the two content-assertion tests entirely** — rejected: the reference files still exist and still document harness-specific mappings worth guarding. Trimming the assertions keeps meaningful coverage (subagent dispatch, task artifact, SKILL.md link) while dropping only the pruned tokens.

No. One root cause — `e7ddc25` orphaned the per-harness mapping-content tests. Both files fail for the identical reason and are fixed the identical way. One problem.

## Existing PRs
- Related PRs: **#1902** (closed) attempted the opposite fix — restoring the pruned antigravity skill-load docs; its closure is what confirms the reference should stay lean and the *tests* should adapt instead. **#1969 / #1970** address a different slice of `e7ddc25` fallout (dead markdown links to the fully-deleted `claude-code-tools.md` / `copilot-tools.md`), not these test assertions. No open PR targets these two test files as a focused fix. (Several large, CONFLICTING, stale sweep-PRs incidentally include these files but do not repair the assertions.)

## Environment tested

| Claude Code | 2.1.143 | Claude Opus 4.8 (1M context) | `claude-opus-4-8[1m]` |

Tests run locally on macOS (`bash` + `node --test`).

N/A — this PR does not add a harness.

## Evaluation

Not a behavior/skill change, so there is no eval-session delta to report. Verification instead:
- **Before** (on `dev`): `tests/antigravity/test-antigravity-tools.sh` → exit 1; `tests/pi/test-pi-extension.mjs` → 5 pass / 1 fail.
- **After**: antigravity → `PASS`; pi → 6 pass / 0 fail.
- Also ran `tests/antigravity/run-tests.sh` (all passed) and the repo shell-lint tests (pass). Confirmed each surviving asserted token is actually present in the pruned reference file.

## Rigor

- [ ] If this is a skills change: … — **N/A, not a skills change** (tests only).
- [x] This change was tested adversarially, not just the happy path — verified failing-first on both `dev` and `main`, and confirmed the new assertions match the pruned references.
- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) — no skill/prose content touched.

## Human review
- [x] A human has reviewed the COMPLETE proposed diff before submission — @gaurav0107 reviewed the full two-file diff and approved it.

### DIFF
--- a/tests/antigravity/test-antigravity-tools.sh
+++ b/tests/antigravity/test-antigravity-tools.sh
@@ -2,8 +2,9 @@
 # Validate the Antigravity (agy) integration. agy installs the existing plugin
 # directly (`agy plugin install <repo-url>`): it loads the bundled skills and
 # runs the SessionStart hook for bootstrap, so there is no agy-specific scaffold
-# to test. What IS agy-specific is the tool mapping — agy has no `Skill` tool and
-# loads skills by reading SKILL.md with view_file — and SKILL.md pointing at it.
+# to test. What IS agy-specific is the tool mapping — subagent dispatch via
+# invoke_subagent (self/research types) and task tracking via a task artifact —
+# and SKILL.md pointing at it.
 #
 # Mirrors tests/pi/test-pi-extension.mjs's "tools reference documents
 # harness-specific mappings" check. CI-safe: does not require `agy` installed.
@@ -22,16 +23,8 @@ echo "test-antigravity-tools: checking Antigravity tool mapping"
 # --- Mapping exists ---------------------------------------------------------
 [ -f "$MAPPING" ] || fail "tool mapping missing at $MAPPING"
 
-# --- Skill-load mechanism: view_file on SKILL.md (IsSkillFile), no Skill tool -
-grep -qiE "view_file" "$MAPPING" \
-  || fail "mapping does not document view_file as the file/skill-read tool"
-grep -qiE "SKILL\.md" "$MAPPING" \
-  || fail "mapping does not document reading SKILL.md as the skill-load path"
-grep -q "IsSkillFile" "$MAPPING" \
-  || fail "mapping does not document setting IsSkillFile when loading a skill"
-
 # --- Core action→tool mappings are documented -------------------------------
-for tool in write_to_file replace_file_content run_command grep_search invoke_subagent; do
+for tool in write_to_file replace_file_content invoke_subagent; do
   grep -q "$tool" "$MAPPING" \
     || fail "mapping does not document the '$tool' tool"
 done
@@ -50,4 +43,4 @@ grep -qE 'ArtifactType.*task|task. artifact' "$MAPPING" \
 grep -q "antigravity-tools.md" "$SKILL" \
   || fail "SKILL.md Platform Adaptation does not reference antigravity-tools.md"
 
-echo "PASS: Antigravity tool mapping valid (view_file skill-load, agy tools, SKILL.md link)"
+echo "PASS: Antigravity tool mapping valid (subagent dispatch, task artifact, SKILL.md link)"
--- a/tests/pi/test-pi-extension.mjs
+++ b/tests/pi/test-pi-extension.mjs
@@ -122,7 +122,7 @@ test('pi tools reference documents pi-specific mappings', async () => {
   assert.equal(existsSync(piToolsPath), true, 'pi-tools.md should exist');
   const text = await readFile(piToolsPath, 'utf8');
 
-  for (const expected of ['Skill', 'Task', 'TodoWrite', 'read', 'write', 'edit', 'bash']) {
+  for (const expected of ['subagent', 'pi-subagents', 'Task', 'TODO.md']) {
     assert.match(text, new RegExp(expected));
   }
 });
--- a/tests/pi/test-pi-extension.mjs
+++ b/tests/pi/test-pi-extension.mjs
@@ -122,7 +122,16 @@ test('pi tools reference documents pi-specific mappings', async () => {
   assert.equal(existsSync(piToolsPath), true, 'pi-tools.md should exist');
   const text = await readFile(piToolsPath, 'utf8');
 
-  for (const expected of ['subagent', 'pi-subagents', 'Task', 'TODO.md']) {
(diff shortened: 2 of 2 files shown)


muunkky:
The antigravity half holds up. I stripped the mapping table out of `antigravity-tools.md` and your test correctly fails on `invoke_subagent`, which is the one token that only ever appears in the table. That's the right shape.

The pi half doesn't hold. All four of the new tokens also occur in the prose, so the assertion survives the thing it exists to detect. On your branch, with the pi mapping table deleted:

```
$ grep -v '^|' skills/using-superpowers/references/pi-tools.md > t && mv t skills/using-superpowers/references/pi-tools.md
$ node --test tests/pi/test-pi-extension.mjs
# pass 6
# fail 0
```

`subagent` matches the intro line, `pi-subagents` and `Task` both match the Subagents paragraph, and `TODO.md` matches the Task lists paragraph. `Task` is the one I'd look at twice: it matches "do not fabricate `Task` calls", a sentence whose entire job is to say Pi has no Task tool.

So the test passes on a `pi-tools.md` with no mapping table at all, which is the regression `e7ddc25` caused in the first place.

Scoping the match to the table rows closes it:

```js
const rows = text.split('\n').filter((line) => line.startsWith('|'));
assert.ok(rows.some((row) => /subagent/i.test(row)), 'mapping table documents subagent dispatch');
assert.ok(rows.some((row) => /todo|task/i.test(row)), 'mapping table documents task tracking');
```

That fails on the stripped file and passes on `dev`. Being straight about it: that snippet is the diff in my #1983, which does the pi half and nothing else. Take it or write your own — yours covers antigravity too, so it's the better PR either way and I'll close mine.

Disclosure: agent-assisted — Claude Opus 4.8 (`claude-opus-4-8[1m]`), Claude Code 2.1.208. Plugins: gitban (muunkky.github.io/gitban-site), frontend-design, skill-creator, claude-patent-creator-standalone. Not `superpowers` itself. Grounded in the runs above, against your branch at `8296a98`.

gaurav0107:
Thanks @muunkky — this is exactly right, and I reproduced it before changing anything. Stripping the table rows from `pi-tools.md` and re-running the test, my token list passed 6/0, so the assertion was surviving the regression it exists to catch. `Task` matching "do not fabricate `Task` calls" was the clearest tell.

Pushed the table-scoped version (`0873545`) — filtering to rows that start with `|`, then asserting `/subagent/i` and `/todo|task/i` against those rows. Verified it now fails on a table-stripped `pi-tools.md` and passes on `dev`. The approach and snippet are from your #1983 — credited in the commit message.

I checked the antigravity half against the same test: it already fails when the table is stripped (`invoke_subagent` only appears in the row), so I left it as-is. Since this PR now covers both halves, feel free to close #1983 — and thanks for doing the adversarial pass.

Disclosure: agent-assisted — Claude Opus 4.8 (`claude-opus-4-8[1m]`), Claude Code 2.1.143, reviewed and approved by @gaurav0107. Plugins unchanged from the PR description.

cwest:
Backing this one — and I've closed my overlapping PR (#1992) in favor of it.

I hit the same red antigravity suite on `dev` and first went the other way:
restoring the pruned rows in `antigravity-tools.md`. Then I re-read #1847 —
that content was cut deliberately ("the harness-specific notes that still
carry weight") — so re-adding it fights the intent. Realigning the stale
tests, as you've done, is the right call.

Reproduced locally:
- `tests/antigravity/run-tests.sh` → green
- the pi mapping case flips red→green against the slimmed `pi-tools.md`
- a clean `agy` run with the plugin installed loads skills and auto-triggers
  `brainstorming` without those mapping lines present — the empirical version
  of "agents mostly don't need 'em"

Honest caveat: I couldn't run pi extension cases 2–5 headlessly (they load the
TS extension and my setup wouldn't), so I verified the mapping assertion, not
those four.

Thanks for taking the pi half down to the table rows too.

PR #1999, opened 2026-07-17
feat: make SDD/brainstorming workspace root configurable via SUPERPOWERS_WORKSPACE_DIR

### LINKED ISSUE #1991: SDD/brainstorming workspace root is hardcoded to .superpowers — make it configurable

My projects keep all agent scratch under a single gitignored `.work/` directory. Specs and plans already land there — the agent writes them and can honor a path preference from CLAUDE.md. SDD scratch cannot: it's written by the plugin's `sdd-workspace` script inside dispatched subagents, so no CLAUDE.md or memory preference changes where that bash script writes.

The workspace root is hardcoded to `<working-tree-root>/.superpowers/` — `sdd-workspace` resolves `$root/.superpowers/sdd`, and brainstorming's `start-server.sh` writes `$root/.superpowers/brainstorm/`. There is no override. The only options are a per-repo `.superpowers` → `.work/superpowers` symlink, or editing the plugin (global, wiped on update).

## Proposed solution

Expose the workspace root via a single config point that `sdd-workspace` (and brainstorming's `start-server.sh`) honor, in precedence order:

1. An env var, e.g. `SUPERPOWERS_WORKSPACE_DIR`.
2. Falling back to today's `$root/.superpowers`.

`sdd-workspace` is already the single source of truth for the SDD path, so this is a one-line change there plus the equivalent in `start-server.sh`.

- **Per-repo symlink** `.superpowers` → `.work/superpowers` — works, but every repo needs it and it's easy to forget.
- **Editing the cached plugin script** — global across projects and wiped on every plugin update.
- **The existing `OUTFILE` 3rd-positional arg** on `task-brief`/`review-package` — only covers those two outputs; `sdd-workspace` itself and `progress.md` stay hardcoded, so the root can't move.

Related but distinct: #337 (configurable *plan* location, closed) and #939/#975 (spec/brainstorm output location) cover the artifacts the agent authors, not the plugin-written workspace root.

Yes — it's domain-independent. Anyone who consolidates agent scratch under a single gitignored directory, or wants SDD/brainstorm state out of the default tracked-tree location, benefits.

## Environment (required)

| Field | Value |
|-------|-------|
| Harness version | (unknown) |
| Your model + version | Claude Opus 4.8 (1M context) |
| All plugins installed | superpowers, caveman, chillicream-skills, dotnet-* suite |

## Context

Hit while redirecting a project's `.superpowers/sdd` into its gitignored `.work/` directory.

_Drafted by Claude (Anthropic AI assistant)._

aznikline:
## Summary

Closes #1991. Makes the SDD and brainstorming workspace directory configurable instead of hardcoded to `<repo-root>/.superpowers`.

### SDD (`sdd-workspace`)
- Replaced `dir="$root/.superpowers/sdd"` with `dir="${SUPERPOWERS_WORKSPACE_DIR:-$root/.superpowers}/sdd"`
- Added header comment documenting the env var

### Brainstorming (`start-server.sh`)
- Replaced `"${PROJECT_DIR}/.superpowers/brainstorm/"` with `"${ws_root}/brainstorm/"` where `ws_root="${SUPERPOWERS_WORKSPACE_DIR:-${PROJECT_DIR}/.superpowers}"`
- Port and token file paths also use `$ws_root` automatically
- Clean fallback: if `SUPERPOWERS_WORKSPACE_DIR` is unset, behaves exactly as before

### Docs
- `visual-companion.md`: Added `$SUPERPOWERS_WORKSPACE_DIR` alongside the hardcoded path mention

## Backward compatibility

Fully backward compatible. When `SUPERPOWERS_WORKSPACE_DIR` is not set, all paths resolve to the same `.superpowers/` locations as before. The env var is a pure additive override — no existing workflows are affected.

## Testing

- `bash -n` syntax check passed for both shell scripts
- Existing brainstorm tests (lifecycle.test.js etc.) should pass unchanged when env var is unset (defaults to same paths)

### DIFF
--- a/skills/brainstorming/scripts/server.cjs
+++ b/skills/brainstorming/scripts/server.cjs
@@ -301,6 +301,33 @@ function browserLauncherForPlatform(url, {
   return null;
 }
 
+// Quote-aware tokenizer for the trusted BRAINSTORM_OPEN_CMD operator input.
+// Splits a single- or double-quoted command string into an argv (no shell, no
+// metacharacter expansion), so execFile can launch it without a `sh -c` layer.
+// Returns null on unbalanced quotes so a malformed command is never handed to
+// a shell as a fallback.
+function shellArgv(input) {
+  const argv = [];
+  let cur = '';
+  let quote = null;
+  for (let i = 0; i < input.length; i++) {
+    const ch = input[i];
+    if (quote) {
+      if (ch === quote) { quote = null; continue; }
+      cur += ch;
+    } else if (ch === '"' || ch === "'") {
+      quote = ch;
+    } else if (ch === ' ' || ch === '\t' || ch === '\n') {
+      if (cur.length) { argv.push(cur); cur = ''; }
+    } else {
+      cur += ch;
+    }
+  }
+  if (quote) return null; // unbalanced quotes — refuse to guess
+  if (cur.length) argv.push(cur);
+  return argv.length ? argv : null;
+}
+
 function isRegularFileInsideContentDir(filePath) {
   let stat, realContentDir, realFilePath;
   try {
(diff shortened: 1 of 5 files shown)


jeneric:
This solves a concrete Perl Module::Build distribution problem for me: .gitignore keeps SDD artifacts out of git status, but distcheck uses MANIFEST.SKIP, so in-tree artifacts become unlisted distribution files. Moving the workspace outside the source tree avoids ecosystem-specific exclusions. Thanks for implementing this.

PR #122, opened 2025-11-26
fix(opencode): Use `synthetic: true` for invisible content injection

joshuadavidthomas:
The OpenCode plugin injects the bootstrap and skill content as visible user messages. OpenCode's `TextPartInput` supports a `synthetic` Boolean property that hides parts from the UI while keeping them visible to the AI -- it's been available in the v2 message format since around the v1.0.0 version bump.

## Motivation and Context

Bootstrap content (the `<EXTREMELY_IMPORTANT>` block) appears as a visible user message at the start of every OpenCode session. When loading skills via `use_skill`, the full skill content also appears as a visible message. While `noReply: true` prevents an AI response, the injected content still appears in the chat UI which makes it just a bit noisy.

## How Has This Been Tested?

Tested locally in OpenCode v1.0.114:
- Fresh session: Bootstrap is now invisible, AI still knows about superpowers
- Loading a skill: Both "Loading skill: X" and full content are now hidden
- AI correctly receives and uses all injected content

Chat session before change: [https://opencode.ai/s/Z3AqwrES](https://opencode.ai/s/Z3AqwrES)
Chat session after change: [https://opencode.ai/s/WjTuvSy6](https://opencode.ai/s/WjTuvSy6)

## Breaking Changes

None. This is purely a UI visibility change - the AI receives identical content.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the MCP Documentation (https://modelcontextprotocol.io) (I mean, I have, but it's a bit irrelvant to this change 😄)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Summary by CodeRabbit

* **Chores**
  * Updated internal message handling to mark certain message parts as synthetic when injecting bootstrap content and launching loaded skills. This may affect how messages are processed downstream but does not change user-visible behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2004, opened 2026-07-18
fix(opencode): port plugin to the OpenCode v2 plugin API

saiqulhaq:

| Field | Value |
|-------|-------|
| Your model + version | Claude Opus 4.8 (1M context), model ID `claude-opus-4-8[1m]` |
| Harness + version | Claude Code 2.1.179 (code authored here) |
| All plugins installed | `superpowers` (Claude Code plugin) — the code was tested against the plugin being changed; plus unrelated MCP servers (ClickUp, Slack, Gmail, Obsidian) that played no role |
| Human partner who reviewed this diff | @saiqulhaq — reviewed the full diff and ran it locally against OpenCode v2 before submission |

The bundled OpenCode plugin (`.opencode/plugins/superpowers.js`) targets the **OpenCode v1 plugin API** and does not function under **OpenCode v2**. OpenCode's own migration guide states plainly: *"V1 plugins will not work in V2."* (https://v2.opencode.ai/migrate-v1)

Concretely, under v2 the current plugin is loaded but never does anything, because v2 replaced the entire plugin contract:

- v1 exported a **named** `SuperpowersPlugin({ client, directory })` factory; v2 expects a **default** `{ id, setup }` export. v2 never calls the v1 factory, so `setup` never runs.
- v1 registered the skills directory by mutating `config.skills.paths` inside a `config` hook; that hook does not exist in v2.
- v1 injected the bootstrap via `experimental.chat.messages.transform` over `{ info, parts }` messages; v2 removed that hook and changed the message shape to `{ role, content: [{ type, text }] }`.

Result on OpenCode v2: no skills registered, no bootstrap injected, so no superpowers skills auto-trigger. Verified on a real `opencode2` (`next-15718`) install: with the old plugin the `using-superpowers` bootstrap is absent and skills never fire.

## What does this PR change?

Rewrites `.opencode/plugins/superpowers.js` against the OpenCode v2 plugin API (`{ id, setup }` default export, `ctx.skill.transform` for skill registration, `ctx.session.hook("context", …)` for bootstrap injection), and updates the OpenCode tests and docs to match. Behavior is otherwise identical to the v1 plugin (same bootstrap content, same module-level caching from #1202, same double-injection guard).

Yes. This is not a new harness or a new skill — it repairs an **existing, first-party** integration that ships in core (`.opencode/`, `tests/opencode/`, `docs/README.opencode.md`). OpenCode support is already part of core; this keeps it working as OpenCode moves to v2. It adds no third-party dependency (the plugin remains zero-dependency; `Plugin.define` is an identity wrapper, so no runtime import is required) and does not touch any skill content.

- **Import `@opencode-ai/plugin/v2` and use `Plugin.define`.** Rejected: `define` is an identity function, and adding a runtime import risks module-resolution failures across the git-package/npm/local install layouts superpowers ships through. A plain default export is equivalent and dependency-free. Types can still be consulted for development.
- **Keep v1 and v2 side by side (dual export).** Rejected: v2 ignores the v1 factory entirely, so it would be dead code, and OpenCode states v1 plugins don't run under v2. A clean v2 rewrite is simpler and matches upstream's direction.
- **Register skills via a config file instead of `ctx.skill.transform`.** Rejected: the transform API is the v2-sanctioned way and keeps the symlink-free, no-manual-config property the v1 plugin had.
- **Hook name `"request"` (as the v2 docs table shows) vs `"context"`.** The shipped `@opencode-ai/plugin` types and the running `opencode2` build both use `"context"`; verified live that `"context"` fires and mutates messages. Used `"context"`.

No. Every change is the single concern of making the OpenCode plugin work on v2: the plugin rewrite, its tests, and its docs. The one-line edits to `docs/porting-to-a-new-harness.md` update that guide's OpenCode references (mechanism, log-check command) to stay consistent with the rewritten plugin.

## Existing PRs
- Related PRs: none found for the v1→v2 plugin-API migration. Searched open + closed PRs and issues for `opencode`, `opencode2`, `opencode v2`, `Plugin.define`, `session.hook`, `v2 plugin`. The closest prior OpenCode work is #122 (merged, v1-era `synthetic:true` injection) and #1247 (an unrelated `import.meta.dirname` refactor of the v1 plugin) — neither addresses the v2 API.

## Environment tested

| OpenCode v2 (`opencode2`) | `v0.0.0-next-15718` | OpenCode default free model | `build · hy3-free` |

## New harness support

Not a new harness — this is a v1→v2 API migration for the existing OpenCode integration. Including the acceptance evidence anyway, since the change affects whether the bootstrap loads.

Ran the canonical acceptance test on a real `opencode2` (`next-15718`) install with this plugin loaded, in a clean empty directory:

> **User:** Let's make a react todo list

The agent auto-triggered the `brainstorming` skill before writing any code (from `--format json` output):

```
"tool":"skill"
"name":"brainstorming"
"text":"Using brainstorming to shape this before building.\n\nLet me first check the current project state."
"text":"Empty project — a clean slate. Let me ask the first clarifying question."
```

Also verified directly (via a sidecar probe plugin) that:
- `ctx.skill.list()` returns the bundled superpowers skills (`brainstorming`, `systematic-debugging`, `test-driven-development`, `using-superpowers`, …) — so `ctx.skill.transform` + `reload` registered the `skills/` directory.
- The `ctx.session.hook("context", …)` callback fires, first-user-message parts have the `{ type, text }` shape, and the injected bootstrap reaches the model.

`tests/opencode/run-tests.sh` (non-integration: plugin-loading + bootstrap-caching) passes with the rewritten v2-driven caching test.

### DIFF
--- a/.opencode/INSTALL.md
+++ b/.opencode/INSTALL.md
@@ -1,16 +1,22 @@
 # Installing Superpowers for OpenCode
 
+> **OpenCode v2 only.** This plugin uses the OpenCode v2 plugin API. OpenCode
+> v1 plugins do not run under v2 — use an older Superpowers release for
+> OpenCode v1.
+
 ## Prerequisites
 
-- [OpenCode.ai](https://opencode.ai) installed
+- [OpenCode v2](https://v2.opencode.ai) installed
 
 ## Installation
 
-Add superpowers to the `plugin` array in your `opencode.json` (global or project-level):
+Add superpowers to the `plugins` array in your `opencode.json` (global or
+project-level). `plugins` is the canonical v2 key; the legacy singular `plugin`
+key still loads for now but may be removed:
 
 ```json
 {
-  "plugin": ["superpowers@git+https://github.com/obra/superpowers.git"]
+  "plugins": ["superpowers@git+https://github.com/obra/superpowers.git"]
 }
 ```
 
@@ -39,6 +45,12 @@ rm -rf ~/.config/opencode/superpowers
 
 Then follow the installation steps above.
 
+## Migrating from OpenCode v1
+
+If your `opencode.json` used the singular `plugin` key, rename it to `plugins`.
+No other changes are needed — the same git package spec works, and the plugin
+now loads through the OpenCode v2 API automatically.
+
 ## Usage
 
 Use OpenCode's native `skill` tool:
@@ -59,7 +71,7 @@ To pin a specific version:
 
 ```json
 {
-  "plugin": ["superpowers@git+https://github.com/obra/superpowers.git#v5.0.3"]
+  "plugins": ["superpowers@git+https://github.com/obra/superpowers.git#v5.0.3"]
 }
 ```
 
@@ -67,9 +79,10 @@ To pin a specific version:
 
 ### Plugin not loading
 
-1. Check logs: `opencode run --print-logs "hello" 2>&1 | grep -i superpowers`
-2. Verify the plugin line in your `opencode.json`
-3. Make sure you're running a recent version of OpenCode
+1. Check logs: `grep -i "loading plugin" ~/.local/share/opencode/log/opencode.log` (OpenCode v2 logs to a file; the CLI `run` command no longer has a `--print-logs` flag)
+2. Verify the plugin line in your `opencode.json` uses the `plugins` key
+3. If the shared background service is stuck, restart it: `opencode service restart`
+4. Make sure you're running OpenCode v2
 
 ### Windows install issues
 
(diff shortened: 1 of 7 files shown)


obra:
It looks like this would immediately break all opencode v1 users. is that right? If so, it's probably a no-go. Is there a way to add opencode 2 support that won't break v1?

obra:
Closing in favor of #2106, which covers the same OpenCode v2 port with dual v1/v2 compatibility — existing v1 users keep working, and it carries acceptance transcripts for both versions. Thanks for being first to flag the v2 API break and for the port work here; the v2 direction is happening, just via the dual-support shape.

PR #1994, opened 2026-07-16
fix: make the codex packaging and SDD skill test suites pass reliably off-macOS

obra:
## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Claude Fable 5 (claude-fable-5) |
| Harness + version | Claude Code CLI (Linux), maintainer session |
| All plugins installed | superpowers (this repo, dev), superpowers-lab, superpowers-chrome, superpowers-developing-for-claude-code, episodic-memory, elements-of-style, plugin-dev, agent-sdk-dev, frontend-design, code-simplifier, claude-code-setup, mcp-server-dev, linear, context7, primeradiant-ops, claude-session-driver, github-triage, release-radar, summarize-meetings, jam, landing-page-design, scenario-testing, simmer, speed-run, worldview-synthesis |
| Human partner who reviewed this diff | Jesse Vincent (@obra) — asked for these failures to be fixed after they surfaced in a full test sweep; reviewing the complete diff in this PR |

## What problem are you trying to solve?

Running the full test matrix on dev (Linux, GNU tar, umask 002, TZ=UTC) during pre-release verification of #1993 turned up two failing suites, both failing on clean dev with no local changes:

1. **`tests/codex/test-package-codex-plugin.sh` fails hard on any Linux box.** `tar: unrecognized option '--uid'` — the packaging script's deterministic-metadata flags are bsdtar spellings that GNU tar rejects, so the tar.gz archive step dies and three assertions cascade. Two more portability bugs hide behind it: staged file modes only came out 755 because git archive's `tar.umask` default (0002 → 775) happened to be re-masked by macOS's default process umask (022 → 755) — on a umask-002 Linux box the executable lands as 775 and the mode assertion fails; and the test's timestamp assertion parses bsdtar's `-tv` column layout expecting epoch 0 rendered as "Dec 31 1969", which is both the wrong column split for GNU tar and the wrong string on any UTC host.
2. **`tests/claude-code/test-subagent-driven-development.sh` flakes.** Observed 3 failures across 4 runs, each different: a spurious timeout (the file's worst case is 9 prompts × 90s = 810s, but the runner kills files at 600s), and two prose-matching misses where the model capitalized the skill's own headings — output containing "Do Not Trust the Report" failed the case-sensitive pattern `not trust`, and a structured answer failed `First:.*spec.*compliance`.

## What does this PR change?

[...]

PR #1901, opened 2026-07-03
Fix Codex package timestamp normalization across timezones

tianma-if:
# Fix Codex package timestamp normalization across timezones

Base: `obra/superpowers:dev`  
Head: `msh01/superpowers:codex/fix-codex-package-timestamps`

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | GPT-5 |
| Harness + version | Codex desktop app (exact app version not exposed in this session) |
| All plugins installed | Codex app tools exposed in this session, including GitHub, OpenAI bundled browser/chrome/computer-use, OpenAI primary runtime document/pdf/spreadsheet/presentation/template tools, Cloudflare, HeyGen, Hugging Face, Neon Postgres, Notion, Vercel, and local custom HyperFrames/media skills |
| Human partner who reviewed this diff | msh01 |

## What problem are you trying to solve?

On a machine using the Asia/Shanghai timezone, `bash tests/codex/test-package-codex-plugin.sh` fails on the archive timestamp assertions even though the package script intends to normalize archive entry timestamps.

Observed failures:

```text
[FAIL] zip archive normalizes entry timestamps
  expected: (1980, 1, 1, 0, 0, 0)
  actual:   (1980, 1, 1, 8, 0, 0)
[FAIL] tar.gz archive normalizes entry timestamps
  expected: Dec 31 1969
  actual:   Jan 1 1970
```

The root cause is timezone handling. The script sets `TZ=UTC` for `touch`, but not for the subsequent `zip` / `tar` archive creation process. The tar test also asserted localized `tar -tv` display text instead of archive metadata.

## What does this PR change?

Runs the `zip` and `tar` archive creation commands with `TZ=UTC`, and changes the tar timestamp test to inspect member `mtime` metadata directly.

## Is this change appropriate for the core library?

Yes. This is a deterministic packaging fix for the core Codex portal package script and its regression test.

## What alternatives did you consider?

I considered changing only the test expectations for the local timezone, but that would leave archive creation dependent on the caller's environment. Setting `TZ=UTC` during archive creation preserves the package script's deterministic archive contract.

## Does this PR contain multiple unrelated changes?

No. Both changed files address the same Codex package timestamp normalization failure.

## Existing PRs

- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: #1876 added the Codex portal package script. I found no open or closed PR specifically fixing timezone-dependent timestamp normalization.

## Environment tested

[...]

PR #2007, opened 2026-07-19
fix(codex): make package timestamps timezone-independent

SeuPut0705:

| Field | Value |
|-------|-------|
| Your model + version | OpenAI GPT-5.6-sol (primary and final review); GPT-5.6-terra (implementation and task review) |
| Harness + version | ChatGPT/Codex desktop 26.715.31925 (build 5551), embedded Codex CLI 0.145.0-alpha.18; local Codex CLI 0.144.4 |
| All plugins installed | documents, pdf, spreadsheets, presentations, template-creator 26.715.12143; sites 0.1.30; browser and chrome 26.715.31925; computer-use 1.0.1000451; visualize 1.0.12; slack, game-studio, github, codex-security 2f1a8948; active Superpowers skill bundle 6.1.1 |
| Human partner who reviewed this diff | @SeuPut0705 |

Running `tests/codex/test-package-codex-plugin.sh` on macOS in Asia/Seoul produced two timestamp failures. ZIP entries were stored as `(1980, 1, 1, 9, 0, 0)` instead of midnight, and the tar assertion compared a localized date string (`Dec 31 1969`) with macOS bsdtar's `Jan 1 1970` display.

Direct reproduction showed that UTC and Asia/Seoul ZIP packages had different SHA-256 values and stored `00:00` versus `09:00`, while both tar packages were byte-identical and had raw member `mtime` values of `0`. The ZIP command inherited the host timezone even though the preceding `touch` ran under UTC; the tar test was inspecting presentation text rather than archive metadata.

## What does this PR change?

Run Info-ZIP with `TZ=UTC`, add a regression test that compares real ZIP bytes produced under UTC and Asia/Seoul, and inspect tar member mtimes through Python's `tarfile` API instead of localized `tar -tvf` output.

Yes. This fixes the repository's Codex packaging script and its integration test without adding dependencies or changing skill behavior. Deterministic portal packages benefit every maintainer producing the official Codex archive.

- Keeping UTC only on `touch`: this is the current behavior and leaves Info-ZIP's DOS timestamp conversion timezone-dependent.
- Relaxing the ZIP timestamp assertion: this would hide non-reproducible package bytes.
- Rebuilding ZIP output with Python: this would replace the existing production archive path for a one-environment-variable fix.
- Comparing `tar -tvf` under a forced locale/timezone: the rendered date remains tool-specific; raw `member.mtime` is the portable archive contract.

No. The one-line packaging fix and both test changes address the same archive timestamp determinism defect.

## Existing PRs

- [x] I have reviewed all open AND closed PRs for duplicates or prior art.
- Related PRs: none found for `package-codex-plugin timestamp`, `archive timezone deterministic`, or `Codex package zip` in open/closed PRs and issues.

## Environment tested

| ChatGPT/Codex desktop on macOS 26.5.2, Asia/Seoul | 26.715.31925 (build 5551), embedded Codex CLI 0.145.0-alpha.18 | OpenAI | GPT-5.6-sol / GPT-5.6-terra |
| Shell integration test: bsdtar 3.5.3, Info-ZIP 3.0, Python 3.14.6 | local Codex CLI 0.144.4 | N/A | N/A |

## New harness support

Not applicable. This PR does not add or change harness support.

## Evaluation

- Initial prompt: the human partner approved proceeding with a PR after the KST packaging test failure was reproduced during Codex package verification.
- Before: explicit UTC and Asia/Seoul ZIP builds had different SHA-256 values and timestamps; the focused test reported two failures.
- After: the focused packaging integration test passes, including byte-identical UTC/Asia/Seoul ZIP output and raw tar `mtime == 0`; the marketplace manifest test and `git diff --check` also pass.
- Model eval sessions after the change: 0. This is packaging infrastructure, not behavior-shaping skill content.

## Rigor

- [x] This is not a skills change; `superpowers:writing-skills` is not applicable.
- [x] The change was tested adversarially with explicit UTC and Asia/Seoul environments.
- [x] No carefully tuned behavior-shaping content was modified.

## Human review

### DIFF
--- a/scripts/package-codex-plugin.sh
+++ b/scripts/package-codex-plugin.sh
@@ -294,7 +294,7 @@ case "$FORMAT" in
     (
       cd "$STAGE"
       rm -f "$OUTPUT"
-      COPYFILE_DISABLE=1 zip -X -q - -@ <"$ARCHIVE_LIST" >"$OUTPUT"
+      COPYFILE_DISABLE=1 TZ=UTC zip -X -q - -@ <"$ARCHIVE_LIST" >"$OUTPUT"
     )
     ;;
   tar.gz)
--- a/tests/codex/test-package-codex-plugin.sh
+++ b/tests/codex/test-package-codex-plugin.sh
@@ -135,6 +135,8 @@ echo "Codex package archive tests"
 
 metadata_source="$TEST_ROOT/metadata-source"
 archive="$TEST_ROOT/superpowers"
+utc_archive="$TEST_ROOT/superpowers-utc.zip"
+seoul_archive="$TEST_ROOT/superpowers-seoul.zip"
 tar_archive="$TEST_ROOT/superpowers.tar.gz"
 extracted="$TEST_ROOT/extracted"
 tar_extracted="$TEST_ROOT/tar-extracted"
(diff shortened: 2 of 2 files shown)


Rob-Morris:
Independent confirmation of this bug and fix, reproduced against `main` @ `44c9b2d` on macOS in `TZ=Australia/Sydney`:

```
[FAIL] zip archive normalizes entry timestamps
   expected: (1980, 1, 1, 0, 0, 0)
   actual:   (1980, 1, 1, 11, 0, 0)
```

With `TZ=UTC` added to the `zip` invocation (this PR's change), the full test suite passes under `TZ=Australia/Sydney`.

Two points worth adding to the record:

1. **Timezones west of UTC pass only by accident.** `1980-01-01T00:00:00Z` is 1979 in local time there, which predates the DOS epoch, so zip clamps up to the `1980-01-01 00:00:00` floor and lands on the expected value. Any post-1980 mtime shows the real behaviour — same file, mtime `2020-06-01T00:00:00Z`, zipped under three TZs records three different DOS timestamps (`(2020,6,1,0,0,0)` / `(2020,6,1,10,0,0)` / `(2020,5,31,20,0,0)`). So the archive is not byte-reproducible across timezones today even where the test passes — this is a packaging-determinism bug that the test only catches east of UTC.
2. This was previously diagnosed and maintainer-validated in #1901 (closed for unrelated reasons). Note the description of merged #1994 says #1901's zip-side `TZ=UTC` fix "already landed" — it didn't; `scripts/package-codex-plugin.sh` on `main` and `dev` still has no `TZ=UTC` on the zip line.

Issue #1560, opened 2026-05-16
[plugin] SessionStart hook needs timeout, descriptions are uniform, one skill is large


jeandeloynes-prog:

arittr:
Duplicate of #1561, which was already closed as not planned (and this copy has no body). Closing.

PR #2019, opened 2026-07-22
Fix/sessionstart hook timeout

TotalyNotABotTrustMe:
## What this fixes

Part of #1560.

The SessionStart hook blocks session start (`async: false`) but has no `timeout` set in `hooks/hooks.json`, so it falls back to Claude Code's 600s default. If the hook ever hangs, that's a 10-minute stall on every session start with no explanation.

In practice the hook just reads one file and does some string escaping — it should finish in milliseconds. Set an explicit 10s timeout so a hang fails fast instead of silently eating 10 minutes.

## What this doesn't cover

#1560 also mentions "descriptions are uniform" and "one skill is large" (writing-skills/SKILL.md is 679 lines vs. the next-largest at 503). The issue body was empty, so I didn't want to guess at what "descriptions are uniform" means — happy to take a pass at the skill-size part separately if that's useful, or if a maintainer can clarify the other one.

## Testing

- `hooks/hooks.json` still parses as valid JSON
- `tests/hooks/test-session-start.sh` passes


obra:
Closing as a duplicate of your own #2020, which carries the same SessionStart-timeout fix as a clean two-file diff against `dev` (the correct target branch). The ~50 extra files here are the already-released v6.1.1 content picked up from a stale base, not new work — nothing in this PR is missing from #2020. Follow-up review will happen on #2020.

— Claude Fable 5 (AI assistant), Claude Code 2.1.218, doing PR triage for this repo

PR #2022, opened 2026-07-23
fix: resolve symlinked skill directories to their real path

tobilafinhangit:
## Summary
- Multi-IDE setups symlink `.cursor/skills/`, `.claude/skills/`, etc. back to a single `.agent/skills/` source of truth (per the multi-IDE symlink strategy documented in this file's own header comment).
- Without resolving symlinks first, the same skill directory can be discovered/loaded more than once under different paths.
- Adds `getRealPath()` to resolve a directory to its real path before use, and exports it alongside the existing helpers.

## Test plan
- [ ] Set up a symlinked skills directory (e.g. `.cursor/skills -> .agent/skills`) and confirm skill discovery no longer double-counts the same skill under both paths.

### DIFF
--- a/lib/skills-core.js
+++ b/lib/skills-core.js
@@ -3,6 +3,42 @@ import path from 'path';
 import { execSync } from 'child_process';
 
 /**
+ * Skills Core Library
+ *
+ * Multi-IDE Skill Directory Standards:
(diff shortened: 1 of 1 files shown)


tobilafinhangit:
Closing — this was opened as part of internal repo cleanup, not intended to go out publicly. Apologies for the noise.

PR #2027, opened 2026-07-23
Release v6.2.0

obra:
Release PR for Superpowers v6.2.0 — merges `dev` into `main`. Release prep (notes + version bumps) landed on dev via #2026.

## ⚠️ Merge with a MERGE COMMIT

Not rebase-merge, not squash. The v6.1.1 rebase-merge minted duplicate SHAs on `main` and left the branches permanently diverged (source of this cycle's conflicts). A merge commit puts dev's real SHAs on `main`, so the next release's merge base is dev's tip and future `dev→main` merges are clean.

## Conflicts

This PR initially shows 11 conflicting files — all residue of the 6.1.1 rebase-merge (version manifests, RELEASE-NOTES.md, the porting-guide table predating #1969, add/add on the Codex package script + test). Every conflict resolves to dev's side; a verified resolution merge commit (merged tree byte-identical to dev's tree, so nothing from main is lost) is ready to push to `dev`, after which this PR merges clean.

## What's in v6.2.0

- **SDD: plan-scoped workspace** — `.superpowers/sdd/<plan>/` per plan, ledger names its plan, workspace deleted once the final review is clean; follow-up plans can no longer adopt a previous plan's ledger.
- **SDD: resume-based fix loop** — fix rounds resume the implementer, scoped re-reviews via `re-review-prompt.md`, five-round circuit breaker with controller adjudication.
- **Skills compression sweep** — recap/social-proof/benefits prose removed across a dozen skills, load-bearing arguments folded into rationalization tables (eval-gated; the one cut that degraded behavior was reworked instead).
- **`testing-anti-patterns.md` → `writing-good-tests.md`** — positive six-rule catalog absorbing falsifiability discipline, string-presence and change-detector traps.
- **`finishing-a-development-branch`** — no more discard offer in the completion menu, forge-agnostic PR creation, worktree-path capture fix.
- **Windows** — SessionStart hook dispatches via Git Bash (`shell: "bash"`), fixing #1751 and #1918.
- **Gemini CLI restored** (#1959) pending proper evaluation of the removal.
- **Fixes** — `find-polluter.sh` pattern matching (#2008/#2011), Codex package script portability beyond macOS/bsdtar, SDD test flake, doc/test cleanup after the v6.1.0 reference pruning (#1969, #2010).

Full notes: `RELEASE-NOTES.md` v6.2.0 entry.

## Post-merge checklist

- [ ] `git tag -a v6.2.0` on the main merge commit, push tag
[...]

PR #2010, opened 2026-07-20
docs(using-superpowers): drop dangling subagent-support anchor

ferrous26:
> **Targets the `dev` branch.**

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Claude Opus 4.8 (1M context) — model ID `claude-opus-4-8[1m]` |
| Harness + version | Claude Code 2.1.215 |
| All plugins installed | A downstream fork of superpowers (`braze-superpowers`, tracking upstream 6.1.1) plus internal Braze plugins. Not the upstream `superpowers` plugin as-shipped. |
| Human partner who reviewed this diff | Mark Rada (@ferrous26) |

## What problem are you trying to solve?

`skills/using-superpowers/references/antigravity-tools.md` contains a link that resolves to nothing.

The dispatch row tells an agent where to learn the difference between the two built-in Antigravity subagent types:

```
| Dispatch a subagent … | `invoke_subagent` with a built-in `TypeName` — `self` for full-capability
work, `research` for read-only (see [Subagent support](#subagent-support)) |
```

There is no `## Subagent support` heading in the file. The only headings are `# Antigravity CLI (agy) Tool Mapping` and `## Task tracking`. An agent that follows the pointer to decide between `self` and `research` lands nowhere.

This is leftover from the prune in `e7ddc25e` ("Prune per-harness tool-mapping boilerplate"). That commit deliberately deleted the `## Subagent support` section but left the inline cross-reference behind:

| tag | file size | `## Subagent support` | cross-reference |
|-----|-----------|----------------------|-----------------|
| v6.0.3 | 5691 B | present | consistent |
| v6.1.0 | 1539 B | **removed** | **left behind** |
| v6.1.1 | 1539 B | removed | left behind |
| `dev` @ cc69047 | 1539 B | removed | left behind |

Still present on `dev` and `main` today.

**How it was found:** not from an Antigravity session. I was syncing a downstream fork to 6.1.1 and ran a static anchor check over the skill tree — every `[text](#anchor)` link compared against the headings actually present in that file. This was the one mismatch.

## What does this PR change?

Removes the dangling `(see [Subagent support](#subagent-support))` parenthetical from the subagent dispatch row in `antigravity-tools.md`. One line, one file.

## Is this change appropriate for the core library?

Yes — it repairs a file already in core, and does not add anything. It is not project-, team-, or tool-specific, and integrates no third-party service. It finishes a cleanup that core already started in `e7ddc25e`.

## What alternatives did you consider?

[...]

Issue #2008, opened 2026-07-20
systematic-debugging: find-polluter.sh find -path never matches (silent no-op)


Mhm0ud:
### Skill: `systematic-debugging` — `find-polluter.sh`

`find-polluter.sh` collects test files with:

```bash
TEST_FILES=$(find . -path "$TEST_PATTERN" | sort)
```

with the documented example `TEST_PATTERN='src/**/*.test.ts'`.

When `find` searches from `.`, every result path is prefixed with `./`, so `-path "src/**/*.test.ts"` matches nothing (paths look like `./src/...`). The script then reports "Found 1 test files" (from `wc -l` on an empty string) and exits without testing anything — a silent no-op that reads as "all tests clean".

**Fix** — match the `./`-prefixed form:

```bash
TEST_FILES=$(find . -path "./$TEST_PATTERN" | sort)
```

(Note: an `-o -path "$TEST_PATTERN"` alternative doesn't help — the non-prefixed pattern still can't match `find .` output.) Alternatively guard against the empty-result case so `TOTAL` reflects 0.

Found while vendoring the skill into an internal project; reporting upstream rather than forking our snapshot.

obra:
Fixed on dev, pending release. Your diagnosis was exactly right — `find .` emits `./`-prefixed paths so the documented pattern never matched, and `wc -l` on the empty result reported "Found 1 test files" while running nothing. I reproduced the silent no-op in a toy repo before merging.

Landed as two changes on dev: #2011 (community fix: `./` prefix + honest zero count) and follow-up commit 0146173, which also strips a caller-supplied leading `./` and matches `**/` against zero directory levels (so `src/**/*.test.ts` no longer skips top-level `src/*.test.ts` files), with a deterministic test suite added. Ships in the next release — reopen if you can still hit it after that.

— Claude Fable 5 (AI assistant), Claude Code 2.1.218, doing issue triage for this repo

Resapea:
can you please remove me from the emails? I am NOT even registered on the
list but get these constantly. Ty

obra:
I can not. You've subscribed to either this issue or all issues on the github side. You'll have to log into github and click into the project to un-watch or un-subscribed.  There's also an unsubscribe link in the footer of the messages

PR #2028, opened 2026-07-24
Release v6.2.0

obra:
Release PR for Superpowers v6.2.0. Supersedes #2027 (dev→main directly), which cannot merge linearly: dev carries the original pre-6.1.1 commits whose rebased twins on main diverged during the last release's conflict resolution, so a flat rebase of dev conflicts.

This branch is `origin/main` + the 51 genuinely-new commits since v6.1.1 (`c809093..d262bc4`), replayed onto main with **zero conflicts**. The resulting tree is **byte-identical to dev's release tree** (`git diff d262bc4 release-v6.2.0-linear` is empty) — same content that landed via #2026, nothing added or lost.

## ⚠️ Merge with "Rebase and merge"

The branch is already based on main's tip, so the replay is trivially clean and main's history stays linear. Do not squash (collapses 51 commits into one) and do not create a merge commit (adds a bubble).

## Post-merge: merge main back into dev IMMEDIATELY

This is the step that prevents next cycle's conflicts. Done right after the release, the trees are identical so the merge is automatically clean. It advances dev's merge base to main's tip without rewriting dev — no open PR against dev breaks.

## What's in v6.2.0

- **SDD: plan-scoped workspace** — `.superpowers/sdd/<plan>/` per plan, ledger names its plan, workspace deleted once the final review is clean; follow-up plans can no longer adopt a previous plan's ledger.
- **SDD: resume-based fix loop** — fix rounds resume the implementer, scoped re-reviews via `re-review-prompt.md`, five-round circuit breaker with controller adjudication.
- **Skills compression sweep** — recap/social-proof/benefits prose removed across a dozen skills, load-bearing arguments folded into rationalization tables (eval-gated; the one cut that degraded behavior was reworked instead).
- **`testing-anti-patterns.md` → `writing-good-tests.md`** — positive six-rule catalog absorbing falsifiability discipline, string-presence and change-detector traps.
- **`finishing-a-development-branch`** — no more discard offer in the completion menu, forge-agnostic PR creation, worktree-path capture fix.
- **Windows** — SessionStart hook dispatches via Git Bash (`shell: "bash"`), fixing #1751 and #1918.
- **Gemini CLI restored** (#1959) pending proper evaluation of the removal.
- **Fixes** — `find-polluter.sh` pattern matching (#2008/#2011), Codex package script portability beyond macOS/bsdtar, SDD test flake, doc/test cleanup after the v6.1.0 reference pruning (#1969, #2010).

Full notes: `RELEASE-NOTES.md` v6.2.0 entry.

## Post-merge checklist

- [ ] Merge main back into dev and push (see above — do this first)
- [ ] `git tag -a v6.2.0` on the main tip, push tag
- [ ] `gh release create v6.2.0` with the v6.2.0 notes section
- [ ] Bump `superpowers-marketplace` registry to 6.2.0 and push
- [ ] Verify `/plugin install superpowers@superpowers-marketplace`

---
*Prepared by Claude Fable 5 (`claude-fable-5`) on Claude Code 2.1.218, directed by @obra.*


arittr (approved):
LGTM

Orvid:
The plan scoped cleanup is a major issue when running a series of plans (1 higher power agent producing 1 spec, decomposing into a set of smaller implementation plans to be driven by a medium power agent), since, at least when running with Claude Sonnet 5, it is common for those to be executed in a way that will defer minor issues to future plans in the series, up to and including various minor issues, and I've found that being able to tell an agent at the end to do a complete verification for all minor issues mentioned by any reviewer at the end of a plan series is a good way to catch a few minor issues that would otherwise have stuck around despite being detected and known. Should it default to either asking if you want to delete it, or else not delete them at all? I have no problems with it deleting the various mechanically generated things, as all of that is trivial to regenerate if it's ever needed, it's just the implementers reports that aren't so easy to recreate.

In my current instance, I was able to recover the part that I actually cared about, the implementers reports, from the conversation logs, but it would be nice to not need to rely on the model's memory to prevent these from being deleted in the future.

PR #2053, opened 2026-07-29
fix(cursor): resolve session hook from plugin root

ShiroKSH:

| Field | Value |
|-------|-------|
| Your model + version | GPT-5 (exact build ID is not exposed by the active harness) |
| Harness + version | Codex Desktop 26.715.72359 |
| All plugins installed | openai-bundled: browser 26.715.72359, sites 0.1.30, visualize 1.0.14; openai-curated-remote: codex-security 0.1.14, figma 2.0.16, github 0.1.8-2841cf9749ae, openai-templates 0.1.0; openai-primary-runtime: documents, pdf, presentations, spreadsheets, template-creator 26.727.11326 |
| Human partner who reviewed this diff | ShiroKSH — approved the complete diff in the task |

Cursor marketplace plugin hooks run with the opened workspace as their working directory. The relative command in `hooks/hooks-cursor.json` therefore looks for `./hooks/run-hook.cmd` in the workspace instead of the installed plugin, so SessionStart cannot inject the Superpowers bootstrap. [Cursor support confirms](https://forum.cursor.com/t/inconsistent-working-directory-for-plugin-hook-commands/153236) that `${CURSOR_PLUGIN_ROOT}` is the supported way to address the plugin directory from a hook command. This regressed after `2b25774`, which replaced the root-qualified command introduced for the Cursor hook integration.

## What does this PR change?

Restore the root-qualified Cursor dispatcher command, update the porting guide to describe the same contract, and add a regression test that rejects a relative Cursor hook command.

Yes. This is the shipped Cursor plugin bootstrap path and affects every Superpowers user running Cursor marketplace hooks.

- Keep a relative command: it depends on Cursor using the plugin cache as its working directory, which it does not for marketplace hooks.
- Invoke `session-start` directly: that reintroduces the Windows extensionless-script failure addressed by the existing `run-hook.cmd` wrapper.

No. The config, documentation, and regression test all enforce the same Cursor hook-path contract.

## Existing PRs

- Related PRs: #709, #1054, #1280. #709 introduced the root-qualified Cursor command; #1054 and #1280 changed the dispatcher for Windows but retained a relative path. This PR repairs the later path regression rather than duplicating either fix.

## Environment tested

| Codex Desktop | 26.715.72359 | GPT-5 | Exact build ID not exposed by the harness |

N/A — this does not add a harness.

## Evaluation

- Initial prompt: review `obra/superpowers` for a proven bug, preserve business logic, and choose a real contribution rather than cosmetic refactoring.
- Skill-behavior eval sessions after the change: 0; this modifies hook wiring only and no `SKILL.md` content.
- Before / after: the prior configuration resolved the dispatcher from the workspace; the new configuration resolves it from Cursor's supplied plugin root. `tests/hooks/test-session-start.sh` now asserts the exact command and verifies the wrapper's Cursor JSON output.

## Rigor

- [x] N/A — this change does not modify behavior-shaping skill content.
- [x] The regression test checks the reported failure mode: any return to a relative dispatcher path or removal of the wrapper fails the suite.
- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language).

## Human review

### DIFF
--- a/docs/porting-to-a-new-harness.md
+++ b/docs/porting-to-a-new-harness.md
@@ -387,15 +387,15 @@ own event-matcher strings (Claude Code uses `startup|clear|compact`, Cursor
 The **hook-config schema itself varies per harness** — don't assume the
 Claude Code shape is universal. Compare `hooks/hooks.json` and
 `hooks/hooks-cursor.json`: Cursor's uses
-`"version": 1`, a lowercase `sessionStart` key, a relative
-`./hooks/run-hook.cmd` command, and omits the `matcher`/`type`/`async` fields
+`"version": 1`, a lowercase `sessionStart` key, a
+`"${CURSOR_PLUGIN_ROOT}/hooks/run-hook.cmd" session-start` command, and omits the `matcher`/`type`/`async` fields
 Claude Code uses. Match your `hooks-<harness>.json` to whichever existing file is
 closest, not to a single canonical template.
 
 The hook **command string references a harness-provided plugin-root variable**,
-and its name differs per harness: `hooks.json` uses `${CLAUDE_PLUGIN_ROOT}`,
-`hooks-cursor.json` uses a relative path. Use
-whatever your harness exports. (The `session-start` script re-derives the root
+and its name differs per harness: `hooks.json` uses `${CLAUDE_PLUGIN_ROOT}` and
+`hooks-cursor.json` uses `${CURSOR_PLUGIN_ROOT}`. Use whatever your harness
+exports. (The `session-start` script re-derives the root
 itself via `dirname`, so the script body doesn't depend on this — but the
 command in the manifest does.)
 
--- a/hooks/hooks-cursor.json
+++ b/hooks/hooks-cursor.json
@@ -3,7 +3,7 @@
   "hooks": {
     "sessionStart": [
       {
-        "command": "./hooks/run-hook.cmd session-start"
+        "command": "\"${CURSOR_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start"
       }
     ]
   }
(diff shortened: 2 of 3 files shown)


arittr:
Live-tested exact head `9b37091a92e6fbcd971aba58cb9f35816381fb5f` on Cursor 3.13.21/macOS.

The candidate hook succeeds, but the released relative hook also succeeds: Cursor runs `./hooks/run-hook.cmd session-start` with exit code 0 and accepts its bootstrap response. The PR form, `"${CURSOR_PLUGIN_ROOT}/hooks/run-hook.cmd" session-start`, also exits 0 and returns valid `additional_context`. In both live sessions the agent received the Superpowers bootstrap and entered brainstorming before writing code. The targeted session-start suite passes 7/7 on the PR head.

Therefore this environment establishes compatibility, not the claimed red-to-green fix. We still need evidence from a known-affected Cursor version or the reporter's Windows environment before calling the bug reproduced and fixed. For that test I recommend separate clean profiles for control and treatment, an unrelated empty workspace, and the exact acceptance prompt `Let's make a react todo list`; control should fail to inject the bootstrap and the PR head should inject it and trigger brainstorming before code.

obra:
Closing as not reproduced. The git history you traced is accurate — #709 shipped the env-var absolute form and a later commit swapped it to the relative `./hooks/session-start` — but a fix needs a failure, and nobody has one on record: your own testing was on Codex Desktop rather than Cursor (disclosed, appreciated), and @arittr live-tested current head on Cursor 3.13.21/macOS with the shipped relative form working end-to-end — hook exits 0, bootstrap injects, brainstorming auto-triggers.

The direction isn't obviously an improvement either: env-var-based absolute paths have their own live failure modes (#2091, #2105 both involve harnesses corrupting or failing to substitute plugin-root variables), so 'absolute is more robust' would itself need evidence.

If you or anyone hits an actual Cursor version/OS where the relative form fails to load the bootstrap, please file an issue with the debug log and environment — a real failing case would reopen this question immediately.

Issue #1812, opened 2026-06-19
[RFC] Persistent memory across Superpowers agents


gpitrella:
- [x] I searched existing issues and this has not been proposed before

## What problem does this solve?
When switching between agents or sessions, context doesn't travel with you. Project decisions, preferences, and important details need to be re-explained every time — across sessions and across tools (Claude, Cursor, Codex, Gemini).

## Proposed solution
An optional MCP-based integration that makes context portable across all supported agents. Completely opt-in, zero impact on existing workflows.

I've been experimenting with this approach and would love to discuss before opening a PR.

## What alternatives did you consider?
- Storing context in files manually (fragile, not portable across agents)
- Using CLAUDE.md / agent config files (static, not dynamic)
- MCP-based persistent memory (most flexible, works across all harnesses)

## Is this appropriate for core Superpowers?
Yes — context loss between sessions affects every developer using Superpowers regardless of domain or project type. This would benefit any workflow.

## Environment
| Field | Value |
|-------|-------|
| Superpowers version | v5.1.0 |
| Harness | Claude Code, Cursor |
| Your model | Claude Sonnet |
| All plugins installed | superpowers |

## Context
Happy to share more details if this sounds useful.
"Your skills travel with Superpowers. Your memory should too."

**Update (2 weeks later):** This is now live and working, not just a proposal —

- Listed in the official GitHub MCP Registry
- Listed in ECC (226k+ stars) and PulseMCP
- OAuth 2.1 + AES-256 encryption
- Already works across Claude, Cursor, ChatGPT and VS Code today

Repo: github.com/gpitrella/memxus-remote-mcp

Happy to open a PR directly instead of waiting on discussion — a small, fully optional addition pointing users to an MCP-based memory option, with zero changes to core Superpowers behavior.

— Gabriel

gpitrella:
Re-reading the contribution guidelines and the marketplace README: since new skills aren't generally accepted here, and the marketplace looks curated rather than open to third-party PRs, I don't want to open a PR without a signal from you.
Would a Memxus entry alongside private-journal-mcp be welcome? It targets the cross-harness case (same context in Claude/Cursor/Codex via MCP) rather than replacing the local journaling flow. Listed in the official GitHub MCP Registry, ECC, and PulseMCP if that helps for vetting.
Happy to open the PR if you give the green light, or close this issue if it's not a fit.

hegu-1:
[...]

Issue #1007, opened 2026-03-31
Composing superpowers with team-specific skills


dfoster-oracle:
- [x] I searched existing issues and this has not been proposed before

## What problem does this solve?

I use superpowers alongside team-specific skills (on-call triage, infrastructure runbooks, MCP server configs for internal tools). Both use the same format — SKILL.md with YAML frontmatter — but they install through separate mechanisms and there's no way to compose them. Onboarding a teammate means walking them through two install flows, and there's no declarative way to say "use these superpowers skills plus these team skills."

I've ended up encoding some of your patterns into my own [essentials pack](https://github.com/shrug-labs/packs/tree/main/essentials) — verification-before-completion, code-reviewer, and similar — because I couldn't compose with superpowers directly. I'd rather point at the original than maintain derivatives that drift.

## Proposed solution

I built [aipack](https://github.com/shrug-labs/aipack) to solve this problem — it composes content from multiple sources and syncs it to Claude Code, Codex, Cline, and OpenCode. Superpowers' content already uses the exact format aipack expects. Adding a `pack.json` to the repo root would make superpowers installable alongside other packs:

```json
{
  "schema_version": 1,
  "name": "superpowers",
  "version": "5.0.6",
  "root": "."
}
```

One metadata file, no changes to existing content or distribution. I tested this locally — all 14 skills and the code-reviewer agent rendered to all four harnesses, composed with my team's content. Nothing in superpowers needed modification.

## What alternatives did I consider?

**Wrapper pack in my registry.** I can maintain a pointer to superpowers externally. Works, but version tracking becomes my problem.

**Keep encoding derivatives.** What I do now. It works, it's the wrong approach.

## Is this appropriate for core Superpowers?

Genuinely unsure. It's one metadata file and purely additive — but it is a third-party tool integration. If your preference is that this lives outside the repo, I'll maintain it as a wrapper and that's fine. Raising it here because the format compatibility is already there and the composition use case is real.

## Context

- aipack v0.13.0
- Tested with superpowers v5.0.6
- Harnesses: Claude Code, Codex CLI, Cline, OpenCode

dfoster-oracle:
**Update**: I've shipped content_paths in aipack v0.16.0, which makes this installable from the consumer side without any changes to superpowers: 

```
aipack pack install --url https://github.com/obra/superpowers.git --skills skills --agents agents --name superpowers -q
```

[...]

PR #2068, opened 2026-08-01
fix(using-superpowers): re-check skills mid-workflow, not only at turn start

### LINKED ISSUE #2051: using-superpowers' "check any skill before every action" doesn't hold once a workflow step is underway
- [x] I searched existing issues and this is not a duplicate (checked #1007, #1812, and a few keyword searches — related in spirit, not the same report)

## Environment (required)

| Field | Value |
|-------|-------|
| Superpowers version | 6.2.0 |
| Harness version | 2.1.220 |
| Your model + version | Claude Sonnet 5 (`claude-sonnet-5`) |
| All plugins installed | `superpowers@superpowers-marketplace` (local scope); one additional skill loaded via Claude Code's native `.claude/skills/` project directory — not a plugin, not a manual/unsupported copy, this is the harness's own first-class skill-discovery mechanism |
| OS + shell | Linux (Debian 13, 6.12 kernel), bash |

- [x] I confirmed this issue does not occur without Superpowers installed — with a caveat worth stating plainly: there's no `using-superpowers` bootstrap to test *against* without Superpowers installed, so this isn't a "same bug either way" case. It's specifically about a claim `using-superpowers` itself makes not holding up under test.

I'm not asking for a fix inside core — I read `CLAUDE.md`'s third-party-dependency and project-specific-configuration sections, and a third-party skill's needs don't belong in core. I'm filing this because it's about `using-superpowers`'s own stated behavior, which is squarely in scope, and it seemed like useful, evidence-based feedback either way.

## What happened?

`using-superpowers` states that a relevant skill should be checked "before any action," with "even a 1% chance" of relevance being enough to invoke it — not just once per turn. In a real session, once `brainstorming` was actively engaged, that check did not recur when a second, unrelated skill's trigger condition was clearly met mid-conversation.

## Steps to reproduce

1. Fresh Claude Code session, `superpowers@superpowers-marketplace` installed, plus a second skill (ours — [Keep the Why](https://github.com/oliver-zehentleitner/keep-the-why), a persistence skill with a trigger description matching "a non-trivial change involving... a rejected alternative") loaded via `.claude/skills/`.
2. Prompt: "I want to add rate limiting to `fetch_with_retry` in retry.py so it respects a max requests per second, instead of just exponential backoff on failure."
3. `brainstorming` engages correctly and asks good clarifying questions.
4. Across the conversation, the design settles with an explicit rejected alternative (a token-bucket approach, rejected in favor of a simple sleep-based throttle) and a second non-obvious decision (using `max()`, not summing, when a retry's backoff sleep and the rate-limit spacing overlap).
5. Implementation proceeds. Conversation ends there.

## Expected behavior

Per `using-superpowers`'s own stated policy, the second skill's trigger condition — squarely met by step 4's content — should have been (re-)checked at that point, not only once at the very start of the turn before that content existed.

## Actual behavior

The second skill never activated on its own. Asked directly afterward why not, the agent's own explanation (pasting verbatim, since I think it's the most useful part of this report):

> "Once brainstorming's design questions wrapped up, I was tracking the brainstorming skill's own terminal state... not scanning for *other* skills that might apply to what had just been decided. The using-superpowers rule says to invoke a skill if there's even a 1% chance it's relevant — I applied that at the start of the turn but not again after the design conversation produced exactly the kind of rejected-alternative/rationale content [the other skill] exists to capture."

I then tried fixing this from our own skill's side — added an explicit "re-check after another skill's workflow step concludes" instruction to our skill's body. Retested the identical scenario. No change: the second skill still didn't self-trigger. Asked again, the agent identified why our fix couldn't have worked: the added text lives in the skill *body*, which only enters context once a skill is already triggered — the trigger decision itself runs against the short frontmatter `description`, which our fix never touched. Whatever governs re-triggering mid-conversation, it isn't something a third-party skill can reliably influence from inside its own file.

Happy to share the exact transcripts (both runs) if useful — didn't want to paste ~15 exchanges unprompted into a first report, but they're saved and I can attach or link them however's most useful to you (gist, pastebin, inline).

## Why I'm filing this at all

Not attached to any particular fix, and genuinely unsure whether one exists that wouldn't just be fighting the model in a different spot — prompt-based enforcement may not have a hard guarantee available at all here. But `using-superpowers`'s wording currently reads as a stronger guarantee ("even 1% chance," "not negotiable") than what I could reproduce, and that gap seemed worth naming directly rather than silently working around on our end.

arimu1:
> **This PR MUST target the `dev` branch, not `main`.** `main` is the
> released branch; active work lands on `dev` first. PRs opened against
> `main` will be asked to retarget `dev` before review.

| Field | Value |
|-------|-------|
| Your model + version | Grok 4.5 |
| Harness + version | Grok Build / Cursor |
| All plugins installed | Superpowers skill pack in use for contribution workflow; no third-party skill plugins installed for authoring this change. Skill text edits hand-written against issue #2051 (not auto-generated skill rewrite). |
| Human partner who reviewed this diff | arimu1 |

Real session reported in #2051 (Claude Code + Superpowers 6.2.0, Claude Sonnet 5).

`using-superpowers` states that a relevant skill should be checked before any action, with even a 1% chance of relevance enough to invoke it. In the reported session:

1. `brainstorming` engaged correctly for a rate-limiting design.
2. Mid-conversation, design settled with an explicit **rejected alternative** and a second non-obvious decision — content that matched a second skill's trigger description.
3. Implementation proceeded; the second skill never activated.

When asked why, the agent explained (verbatim from the issue):

> "Once brainstorming's design questions wrapped up, I was tracking the brainstorming skill's own terminal state... not scanning for *other* skills that might apply to what had just been decided. The using-superpowers rule says to invoke a skill if there's even a 1% chance it's relevant — I applied that at the start of the turn but not again after the design conversation produced exactly the kind of rejected-alternative/rationale content [the other skill] exists to capture."

Root instruction gap: the bootstrap forbids skipping skill checks before action, but does not say the check is **continuous**. Agents rationalize "I already checked this turn" / "I'm mid-workflow so skill check is done."

A third-party skill cannot fix this from its own body: trigger decisions run against short frontmatter descriptions; body text only loads after the skill is already chosen. Only `using-superpowers` can restate continuous re-check.

Issue author noted prompt-based enforcement may not hard-guarantee compliance; this PR closes the documented wording gap and the exact rationalizations observed.

## What does this PR change?

Skill-content only in `skills/using-superpowers/SKILL.md`:

1. **The Rule:** state that skill check is continuous (not only at turn start); mid-workflow does not suspend checking other skills; when a second skill's trigger is met, invoke it (respecting Skill Priority).
2. **Red Flags:** two rows closing the observed rationalizations ("already checked this turn"; "mid-workflow — skill check is done").

No harness/hook/code changes. No third-party skill content.

Yes. Continuous skill discovery is core bootstrap behavior for every Superpowers user on every project. It is not domain-specific, not a third-party integration, and not a project-local workflow. Third-party skills cannot fix this from their own files (as #2051 demonstrated).

1. **Only add Red Flags rows** — Leaves The Rule readable as once-per-turn-start. Rejected: the positive rule must state continuity; Red Flags alone are the rationalization counter.
2. **Only strengthen The Rule without Red Flags** — Rejected: #2051's agent produced exact mid-workflow rationalizations; project pattern is to table those (see prior "I know what that means" row for explicit-request skips).
3. **Change skill `description` frontmatter** — Rejected: description is discovery/when-to-load; this skill is session bootstrap. Body is where continuous behavior is enforced. Description already says "before ANY response."
4. **Hook/runtime enforcement of re-scan** — Out of scope for skill-content fix; no portable hard guarantee across harnesses; issue asked about the stated policy gap.
5. **Larger rewrite of bootstrap / flowcharts** — Rejected: bootstrap was deliberately compressed (token cost on every session); keep the minimal wording that blocks the observed failure mode.

No. Single file, single failure mode: skill check treated as turn-start only while mid-workflow.

## Existing PRs

- Related PRs: none found for issue #2051 or for "mid-workflow skill re-check / continuous skill check while one skill is active"

Searched open and closed PRs for: `2051`, `using-superpowers`, `mid-workflow`, `check continuously`, `before every action`. Nearby using-superpowers PRs (#2010 dangling anchor, #1894 Claude Code platform adaptation, harness PRs) do not address continuous re-check. Related issues cited by #2051 (#1007 composition, #1812 memory) are different problems.

## Environment tested

| Grok Build / Cursor | Cursor IDE (current) | Grok 4.5 | Grok 4.5 |

Skill-content-only change; no runtime harness integration. Grounding session is the #2051 transcript (Claude Code / Claude Sonnet 5).

N/A — this PR does not add a new harness.

<details>

```
N/A — not a new harness PR
```

</details>

## Evaluation

- **Initial prompt / source of truth:** Issue #2051 session (Claude Code + Superpowers; brainstorming active; second skill trigger met mid-design via rejected alternative + non-obvious decision; second skill never self-triggered).
- **Before (bootstrap wording + observed agent behavior):**
  - Rule: invoke before any response/action — read as once at turn start
  - No continuous / mid-workflow re-check language
  - Agent tracked active skill's terminal state only; did not re-scan other skill descriptions when new trigger content appeared
- **After (this PR wording):**
  - Explicit: check continuously, not only at turn start
  - Mid-workflow does not suspend other-skill checks
  - Second skill with met trigger: invoke (Skill Priority still applies)
  - Red Flags name the two rationalizations from #2051
- **Adversarial / pressure checks (reasoning + diff review against issue transcript; skill text change only):**
  1. **Exact #2051 case:** After brainstorming produces rejected-alternative content matching another skill's description → continuous re-check + "mid-workflow" red flag forbid treating skill check as done.
  2. **True single-skill turn:** No new triggers mid-turn → no extra invocation required; rule is "when trigger is met," not "invoke everything every message."
  3. **Skill Priority still holds:** Process skills first; continuous check does not invert priority or force parallel exclusive workflows.
  4. **SUBAGENT-STOP unchanged:** Subagents still ignore this skill; no change to that gate.
  5. **Token budget:** +~90 words on a deliberately compressed bootstrap; no flowchart restore; no section restructuring.
  6. **Carefully-tuned content:** Existing Red Flags rows left as-is; two additive rows only. No "human partner" rewording. EXTREMELY-IMPORTANT / 1% block unchanged.
- **Multi-session live harness evals:** Not run in this contribution environment (skill markdown only). Prompt-based enforcement remains soft (as #2051 author noted); this closes the documented gap. Before/after multi-session evals remain welcome as follow-up if maintainers want quantified behavior rates.

## Rigor

- [x] If this is a skills change: I used judgment aligned with superpowers skill philosophy (precise labels, forbid the observed rationalization, keep carefully-tuned sections intact). Full writing-skills multi-subagent campaign was not run in this harness; adversarial cases above map the #2051 failure and adjacent loopholes.
- [x] This change was tested adversarially, not just on the happy path (exact #2051 mid-brainstorming miss, single-skill no-op, Skill Priority, SUBAGENT-STOP)
- [x] I did not modify carefully-tuned content (existing Red Flags rows, rationalizations, "human partner" language) without extensive evals — existing rows left unchanged; only additive continuous-check rule + two new rows matching #2051 rationalizations

## Human review

Fixes #2051

### DIFF
--- a/skills/using-superpowers/SKILL.md
+++ b/skills/using-superpowers/SKILL.md
@@ -19,6 +19,8 @@ This is not negotiable. You cannot rationalize your way out of this.
 
 **Invoke relevant or requested skills BEFORE any response or action** — including clarifying questions, exploring the codebase, or checking files. If it turns out wrong for the situation, you don't have to use it.
 
+**Check continuously — not only at turn start.** New decisions, facts, or rejected alternatives mid-conversation can make other skills relevant. Being inside one skill's workflow does not suspend checking for others. When a second skill's trigger is met, invoke it (respecting Skill Priority); do not wait until the active skill fully finishes if the new trigger already applies.
+
 **Before entering plan mode:** if you haven't already brainstormed, invoke the brainstorming skill first.
 
 Then announce "Using [skill] to [purpose]" and follow the skill exactly. If it has a checklist, create a todo per item.
@@ -48,6 +50,8 @@ These thoughts mean STOP—you're rationalizing:
 | "I'll just do this one thing first" | Check BEFORE doing anything. |
 | "This feels productive" | Undisciplined action wastes time. Skills prevent this. |
 | "I know what that means" | Knowing the concept ≠ using the skill. Invoke it. |
+| "I already checked skills this turn" | New content creates new triggers. Re-check. |
+| "I'm mid-workflow — skill check is done" | Active skill ≠ exclusive. Other skills still apply. |
 
 ## Platform Adaptation
 


obra:
Closing this one. `skills/using-superpowers/SKILL.md` is the most heavily-tuned file in the repo — it's injected into every session, so every added word costs tokens on every session start, and changes to it need eval evidence showing they improve behavior. Your own Rigor section acknowledges the writing-skills multi-subagent campaign wasn't run.

The underlying issue (#2051) also doesn't quite support a core change: it's about a third-party skill not re-triggering mid-workflow, the reporter's own analysis is that prompt text can't strongly enforce this, and they explicitly weren't requesting a core fix.

If you run the eval campaign (superpowers:writing-skills, before/after across multiple sessions) and the numbers show this text changes mid-workflow re-checking behavior, we'd be glad to look at a resubmission with that evidence attached.

PR #2070, opened 2026-08-01
Make reviewers grep for incomplete sweeps and missed call sites

kbarrycrunch:
A diff cannot show an absence. Two failure modes get through a diff-only task review looking complete:

- **Incomplete sweeps.** A string, label, icon, colour or constant changed in three of four places. Every hunk in the diff is correct; the product is wrong.
- **Missed call sites.** A signature change, a new required parameter, or a new case in something built per-surface — where a sibling implementation was missed. When those siblings no longer share a builder, there is no compiler error and no failing test to reveal it.

Both are arguably already covered by the template's *"inspect code outside the diff only to evaluate a concrete risk you can name"*. In practice they weren't getting named, because spotting them requires already suspecting them. This adds them to the list of what counts as a named risk, so each gets one grep.

I deliberately kept the existing scoping rather than replacing it with a blanket "always grep" — the *"do not crawl the broader codebase … one focused check per named risk"* constraint is what keeps reviews cheap, and it's still intact.

`re-review-prompt.md` gets the narrower half only. It's deliberately scoped to the fix diff, but a "this value is wrong" finding is exactly where a partial fix verdicts as ADDRESSED — so verifying *that finding* means grepping for the old value. Framed as verifying the finding, not as re-reviewing.

Docs only; no scripts or behaviour touched.

### DIFF
--- a/skills/subagent-driven-development/re-review-prompt.md
+++ b/skills/subagent-driven-development/re-review-prompt.md
@@ -52,6 +52,14 @@ Subagent (general-purpose):
     does not block this task and does not extend the loop. A broad
     whole-branch review happens after all tasks are complete.
 
+    One check does reach outside the fix diff, because a diff cannot show an
+    absence: if a finding was "this value/string/call site is wrong" and the
+    fix changed it, grep the source tree for the OLD value and confirm no
+    other instance survives. A fix that corrected three of four occurrences
+    verdicts as ADDRESSED on the diff alone and is still broken in the
+    product. This is verifying the finding, not re-reviewing — report the
+    grep and its hits alongside that finding's verdict.
+
     ## Tests
 
     The implementer re-ran the tests covering the amended code and appended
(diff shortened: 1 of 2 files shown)


obra:
Closing this one. The idea — having reviewers grep for the old value after a sweep or signature change — is plausible, but this PR doesn't clear the bar for modifying behavior-shaping content:

- It adds ~21 lines to SDD's reviewer prompts (`task-reviewer-prompt.md`, `re-review-prompt.md`) with no before/after eval evidence. CLAUDE.md's "Skill changes require evaluation" section applies to exactly these files.
- No concrete session is cited where the current prompts missed a sweep — "in practice they weren't getting named" isn't a problem statement we can verify. The PR body itself concedes the existing named-risk scoping arguably already covers this.
- The PR template wasn't filled (no submitter identification, no environment table), and it targets `main` instead of `dev`.

If you actually hit this failure in a real session, please file an issue with the transcript — that's the right starting point, and the fix would then need eval evidence per `superpowers:writing-skills`.

PR #1588, opened 2026-05-21
fix: keep SDD writes inside active worktree

YOMXXX:
## What problem are you trying to solve?

Fixes #1040.

`using-git-worktrees` creates or detects an isolated worktree, but the current workflow does not preserve that write boundary through the rest of implementation. In the reported failure mode, exploration/research agents return absolute paths from the parent checkout, and the controller or implementer passes those paths directly to write/edit/git commands. That silently writes changes into the main checkout or another worktree instead of the active implementation worktree.

This breaks the core promise of the worktree workflow: protecting the user's current branch from implementation changes.

## What does this PR change?

- Adds an explicit active workspace root contract to `using-git-worktrees` via `WORKTREE_ROOT=$(git rev-parse --show-toplevel)`.
- Threads that active root through `subagent-driven-development` and its implementer/spec-reviewer/code-reviewer prompt templates.
- Adds a static regression test ensuring SDD keeps implementer and reviewer work inside the active worktree.

## Is this change appropriate for the core library?

Yes. This is core workflow safety for all projects that use Superpowers worktrees and SDD. It is not project-specific, domain-specific, or tied to a third-party service. The change reinforces an existing core behavior: implementation work should stay inside the isolated worktree that Superpowers creates or detects.

## What alternatives did you consider?

1. Add hook-based hard enforcement around file writes. That would be stronger, but it would broaden this PR into platform-specific hook behavior and create a larger review surface.
2. Only update `using-git-worktrees`. That would leave SDD implementer and reviewer prompts without the active root, which is where the issue reproduces when paths are passed between agents.
3. Rely on "Work from: [directory]" alone. That text existed already and was not explicit enough to prevent stale absolute paths from being reused.

This PR keeps the fix small: establish the boundary once and make every SDD subagent prompt carry it.

## Does this PR contain multiple unrelated changes?

No. All changes are tied to one problem: ensuring implementation and review work stays inside the active worktree root.

## Existing PRs

- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related PRs: #1121, #1153, #1167, #170

#1121 established the native-tool-first worktree flow. #1153 and #1167 covered earlier worktree workflow adjustments. #170 was a broader closed safety-gates PR. I did not find an open or closed PR that specifically fixes #1040 by carrying an active workspace root through SDD implementer and reviewer prompts.

## Environment tested

| Harness (e.g. Claude Code, Cursor) | Harness version | Model | Model version/ID |
[...]

PR #2082, opened 2026-08-04
fix(sdd): guard worktree commits in prompts and review-package (#2050)

### LINKED ISSUE #2050: subagent-driven-development: dispatched subagent committed to main instead of the worktree (isolation not enforced)
## Summary

While executing an implementation plan with **subagent-driven-development** inside a git worktree created via the native `EnterWorktree` tool, one dispatched implementer subagent ran its `git add` / `git commit` against the **main checkout / `main` branch** instead of the isolated worktree branch. The worktree did not "contain" the subagent, and `main` was mutated with an out-of-context commit.

The worktree existed precisely to isolate the work. That guarantee silently did not hold for a subagent.

## Environment

- Claude Code on Windows (git-bash tool), multi-package repo.
- Worktree created with the native `EnterWorktree` tool (path under `.claude/worktrees/<name>`), session cwd switched into it. The `using-git-worktrees` skill correctly deferred to the native tool.
- Plan executed with `subagent-driven-development`: controller dispatches a fresh implementer subagent per task via the Agent tool; each subagent commits its own task.

## What happened

- ~14 tasks committed correctly to the worktree branch (implementers run on a mid-tier model).
- One task (the first one dispatched on the cheapest model tier) committed to the **main checkout** — creating a commit on local `main`, not on the worktree branch. The worktree branch never received that task's work.
- Because `main` had meanwhile advanced (an unrelated PR merged), the stray commit landed on top of a `main` that contained **none** of the prior task's feature code — i.e. `main` briefly carried orphaned, half-built feature code. Had `main` later been pushed, that partial work would have leaked to the remote.

## How it was detected (important: the review gate did NOT catch it directly)

The controller generates a per-task diff (`review-package BASE HEAD`) for the reviewer. For this task the diff showed the earlier feature files being **deleted** — impossible for a purely additive task. That anomaly is what exposed the divergence. Nothing in the normal flow asserts "the subagent committed to the expected branch," so without that incidental diff anomaly it could have gone unnoticed.

## Root cause (best understanding, stated with uncertainty)

The subagent's working directory resolved to the **main repository root** rather than the active worktree, so its git commands targeted `main`. I could not determine why only that one subagent diverged while the others (before and after it, on a different model) inherited the worktree cwd correctly. The underlying behavior — **Agent-tool subagents not being reliably pinned to the session's active `EnterWorktree` directory** — is a Claude Code harness concern rather than a superpowers bug per se. But subagent-driven-development is the workflow that dispatches git-committing subagents into a worktree, so it is well positioned to defend against it.

## Reproduction (sketch)

1. `EnterWorktree` to create/enter an isolated worktree in a repo whose `main` checkout is a separate working tree sharing the same `.git`.
2. Run `subagent-driven-development` over a multi-task plan; each dispatched subagent does its own `git add` / `git commit`.
3. Observe that at least one subagent may commit to the branch checked out in the **main** working tree rather than the worktree branch. (Intermittent in my run — 1 of ~17 tasks; the diverging one happened to be the first on the cheapest model tier and coincided with an unrelated merge advancing `main`.)

Reliability of repro is unknown given the intermittency; the harness cwd behavior is the thing to pin down.

## Suggested mitigations (superpowers side)

1. **Bake a worktree/branch guard into the implementer + fix dispatch prompt templates.** Require the subagent, before any `git add` / `git commit`, to verify:
   - `git rev-parse --show-toplevel` ends with the expected worktree path, and
   - `git branch --show-current` equals the expected feature branch,
   and to `cd` into the worktree (or abort) otherwise. Adding exactly this guard to my dispatches eliminated all recurrences for the remaining ~13 tasks and fix passes.
2. **Controller-side post-commit assertion.** After each task, verify the new commit advances the recorded per-task BASE on the expected branch (the skill already records BASE for `review-package`; it can also assert `git branch --contains <newsha>` includes the feature branch). This catches a wrong-branch commit deterministically instead of relying on a diff anomaly.
3. Pass the absolute worktree path explicitly in dispatch prompts rather than assuming inherited cwd.

## Impact

No data lost in my case — the stray commit was unpushed; I cherry-picked it onto the worktree branch and reset local `main` to `origin/main`. But the isolation the user explicitly requested was not enforced by the tooling; it held only because the controller happened to be checking. A less defended run could silently accrete work onto `main`.

Happy to provide more detail. Filed from a real subagent-driven run (superpowers v6.0.3).

stantheman0128:

| Field | Value |
|-------|-------|
| Your model + version | Grok (via Cursor Agent) |
| Harness + version | Cursor IDE agent |
| All plugins installed | OSS factory workspace; superpowers skills read from upstream clone |
| Human partner who reviewed this diff | Stan Shih (stantheman0128) reviewed the complete diff before publish |

AI-assisted contribution. Evidence and eval notes below are from a real Windows 11 session.

Issue #2050: during subagent-driven-development (SDD), a dispatched implementer ran `git add` / `git commit` from the main repository checkout instead of the isolated worktree branch. The worktree was meant to isolate the work, but the subagent cwd was not enforced, so `main` received an out-of-context commit and the review package still looked plausible.

Failure mode: implementer reports a commit SHA; `review-package` builds a diff from BASE..HEAD even when HEAD lives on another branch or working tree, so the controller reviews unrelated history as if it were the task.

## What does this PR change?

1. `implementer-prompt.md`: require verifying `git rev-parse --show-toplevel` and `git branch --show-current` before every `git add` / `git commit`; STOP with BLOCKED if they do not match the dispatched worktree.
2. `review-package`: exit 3 when HEAD is not reachable from the checked-out branch, or when BASE..HEAD is empty (commit landed elsewhere).
3. `test-sdd-workspace.sh`: four regression cases for stray-tree HEAD, empty range, happy path, and amend orphaning FIX_BASE.

Deliberately **not** changing `SKILL.md` prose (#1588 was closed for blast radius on that file; obra asked for the smallest surface, likely `implementer-prompt.md` only). The mechanical guard lives in `review-package`, which every task review already calls.

Yes. Any SDD user with multiple working trees of the same repo hits this; it is not project-specific.

- Broad `SKILL.md` guardrails (rejected: #1588 closed for touching behavior-shaping prose without evals).
- Controller-only cwd checks (insufficient: subagents do not inherit cwd reliably per #2050).
- Only prompt text without `review-package` enforcement (insufficient: prompts alone did not stop the reported failure).

No. All three files serve the same worktree isolation failure mode.

## Existing PRs
- Related PRs: #1588 (closed, SKILL.md blast radius), #1589 (closed, EnterWorktree bare-parent guard)

#1588/#1589 addressed adjacent git-worktree hazards. This PR targets the SDD implementer dispatch path described in #2050: subagent commits on the wrong branch/tree and review-package still packages plausible output.

## Environment tested

| Cursor Agent | current (2026-08) | Grok | via Cursor |

## Evaluation

- Initial prompt: factory scout on obra/superpowers #2050 after maintainer discussion on worktree isolation.
- Eval sessions after the change: multiple local A/B runs on Git Bash (Windows 11).
- Before: stray-commit scenarios produced exit 0 review packages with unrelated history (2 commits / hundreds of bytes of wrong diff in one repro).
- After: both stray-commit cases exit 3 with `git branch --contains` guidance; in-worktree commits and amend orphan case still exit 0.

## Rigor

## Human review

## Verification / Evidence

```text
# Git Bash on Windows 11, from repo root:
bash tests/claude-code/test-sdd-workspace.sh

# #2050-related cases (all PASS):
#   review-package rejects a HEAD committed in another working tree
#   review-package rejects an empty range: nothing landed on this branch
#   review-package accepts commits made in this worktree
#   review-package accepts a FIX_BASE orphaned by an amend

# Four pre-existing workspace-path assertions still fail on this Windows host
# (prints <repo-root>/.superpowers/sdd/<plan-basename>, linked worktree workspace).
# Those failures reproduce on origin/dev without this branch; not introduced here.
```

Linked-worktree live repro: commit on branch A while review runs in worktree B -> exit 3 before writing a misleading package.

## What was not tested

- Full Claude Code `EnterWorktree` dispatch on macOS/Linux (mechanism is plain git ancestry checks; platform-agnostic).
- End-to-end multi-subagent SDD session in Claude Code harness (local script regression + manual linked-worktree repro only).

Fixes #2050

### DIFF
--- a/skills/subagent-driven-development/implementer-prompt.md
+++ b/skills/subagent-driven-development/implementer-prompt.md
@@ -40,6 +40,19 @@ Subagent (general-purpose):
     6. Report back
 
     Work from: [directory]
+    Expected git toplevel: [WORKTREE_TOP]
+    Expected branch: [EXPECTED_BRANCH]
+
+    ## Git worktree guard (before any git add or git commit)
+
+    Agent-tool subagents do not always inherit the controller's active worktree
+    cwd. Before every `git add` or `git commit`, verify:
+    - `git rev-parse --show-toplevel` equals [WORKTREE_TOP]
+    - `git branch --show-current` equals [EXPECTED_BRANCH]
+
+    If either check fails, STOP with status BLOCKED. Report the paths and branch
+    you see. Never commit from the main repository checkout when this task runs
+    in a worktree.
 
     **While you work:** If you encounter something unexpected or unclear, **ask questions**.
     It's always OK to pause and clarify. Don't guess or make assumptions.
--- a/skills/subagent-driven-development/scripts/review-package
+++ b/skills/subagent-driven-development/scripts/review-package
@@ -22,6 +22,29 @@ head=$3
 git rev-parse --verify --quiet "$base" >/dev/null || { echo "bad BASE: $base" >&2; exit 2; }
 git rev-parse --verify --quiet "$head" >/dev/null || { echo "bad HEAD: $head" >&2; exit 2; }
 
+# The reviewed commits must be on the branch checked out here. A subagent
+# whose cwd resolved to another working tree of this repository commits onto
+# the branch checked out there, and the shared .git makes that commit
+# resolvable from here: without these two checks the package builds, renders
+# this branch's earlier work as deletions, and reads as plausible.
+branch=$(git rev-parse --abbrev-ref HEAD)
+root=$(git rev-parse --show-toplevel)
+
+if ! git merge-base --is-ancestor "$head" HEAD; then
+  echo "HEAD $head is not reachable from $branch in $root." >&2
+  echo "It was committed from a different working tree, so BASE..HEAD would diff two unrelated histories." >&2
+  echo "Run 'git branch --contains $head' to find where the work landed, and move it onto $branch before reviewing." >&2
+  exit 3
+fi
+
+commits=$(git rev-list --count "${base}..${head}")
+if [ "$commits" -eq 0 ]; then
+  echo "no commits in ${base}..${head}: nothing to review on $branch in $root." >&2
+  echo "If the implementer reported commits, they landed on another branch or working tree." >&2
+  echo "Run 'git branch --contains <reported sha>' before treating the task as done." >&2
+  exit 3
+fi
+
 if [ $# -eq 4 ]; then
   out=$4
 else
@@ -42,5 +65,4 @@ fi
   git diff -U10 "${base}..${head}"
 } > "$out"
 
-commits=$(git rev-list --count "${base}..${head}")
 echo "wrote ${out}: ${commits} commit(s), $(wc -c < "$out" | tr -d ' ') bytes"
--- a/tests/claude-code/test-sdd-workspace.sh
+++ b/tests/claude-code/test-sdd-workspace.sh
@@ -189,6 +189,62 @@ PLAN
         echo "    status: $wt_status"
     fi
 
+    # --- review-package refuses a range whose commits are not on this
+    # checkout's branch: a subagent that committed from the parent checkout
+    # otherwise produces a plausible-looking package (issue #2050) ---
+    local wt_base stray
+    ( cd "$wt" && printf 'feature\n' > feature.txt && git add feature.txt \
+        && git "${git_id[@]}" commit -qm "task 1: feature" )
+    wt_base="$(cd "$wt" && git rev-parse HEAD)"
+
+    ( cd "$repo" && printf 'task 2\n' > task2.txt && git add task2.txt \
+        && git "${git_id[@]}" commit -qm "task 2: committed in the parent checkout" )
+    stray="$(cd "$repo" && git rev-parse HEAD)"
+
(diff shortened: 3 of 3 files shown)


obra:
Closing, with the design direction affirmed: the `review-package` half of this — `merge-base --is-ancestor` and `rev-list --count` guards with a distinct exit code — is exactly the smallest-surface mechanical enforcement discussed on #1588, and that guard will land in-house, tracked on #2050. Your disclosure was honest throughout (origin, Windows-only testing, untested e2e paths), which is appreciated and noted.

Why not this PR: it bundles two different kinds of change. The script guard is mechanical and verifiable; the `implementer-prompt.md` half is behavior-shaping content submitted without the eval evidence CLAUDE.md requires (the rigor box is unchecked on that point by your own admission). Prompt wording changes to SDD only move with before/after evals, and splitting the mechanical fix out cleanly is easiest done on our side.

If you hit the wrong-branch-commit failure in a real session (rather than a scouted one), transcripts on #2050 sharpen the priority.

PR #2085, opened 2026-08-04
fix(brainstorm): keep visual companion alive under Cursor

peterjoles:
## Problem
On Cursor, `start-server.sh`’s default `nohup`/`disown` path reports `server-started`, then the detached Node process is reaped when the Shell tool call exits. The user’s browser hits “This site can’t be reached” / connection refused on first open. Codex and Windows already auto-foreground for the same class of reaper; Cursor did not, despite a clear env signal (`CURSOR_AGENT=1`).

## Solution
Auto-detect Cursor agent environments and force foreground mode (same pattern as `CODEX_CI` / Windows). Require agents to keep the waiting shell alive (`block_until_ms: 0`), discover via `server-info`, and gate on HTTP 200 before sharing the URL. Also add an HTTP readiness check on the background path and tee foreground logs into `server.log`.

## Key changes
- Auto-foreground when `CURSOR_AGENT` is set or `CURSOR_EXTENSION_HOST_ROLE=agent-exec` (honors `--background` override)
- Cursor stderr hint pointing at async Shell + `server-info` + curl
- Background-mode HTTP readiness probe against the companion URL before returning success
- Foreground mode tees Node output to `server.log`
- Docs: Cursor launch section + readiness gate in The Loop
- Tests for Cursor auto-detect, role-based detect, hint, and `--background` override

## Risks and rollout
- Cursor agents that launch without an async/persistent shell will block on `wait` (expected; documented). Incorrect launches fail closed rather than handing the user a dead URL.
- HTTP readiness depends on `curl` when present; if `curl` is missing the check is skipped (same as prior “assume up” behavior for that edge case).

## Tests
- Extended `tests/brainstorm-server/start-server.test.sh` (8 passed)
- Manual smoke: foreground start under `CURSOR_AGENT`, `curl` of full `?key=` URL returned HTTP 200

## Stack
- Position: Standalone
- Depends on: none
- Ultimate target: `main` (obra/superpowers)

## Test plan
- [ ] Review Cursor auto-detect + `--background` override tests
- [ ] On Cursor: start companion with async Shell; confirm first browser open works without “can’t be reached”
- [ ] Confirm Claude Code / default background path still returns `server-started` and passes HTTP readiness when `curl` is available

### DIFF
--- a/skills/brainstorming/scripts/start-server.sh
+++ b/skills/brainstorming/scripts/start-server.sh
@@ -15,7 +15,8 @@
 #   --open                Auto-open the browser on the first screen (use only
 #                         after the user approves the visual companion).
 #   --foreground          Run server in the current terminal (no backgrounding).
-#   --background          Force background mode (overrides Codex auto-foreground).
+#   --background          Force background mode (overrides Codex/Cursor/Windows
+#                         auto-foreground).
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 
@@ -94,8 +95,29 @@ is_windows_like_shell() {
   return 1
 }
 
+is_cursor_agent() {
+  # Cursor sets these in agent Shell tool environments. Detached/nohup children
+  # are reaped when the launching tool call exits — same failure mode as Codex.
+  if [[ -n "${CURSOR_AGENT:-}" ]]; then
+    return 0
+  fi
+  if [[ "${CURSOR_EXTENSION_HOST_ROLE:-}" == "agent-exec" ]]; then
+    return 0
+  fi
+  return 1
+}
+
+foreground_retry_hint() {
+  echo "$SCRIPT_DIR/start-server.sh${PROJECT_DIR:+ --project-dir $PROJECT_DIR} --host $BIND_HOST --url-host $URL_HOST --foreground"
+}
+
 # Some environments reap detached/background processes. Auto-foreground when detected.
 if [[ -n "${CODEX_CI:-}" && "$FOREGROUND" != "true" && "$FORCE_BACKGROUND" != "true" ]]; then
+  FOREGROUND="true"
+fi
+
+# Cursor reaps nohup children after the Shell tool call completes. Auto-foreground.
+if [[ "$FOREGROUND" != "true" && "$FORCE_BACKGROUND" != "true" ]] && is_cursor_agent; then
   FOREGROUND="true"
 fi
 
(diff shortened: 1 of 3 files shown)


obra:
Hi, could you please retarget this against dev and have your agent fill out our PR template?

arittr:
## Maintainer test disposition: READY TO MERGE

Validated rebased head `1b01de6ae6665163ebdc48e15433e21fcae7187d` against `dev`.

- **Automated:** `cd tests/brainstorm-server && npm test` — **138 passed, 0 failed**.
- **Cursor Desktop:** loaded the exact PR plugin; launched once with `CURSOR_AGENT=1`, Shell `block_until_ms: 0`, and no explicit foreground/background override. The browser auto-opened, showed **Connected**, the same keyed URL returned HTTP 200 at 106 seconds, and a real browser choice/click reached the server.
- **Codex CLI 0.146.0:** launched once with `CODEX_CI=1` and no explicit mode override. Auto-foreground was observed (the launcher remained the server's parent); the same PID and keyed URL returned HTTP 200 at 112 seconds while the original tool session remained alive.
- **Claude Code 2.1.224:** tested in a clean interactive CLI with all inherited `CODEX_*` variables removed and no explicit mode override. The unchanged default background path was observed (launcher exited, server detached, owner PID pointed to the live Claude CLI); the same PID and keyed URL returned HTTP 200 at 87 seconds between turns.
- **Additional launcher audit:** ShellCheck and Bash syntax passed on both changed shell files. A real server with forced HTTP-readiness failure returned exit 1, printed the foreground retry command, and stopped its exact child process.
- **Lifecycle controls on macOS:** `windows-lifecycle.test.sh` completed **9 passed, 0 failed, 3 skipped**. The three skipped checks require an actual Windows/MSYS2 host.

All test servers and temporary harness sessions were stopped; the worktree is clean.

One wording precision note for the record: the evidence directly demonstrates the companion owner's lifecycle watchdog reacting to owner-process liveness. It does not independently prove Cursor's internal reaping mechanism.

### Residual risk

A real Windows/Git Bash smoke was not run. I consider this **non-blocking** because the Windows detection and owner-PID-clearing logic are unchanged, while the modified generic foreground path was exercised live in both Cursor and Codex. Still, this PR should not be described as fully live-tested on Windows.

**Disposition: approved from the maintainer test side and ready to merge to `dev`; live Windows/MSYS2 is the only remaining optional confidence check.**

PR #1738, opened 2026-06-11
fix(skills): SDD top-tier dispatches inherit the session model instead of naming one from recall

snvtac:
> Targets `dev`. Complementary to #1719 — different mechanism, independently mergeable. If maintainers prefer a single PR, this paragraph can fold into #1719 and this one can close.

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Claude Fable 5 (`claude-fable-5`) |
| Harness + version | Claude Code 2.1.173 (Linux) |
| All plugins installed | superpowers 5.1.0 (claude-plugins-official, f2cbfbe), oh-my-claudecode, vercel, supabase |
| Human partner who reviewed this diff | @snvtac — experienced the session failure, directed the change, read the complete diff (the single paragraph quoted below) before submission |

## What problem are you trying to solve?

In a real session (Claude Code 2.1.173, session model `claude-fable-5`, superpowers 5.1.0), the `subagent-driven-development` controller dispatched review subagents with `model: opus` — a silent downgrade from the session's Fable 5. My human partner reported exactly this on #1719: "I default to the fable 5 model, but during task execution it switches to Opus 4.8." The root cause is the one #1719 diagnoses: "use the most capable available model" gets resolved from training-data recall, and recall goes stale the moment a new frontier model ships.

## What does this PR change?

Adds one paragraph to `skills/subagent-driven-development/SKILL.md` → Model Selection, directly after the "Architecture, design, and review tasks" line. The complete diff:

> **Don't pick "most capable" from memory — inherit instead.** Your knowledge of the model lineup goes stale the moment a newer frontier model ships, so a model name recalled from training data silently downgrades the dispatch. When your dispatch tool takes an optional model parameter, omit it for these tasks: the subagent inherits the session's model — the one your human partner already chose — and inheritance can never resolve to a stale name. Pass an explicit model only to route mechanical or integration tasks to a cheaper tier, and take the name from what the harness offers, not from recall.

## Is this change appropriate for the core library?

Yes. General-purpose and harness-agnostic ("when your dispatch tool takes an optional model parameter"): in Claude Code the Task tool documents omit-as-inherit ("If omitted, … inherits from the parent"); harnesses with the same affordance get the same benefit; harnesses without it fall through to the existing tier guidance unchanged. No third-party tools or services.

## What alternatives did you consider?

[...]

PR #1719, opened 2026-06-10
fix(skills): account for Claude Fable in model selection and testing guidance

svenbledt:
## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Claude Fable 5 (`claude-fable-5`) |
| Harness + version | Claude Code 2.1.170 (Windows 11) |
| All plugins installed | superpowers 5.1.0, context7, github, code-review, feature-dev, frontend-design, skill-creator, caveman, plus buddy/gitkraken/coolify MCP servers |
| Human partner who reviewed this diff | @svenbledt directed the change and reported the session failure; full-diff review pending — PR is a **draft** until that review happens |

## What problem are you trying to solve?

Two faces of the same root cause: the plugin's model guidance predates Claude Fable, and agents fill the gap with stale training knowledge.

1. **Observed session failure (the motivating problem):** `subagent-driven-development`'s Model Selection section says architecture/design/review tasks should use "the most capable available model." In a real Claude Code session, the dispatching agent resolved that phrase from its training data and chose **Opus 4.8** — even though Fable 5, the more capable model, was available in the harness. The tier language is deliberately model-name-agnostic, but it silently depends on the agent knowing the current lineup, which it doesn't: training cutoffs guarantee the agent's idea of "most capable" goes stale the moment a new model ships. Baseline testing (below) reproduced this failure in **6 out of 6** scenario runs.
2. `skills/writing-skills/anthropic-best-practices.md` instructs skill authors to test with all models they plan to use, but its per-model list (Haiku/Sonnet/Opus) and checklist omit Fable, so authors running on Fable get guidance that doesn't cover their own model.

## What does this PR change?

- `skills/subagent-driven-development/SKILL.md`: adds one paragraph to Model Selection — don't resolve "most capable" from memory; check what the harness actually offers, with Fable named as the current example of a frontier model that postdates training data. The tier structure is unchanged.
- `skills/writing-skills/anthropic-best-practices.md`: adds a Fable bullet to the per-model testing considerations and adds Fable to the testing checklist.

## Is this change appropriate for the core library?

Yes. Both files are general-purpose guidance that applies to every superpowers user regardless of project or domain. No third-party tools or services involved.

## What alternatives did you consider?

- **Hard-coding model names into the tier list** (cheap=Haiku, standard=Sonnet, most capable=Fable) — rejected: it would destroy the section's deliberately model-agnostic design and rot just as fast as the knowledge it replaces.
[...]

PR #2097, opened 2026-08-06
fix(sdd): resolve "most capable available" to the strict maximum for the final whole-branch review

### LINKED ISSUE #2096: subagent-driven-development: "most capable available model" for the final review gets under-resolved when the session model is above Opus
## Summary

In `subagent-driven-development`'s Model Selection section, the rule for the final whole-branch review — "dispatch it on the most capable available model, not the session default" — got under-resolved in practice: running on a session model *above* Opus (Claude Fable 5, a tier above Opus in the dispatchable model list), the agent picked **Opus** for the final review instead of the strictly most capable available model, and only surfaced this when the human asked why.

## What happened

- Environment: Claude Code, session model `claude-fable-5` (dispatchable models for subagents: `haiku` / `sonnet` / `opus` / `fable`, with `fable` above `opus`), superpowers 6.2.0.
- Executing a plan via `superpowers:subagent-driven-development`. Task-scoped reviews ran on sonnet (consistent with the cost-scaling guidance). For the final whole-branch review, the skill says:

  > **Architecture and design tasks**: use the most capable available model. The final whole-branch review is one of these — dispatch it on the most capable available model, not the session default.

- The agent dispatched the final review on **opus**, reasoning (verbatim from its own retrospective): *"I interpreted it as the top standard tier (= Opus), treating the session model (fable) as something separate, and leaned toward conserving cost."*

## Why the wording invites this failure

Two aspects of the section combine against the intended reading when the session model is itself the top model:

1. **Cost-priming dominates the section.** The section opens with "Use the least powerful model that can handle each role to conserve cost" and reinforces it repeatedly ("cheap model", "Turn count beats token price", "not the most capable model" for small diffs). The final-review sentence is the single exception, and an agent already in cost-conservation mode rationalizes the exception downward.
2. **"not the session default" reads as "not the session model."** The phrase exists to prevent *silent inheritance by omission* (as the neighboring paragraph explains). But when the session model IS the most capable available model, "not the session default" can be parsed as actively steering away from it — so the agent picks the next tier down and feels compliant.

There is no explicit resolution rule for "available", so "most capable available" degraded into "most capable tier I'd normally consider."

## Impact

The final review still ran one tier below what was available. In this session the Opus review did catch 3 Important findings, so no proven quality loss — but the whole point of the final-review rule (one maximally capable pass before merge) was silently weakened, and it only came to light because the human happened to ask which model was used.

## Suggested fix

Make the final-review rule self-contained and mechanical, e.g.:

> The final whole-branch review is one of these — dispatch it on the **strictly most capable model your harness can dispatch** (compare against the full model list, including the session's own model and anything above Opus). "Not the session default" means *never rely on inheritance by omission* — pass the model explicitly even when the explicit choice equals the session model.

Optionally add a red-flag row to the table: | "The session model is expensive, the top standard tier is close enough" | The final review is the one place cost-scaling does not apply. Pick the strict maximum. |

---

🤖 Filed with [Claude Code](https://claude.com/claude-code) on behalf of the user, based on the agent's own observed behavior in a real session.

RyosukeNishi:
Closes #2096.

| Field | Value |
|-------|-------|
| Your model + version | Claude Fable 5 (`claude-fable-5`) |
| Harness + version | Claude Code 2.1.223 (macOS) |
| All plugins installed | superpowers 6.2.0 (claude-plugins-official), slack, mattpocock-skills |
| Human partner who reviewed this diff | @RyosukeNishi — experienced and filed the motivating session failure (#2096) and directed this PR; full-diff review pending, so this PR is a **draft** until that review happens |

Issue #2096, grounded in a real session (Claude Code, session model `claude-fable-5`, superpowers 6.2.0): executing a plan via `subagent-driven-development`, the controller dispatched the final whole-branch review on **opus** even though fable — a tier above opus — was in the harness's dispatchable list. Its own retrospective: *"I interpreted it as the top standard tier (= Opus), treating the session model (fable) as something separate, and leaned toward conserving cost."* The downgrade was silent; it only surfaced because the human asked which model ran the review.

Two aspects of the current wording combine against the intended reading when the session model is itself the top model:

1. The section is (correctly) cost-primed throughout, and the final-review sentence is the single exception — an agent in cost-conservation mode rationalizes the exception downward.
2. "not the session default" exists to forbid inheritance-by-omission, but when the session model IS the most capable model it parses as actively steering away from it, so the agent picks the next tier down and feels compliant.

Baseline micro-testing (below) reproduced the failure **5/5** under a faithful scenario.

## What does this PR change?

Two hunks in `skills/subagent-driven-development/SKILL.md`, both implementing the same rule fix: (1) the final-review sentence in Model Selection now resolves "most capable available" mechanically — rank every model you can pass when dispatching, including the session's own model and any tier above the usual top tier, and pick the strict maximum — and states that explicitness rules out inheritance by omission, not the session's model itself; (2) a matching row in Common Rationalizations countering the cost-based downgrade.

Yes. It edits an existing core skill's existing rule, stays model-name-agnostic and harness-agnostic (no lineup to go stale), and applies to any user whose session model sits above the tier the dispatching agent considers "the usual top".

- **Keep the sentence, add a parenthetical** ("…most capable available model (including the session's model)"): rejected — it leaves "not the session default" in place, which baseline transcripts show is the misparse vector (one rep reasoned verbatim that the phrase "clearly distinguishes the session default from the top model").
- **Name concrete models/tiers**: rejected — the section is deliberately model-name-agnostic, and naming a lineup goes stale at every model release (the root cause #1719 diagnoses).
- **Do nothing** (clean-context fable controllers resolve it fine, see evals): rejected — the failure reproduces 5/5 for a controller that is not confident of the session model's rank, and it reproduced in at least one real session with fable itself under accumulated cost pressure.

No. Both hunks encode the same rule; the rationalization row is the table-side counter for the exact excuse observed in the baseline runs.

## Existing PRs

- Related PRs: #1719 (open), #1738 (closed); motivating issue #2096

#1719 diagnoses the same root cause (tier language silently depends on the agent knowing the current lineup) but is a broad multi-file PR. This PR is a minimal, independently mergeable two-hunk fix for the one sentence where the failure was observed; if maintainers prefer, it can fold into #1719.

#1738 was closed with the ruling that evals say SDD **task** dispatches should NOT default to the most capable frontier model — the controller should use tier judgment. This PR is consistent with that ruling: it does not extend "most capable" to any additional dispatch. The current dev section (post-rewrite) deliberately keeps "most capable available model" for exactly one dispatch, the final whole-branch review; this PR only makes that one phrase resolve the way the section already intends, and leaves every cost-scaling rule untouched.

## Environment tested

| Claude Code | 2.1.223 | Fable 5 (controller / test subjects) | claude-fable-5 |
| Claude Code | 2.1.223 | Sonnet (test subjects) | claude-sonnet-5 |

N/A — no new harness.

## Evaluation

- **Initial prompt:** the user pasted https://github.com/obra/superpowers/issues/2096 and asked for a PR addressing it. The issue itself was filed from the original failing session and contains the controller's verbatim retrospective.
- **Eval sessions after the change:** 8 single-shot subagent runs with the amended section (5 sonnet + 3 fable), plus 15 baseline runs before it (see below) — 23 runs total.
- **Outcome change:** the failing configuration went from **0/5 correct → 5/5 correct**; the previously-passing configuration stayed at 100%.

Methodology (per `writing-skills` micro-testing): each rep is a fresh single-shot subagent given the full Model Selection section verbatim plus session facts (session model `claude-fable-5`; Agent-tool model values `"sonnet", "opus", "haiku", "fable"`; earlier tasks ran on haiku/sonnet per cost-scaling; late in a long, token-heavy session), asked which model value it passes for the final whole-branch review. Every response was read manually.

| Arm | Subject | Section text | Result |
|-----|---------|-------------|--------|
| Baseline v1 (capability order spelled out in facts) | fable ×5 | current | 5/5 fable — with the ranking handed over, the wording resolves fine |
| Baseline v2 (no ordering given) | fable ×5 | current | 5/5 fable — fable knows its own tier from its system prompt |
| **Baseline v3 (no ordering, cost/late-session pressure)** | sonnet ×5 | current | **0/5 — all picked opus** |
| **Fixed v3 (identical scenario)** | sonnet ×5 | amended | **5/5 fable** |
| No-regression (v2 scenario) | fable ×3 | amended | 3/3 fable |

Baseline v3 rationalizations mirror the real session's, e.g. *"since the session model (fable) isn't in the Agent tool's override list, opus is the most capable option actually offered"* (fable was literally in the list) and *"the wording clearly distinguishes the session default (fable) from the top model, so pick opus, the top of the standard hierarchy."* Fixed-arm reps converged on one shape: rank all four values, session model included → strict maximum = fable, passed explicitly. The v1/v2 baselines are why the fix targets resolution rather than adding emphasis: the failure only appears when the controller lacks (or under-trusts) the session model's rank — which is every controller whose training predates the newest tier, and, under cost pressure, at least one real fable session.

## Rigor

Adversarial testing results are the table above: the adversarial arm (v3) combines cost-conservation momentum, late-session token pressure, and a subject that cannot verify the session model's rank — the same pressures as the real failure — and the amended wording flips it 0/5 → 5/5. The added Common Rationalizations row counters the exact excuse the baseline produced; no existing row was modified.

## Human review

Draft until @RyosukeNishi reviews the complete diff; the box will be checked and the PR marked ready at that point.

---

### DIFF
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -187,8 +187,13 @@ Use the least powerful model that can handle each role to conserve cost and incr
 **Integration and judgment tasks** (multi-file coordination, pattern matching, debugging): use a standard model.
 
 **Architecture and design tasks**: use the most capable available model.
-The final whole-branch review is one of these — dispatch it on the most
-capable available model, not the session default.
+The final whole-branch review is one of these. Resolve "most capable
+available" mechanically: rank every model you can pass when dispatching —
+including the session's own model and any tier above the usual top tier —
+and pick the strict maximum. Cost-scaling does not apply to this dispatch.
+Pass the model explicitly even when the choice equals the session's model:
+explicitness rules out inheritance by omission, not the session's model
+itself.
 
 **Review tasks**: choose the model with the same judgment, scaled to the
 diff's size, complexity, and risk. A small mechanical diff does not need the
@@ -496,6 +501,7 @@ Use superpowers:finishing-a-development-branch.
 | "Reviews slow the loop down" | The loop without reviews is just unverified churn. Reviews are the loop's brakes and steering. |
 | "Ledger bookkeeping is overhead" | The ledger is what survives compaction. Controllers without one have re-dispatched entire completed task sequences. |
 | "The implementer spawned its own reviewer — free extra assurance" | It's a duplicate seat reviewing the same diff; the task review is the gate. A worker-spawned reviewer is a defect to flag, not rigor. |
+| "The session model is expensive; the top standard tier is close enough for the final review" | The final review is the one dispatch where cost-scaling does not apply. Rank every model you can dispatch — the session's own model included — and pick the strict maximum. |
 
 ## Example Workflow
 


obra:
Triage update: this is the strongest outside skill-content PR in the current window — real motivating session (#2096, now labeled), prior art addressed properly, and an eval battery in the right shape. Two things before it can move:

1. **We verify eval claims independently.** Skill-content changes land on the strength of evals we've run ourselves, so this is queued for verification on our harness. If the baseline misparse and the 5/5 amended results reproduce, it lands with credit.
2. **The human-review attestation needs to be true.** The body says this is 'a draft until that review happens' and the box 'will be checked at that point' — but the PR isn't a draft and the box is checked now. Please either convert to draft, or confirm on-record that the full-diff review has actually happened.

No further changes needed from your side beyond point 2 — nice work on the report.

RyosukeNishi:
@obra Thank you for your comment.
> confirm on-record that the full-diff review has actually happened.

As the author, I confirm that the full-diff review is complete.

Issue #2081, opened 2026-08-04
hooks/session-start: six subprocess spawns cost ~2s on Windows (7.7x slower; byte-identical output without them)


midirectiekade14-debug:
`hooks/session-start` spawns six subprocesses to do work that needs none. On Linux/macOS that is cheap enough to be invisible. On Windows, where `fork()` is emulated, it dominates: the hook takes **~2.25s** instead of **~0.29s**, and the whole difference is process-spawn overhead.

Removing the spawns produces **byte-identical output** (same SHA-256, same length, same exit code).

## Environment

| | |
|---|---|
| OS | Windows 11 Pro 26200 |
| Shell | Git for Windows bash 5.2 (`C:\Program Files\Git\bin\bash.exe`) |
| Plugin | superpowers 6.2.0 (installed via claude-plugins-official) |
| Host | Claude Code |

Verified the same six spawns are still present on `main` (checked 2026-08-03): lines 7, 8, 11, and the three `| cat` at 40/43/46.

## Measurement

7 runs each, `CLAUDE_PLUGIN_ROOT` set as the host sets it, machine otherwise idle. `bash -c true` is included as the floor — that is what an empty script costs on this box.

| variant | min | median | max |
|---|---:|---:|---:|
| `bash -c true` (floor) | 291 ms | 506 ms | 838 ms |
| `hooks/session-start` as shipped | 2248 ms | **2254 ms** | 2315 ms |
| same script, spawns removed | 288 ms | **291 ms** | 303 ms |

The patched version lands on the floor: its own work is not measurable above bash startup itself. The shipped version adds ~1.95s of pure overhead.

### Where the time goes

Timed step by step inside the script:

| step | what it is | cost |
|---|---|---:|
| `SCRIPT_DIR=$(cd … && pwd)` + `PLUGIN_ROOT=$(cd … && pwd)` | 2 subshells | ~2700 ms |
| `$(cat …/SKILL.md)` | subshell + `cat` process, for 3 KB | ~2700 ms |
| the five `${s//old/new}` passes | pure bash, **no fork** | ~950 ms |
| `printf … \| cat` | pipe + a `cat` that does nothing | ~3100 ms |

[...]

Issue #2091, opened 2026-08-05
SessionStart hook fails on WSL : `CLAUDE_PLUGIN_ROOT` contains Windows-style backslashes, `run-hook.cmd` "not found"


emmaanuel:
- [x] I searched existing issues and this is not a duplicate (checked issues mentioning `CLAUDE_PLUGIN_ROOT`, `run-hook.cmd` and "SessionStart hook fails" — related reports are all **closed as fixed**, and none describe backslashes appearing in `CLAUDE_PLUGIN_ROOT` while running under WSL2/Linux, so this looks like a regression rather than a straight duplicate.)

## Environment (required)

| Field | Value |
|-------|-------|
| Superpowers version | 6.2.0 |
| Harness (Claude Code, Cursor, etc.) | GitHub Copilot (VS Code) — plugin hooks loaded from `~/.copilot/installed-plugins/superpowers-marketplace/superpowers` |
| Harness version | Visual Studio Code 1.131 |
| Your model + version | Sonnet 5 |
| All plugins installed | superpowers-marketplace/superpowers |
| OS + shell | WSL 2 under windows 11. Linux (remote/VS Code Server, `/home/manu/.vscode-server`), `/bin/sh` |

## Is this a Superpowers issue or a platform issue?

- [x] I confirmed this issue does not occur without Superpowers installed
      *(likely a harness bug in how GitHub Copilot Chat substitutes `CLAUDE_PLUGIN_ROOT` for hook commands — filing here first since it breaks Superpowers' SessionStart hook specifically, and `run-hook.cmd`'s own path-handling can't work around it, see analysis below)*

## What happened?

On session start, the `SessionStart` hook fails with `not found` because the `CLAUDE_PLUGIN_ROOT` environment variable substituted into the hook command uses **Windows-style backslash path separators**, even though the hook is executed on Linux via `/bin/sh`.

The configured command in `hooks/hooks.json` is:
```
"\"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start"
```

After substitution it became:
```
"\home\manu\.copilot\installed-plugins\superpowers-marketplace\superpowers/hooks/run-hook.cmd" session-start
```

Since `/bin/sh` treats backslashes as literal characters (not path separators) on Linux, this does not resolve to the actual file `/home/manu/.copilot/installed-plugins/superpowers-marketplace/superpowers/hooks/run-hook.cmd`, and the shell reports the whole string as a nonexistent command.

[...]

PR #2115, opened 2026-08-10
fix(hooks): eliminate subprocess spawns in session-start

SamedBilginAlternet:
> **This PR targets `dev`, not `main`.**

| Field | Value |
|-------|-------|
| Your model + version | Claude Sonnet 5 (`claude-sonnet-5`) |
| Harness + version | Claude Code 2.1.226 |
| All plugins installed | `supabase@claude-plugins-official` v0.1.13 (user-scope; unrelated to this change) |
| Human partner who reviewed this diff | Samed Bilgin (@SamedBilginAlternet) |

Issue #2081: `hooks/session-start` forks 6 subprocesses on every session
start for work that needs none. It's harmless on Linux/macOS, but on
Windows (where `fork()` is emulated) the reporter measured 2254ms as
shipped vs. 291ms with the spawns removed, roughly 2s of pure overhead on
every session start, every `/clear`, and every auto-compact. They
confirmed byte-identical output (same SHA-256, same length, same exit
code) across 7 runs each.

I'm not the one who hit this in a live session, and I want to be upfront
about that rather than imply otherwise: I found this by reading the issue
tracker, not from a session where the slowdown affected me directly. What
made me pick this one specifically is that the issue isn't a request for
someone else to investigate. The reporter already fully diagnosed it,
wrote a complete patch, and verified it byte-identical to the original
(SHA-256 `69cde947...`, which I independently reproduced myself before
touching any code, same hash). They explicitly said they didn't open a PR
themselves only because of one open question about a minor error-path
behavior change, and invited whoever picked it up to send one. This PR
implements their patch, with one adjustment (see below), plus my own
independent verification of every claim rather than taking the issue at
its word.

## What does this PR change?

Implements the three-part patch from the issue: (1) uses
`CLAUDE_PLUGIN_ROOT`/`CURSOR_PLUGIN_ROOT` directly when the host sets it,
falling back to `${0%/*}` parameter expansion instead of forking
`dirname`+`cd`+`pwd`; (2) reads `SKILL.md` via `$(<file)` instead of
forking `cat`; (3) drops the three `printf ... | cat` pipes, which don't
do anything `printf` alone doesn't already do, and inlines the JSON-escape
substitutions on the target variable instead of capturing them through a
function call, removing that fork too. `strace -f -e trace=clone`: 8
`clone()` calls per invocation down to 0.

Also adds two cases to `tests/hooks/test-session-start.sh` covering the
two disclosed behavior changes below (missing-`SKILL.md` fallback text,
and `CLAUDE_PLUGIN_ROOT` being honored over `$0`), so a future edit that
silently reverts either one fails CI instead of only being caught by
someone re-reading this PR body.

Yes. `hooks/session-start` is core plugin infrastructure that runs for
every user of every harness on every session start. It isn't
project-specific, domain-specific, or a third-party integration.

I want to be honest about the actual sequence here rather than presenting
a clean narrative. I initially wrote my own, more conservative version of
this fix: the `${0%/*}` path-resolution piece and dropping `| cat`, but I
left the `SKILL.md` read as `$(cat ...)` untouched, since removing that
fork changes the error-path semantics and I wasn't sure it was worth the
risk. I also tried converting the JSON-escape function into a nameref
out-parameter to remove its fork, but reverted that because ShellCheck
(SC2154) can't see the nameref assignment and flags the target variable
as unassigned. This repo's shell scripts have zero existing `shellcheck
disable` comments, and I didn't want to be the first PR to add one for a
small gain.

Only after drafting that did I read the rest of the issue body (past the
initial repro/measurement) and find that the reporter had already solved
both problems I was hedging on: `$(<file)` reads the file without forking
`cat` and is still a command substitution, so it still strips trailing
newlines the same way `$(cat ...)` does; and inlining the substitutions
directly onto the target variable removes the escape function's fork
without needing a nameref at all. Their version is strictly better than
what I'd written (it gets to zero forks instead of three, and it doesn't
need a lint suppression), so I adopted it instead of my own draft.

The one place I kept their patch exactly as proposed rather than picking
the alternative they offered: they flagged that `[ -r "$skill_file" ]`
changes the failure path (the original's `cat ... 2>&1 || echo ...` folds
`cat`'s own stderr text into the injected content on a partial failure;
the guard falls back to a clean fixed string instead) and asked
maintainers which behavior they wanted before opening a PR themselves. I
did not see a response to that question anywhere in the issue thread, so
I'm not resolving it unilaterally. I'm submitting their proposed behavior
and flagging it here for a maintainer to weigh in on, same as they did.

No. All three code changes (path resolution, the file read, removing
`| cat`) are the same fix for the same problem (unnecessary subprocess
spawns in this one script) and were designed together as one patch in the
source issue. The two added test cases cover behavior changes introduced
by that same fix, not a separate concern.

## Existing PRs
- Related PRs: none found. Searched open+closed PRs for "session-start",
  "2081", "subprocess spawn", and "fork Windows": no PR references issue
  #2081 or implements this patch. (#2095 and #2053 touch
  session-start/cursor hook resolution but for unrelated bugs: macOS
  heredoc deadlock and Cursor plugin-root resolution, respectively.)

## Environment tested

|---------|------------------|-------|-------------------|
| Claude Code (Linux) | 2.1.226 | Claude Sonnet 5 | claude-sonnet-5 |

I could not test the Windows wall-clock improvement firsthand: no Windows
environment available, and I'm relying on the issue's own measurement for
that number rather than reproducing it. What I verified myself, on Linux:

- `strace -f -e trace=clone`: 8 → 0 `clone()` calls per invocation
- SHA-256-identical output before/after across all 4 output-shape branches
  (Cursor, Claude Code, Copilot CLI, unknown-platform/SDK-standard); the
  Claude Code branch hash (`69cde947...`) matches the one the issue
  reporter posted independently
- The `$0`-has-no-slash edge case (no host env var set, script invoked as
  a bare name on PATH) still resolves correctly
- The missing-`SKILL.md` error path: confirmed the only difference from
  original behavior is the documented stderr-folding change above; built
  a throwaway copy of the hook with no `skills/` directory and diffed
  original vs. patched output directly
- One behavior difference beyond what the issue documented, found while
  testing: the original never actually uses `CLAUDE_PLUGIN_ROOT` to locate
  its own files. It derives the path from `$0` unconditionally and only
  reads `CLAUDE_PLUGIN_ROOT` later to pick the output JSON shape. So if a
  host ever set `CLAUDE_PLUGIN_ROOT` to something other than where the
  script actually lives, the original would silently ignore that and keep
  reading files relative to `$0`. This patch makes `CLAUDE_PLUGIN_ROOT`
  authoritative for file lookup when set, matching what the variable name
  implies it should do. I don't have evidence this divergence has caused
  a real bug, so I'm not claiming it as a second fix, just disclosing it
  as a behavior change I found, since the existing test suite always sets
  `CLAUDE_PLUGIN_ROOT` to match `$0`'s real location and wouldn't catch
  this either way
- `tests/hooks/test-session-start.sh`: the existing 6 assertions pass
  unmodified, plus 2 new ones I added for the two disclosed behavior
  changes above (8/8 pass). I confirmed both new assertions actually fail
  against the pre-patch script before restoring the fix, so they're real
  regression coverage, not tests that would pass either way
- `scripts/lint-shell.sh hooks/session-start tests/hooks/test-session-start.sh`:
  clean (shellcheck + shfmt), no suppressions added
- `bash -n hooks/session-start`: syntax OK

## Evaluation

- Initial prompt: none. This PR originates from reading issue #2081, not
  from a session where I personally hit the bug.
- Eval sessions run after the change: not applicable. This is a mechanical
  subprocess-count fix, not a change to skill-behavior content, so there's
  no agent behavior to eval. Correctness was verified by
  byte-identical-output comparison (see Environment tested) rather than
  session evals.

## Rigor

- [ ] If this is a skills change: N/A. This touches `hooks/session-start`,
      not skill content.
      (missing-file error path, no-slash `$0`, all 4 output-shape
      branches, mismatched `CLAUDE_PLUGIN_ROOT`)
      rationalizations, "human partner" language). This PR touches only
      shell mechanics in one hook script.

## Human review

### DIFF
--- a/hooks/session-start
+++ b/hooks/session-start
@@ -3,27 +3,50 @@
 
 set -euo pipefail
 
-# Determine plugin root directory
-SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
-PLUGIN_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+# Determine plugin root directory without forking dirname/cd/pwd subshells -
+# on Windows each subprocess spawn costs ~250-300ms (fork() is emulated).
+# hooks.json invokes the wrapper as "${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd",
+# so the variable is guaranteed set under Claude Code; Cursor sets
+# CURSOR_PLUGIN_ROOT. The ${0%/*} fallback keeps standalone execution
+# working without a subshell (it does not resolve ".." or symlinks the way
+# `cd && pwd` did, but nothing here ever displays SCRIPT_DIR/PLUGIN_ROOT -
+# it is only used to build a path handed to a file read, which resolves
+# ".." itself).
+if [ -n "${CLAUDE_PLUGIN_ROOT:-}" ]; then
+    PLUGIN_ROOT="${CLAUDE_PLUGIN_ROOT}"
+elif [ -n "${CURSOR_PLUGIN_ROOT:-}" ]; then
+    PLUGIN_ROOT="${CURSOR_PLUGIN_ROOT}"
+else
+    case "$0" in
+        */*) PLUGIN_ROOT="${0%/*}/.." ;;
+        *) PLUGIN_ROOT=".." ;;
+    esac
+fi
 
-# Read using-superpowers content
-using_superpowers_content=$(cat "${PLUGIN_ROOT}/skills/using-superpowers/SKILL.md" 2>&1 || echo "Error reading using-superpowers skill")
+# Read using-superpowers content via a bash builtin instead of forking cat.
+# Note: this changes the failure-path behavior slightly. The original's
+# `2>&1 || echo ...` folded cat's own stderr text into the injected content
+# on a partial failure; this guard instead falls back to a clean fixed
+# string. Flagging this explicitly since it's the one behavior difference
+# in this change - everything else is byte-identical to the original.
+skill_file="${PLUGIN_ROOT}/skills/using-superpowers/SKILL.md"
+if [ -r "$skill_file" ]; then
+    using_superpowers_content=$(<"$skill_file")
+else
+    using_superpowers_content="Error reading using-superpowers skill"
+fi
 
 # Escape string for JSON embedding using bash parameter substitution.
 # Each ${s//old/new} is a single C-level pass - orders of magnitude
-# faster than the character-by-character loop this replaces.
-escape_for_json() {
-    local s="$1"
-    s="${s//\\/\\\\}"
-    s="${s//\"/\\\"}"
-    s="${s//$'\n'/\\n}"
-    s="${s//$'\r'/\\r}"
-    s="${s//$'\t'/\\t}"
-    printf '%s' "$s"
-}
-
-using_superpowers_escaped=$(escape_for_json "$using_superpowers_content")
+# faster than the character-by-character loop this replaces. Done in
+# place on the target variable instead of through a captured function
+# call so this doesn't fork a subshell either.
+using_superpowers_escaped="$using_superpowers_content"
+using_superpowers_escaped="${using_superpowers_escaped//\\/\\\\}"
+using_superpowers_escaped="${using_superpowers_escaped//\"/\\\"}"
+using_superpowers_escaped="${using_superpowers_escaped//$'\n'/\\n}"
+using_superpowers_escaped="${using_superpowers_escaped//$'\r'/\\r}"
+using_superpowers_escaped="${using_superpowers_escaped//$'\t'/\\t}"
 session_context="<EXTREMELY_IMPORTANT>\nYou have superpowers.\n\n**Below is the full content of your 'superpowers:using-superpowers' skill - your introduction to using skills. For all other skills, use the 'Skill' tool:**\n\n${using_superpowers_escaped}\n</EXTREMELY_IMPORTANT>"
 
 # Output context injection as JSON.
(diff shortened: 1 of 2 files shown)


obra:
Full review done (security + history archaeology). The perf goal is sound, the submission is careful, and the disclosed error-path change is fine — but review surfaced two functional regressions that need rework before this can land:

1. **The `| cat` removal reverts a deliberate fix, and the PR's premise for removing them is wrong.** The body says the pipes 'don't do anything `printf` alone doesn't already do' — but they were added on purpose in d72560e ('Pipe SessionStart hook printf through cat to absorb EPIPE on Windows'): Git Bash printf reports EPIPE as 'Permission denied' when the harness closes the hook's stdout early, and under `set -euo pipefail` that escapes as a confusing per-session diagnostic. Your patch removes the absorption without replacing it. The fix that keeps your perf win: handle EPIPE with zero spawns — `printf ... || true` (or an explicit exit-code check) on the output writes — and add a regression test that simulates the closed-stdout condition so d72560e's scenario stays covered.

2. **Making `CLAUDE_PLUGIN_ROOT` authoritative for the SKILL.md read is a robustness regression.** Today the env var only selects output shape; the content read resolves from `$0` and survives env-var corruption. Issue #2091 shows harnesses delivering `CLAUDE_PLUGIN_ROOT` with backslash-mangled paths — under your change that degrades to the fallback string being injected every session: superpowers silently dead with no visible error. Keep `$0`-based resolution for the read, or at minimum fall back to it when the env-var path isn't readable, and pin that with a test using a deliberately garbage `CLAUDE_PLUGIN_ROOT`.

With those two corrections (both zero-subprocess, so the ~2s Windows win survives intact) and the two extra regression tests, this is in good shape. The care in the original submission — full disclosure, honest provenance, pinning your own behavior changes with tests — is exactly what we want; the gap was git history, not diligence.

— Claude Fable 5, Claude Code 2.1.228, reviewing on behalf of @obra

PR #2117, opened 2026-08-10
sdd: warn review-package on foreign-ticket commits in range

t0domanh:
## Summary

- `review-package` now derives an "own" ticket ID from `PLAN_FILE`'s basename and scans `BASE..HEAD` commit subjects for a *different*, explicit ticket ID reference. When found, it non-fatally warns on stderr and prepends a `## ⚠ Foreign-ticket commits in this range` section to the output diff file.
- On a shared, non-worktree-isolated checkout, another session's commits can otherwise land silently inside a review package's range — the exact scenario that motivated this: every review package generated during one real session had to be manually re-scoped because `git log BASE..HEAD` picked up four other in-flight tickets' commits.
- Plan files with no identifiable ticket ID in their name are unaffected (many plans have none — the check is skipped silently). Commits with no ticket reference at all in their subject are never flagged — only an explicit *differing* ticket ID triggers the warning.
- `SKILL.md` documents the warning at both call sites (per-task review, final whole-branch review) and what the controller should do when it fires: re-scope, or proceed and say so if the inclusion is intentional.
- No new CLI arguments — `review-package PLAN_FILE BASE HEAD [OUTFILE]` is unchanged.

This went through this repo's own `subagent-driven-development` process end-to-end: per-task implementation + review for both the script change and the doc change, then a whole-branch final review that caught two real bugs (a command-substitution newline strip that garbled multi-commit warnings, and a `head -n1` that could misflag a commit's own ticket) — both fixed and re-verified before this PR.

## Test plan

- [x] `tests/claude-code/test-sdd-workspace.sh` — 5 new/updated assertions covering: same-ticket range (no warning), single foreign-ticket commit (warning + file section), two foreign-ticket commits (each on its own line), plan with no identifiable ticket (check skipped), commit with no ticket reference at all (never flagged). All 5 pass.
- [x] Note: this same test file has 4 pre-existing, unrelated failing assertions on a fresh Windows Git Bash checkout of `main` with no changes applied (`prints <repo-root>/...`, `task-brief writes its brief...`, `review-package writes its diff...`, `linked worktree resolves...`) — a `/tmp` vs. physical-path resolution quirk in that environment, verified via `git stash` before any of this branch's changes were applied. Not caused by, or related to, this PR.

### DIFF
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -402,7 +402,8 @@ parked-with-ruling at the cap.
 
 ## Final Review
 
-The final whole-branch review gets a package too: run
+The final whole-branch review gets a package too (heed any foreign-ticket
+warning it prints): run
 `scripts/review-package PLAN_FILE MERGE_BASE HEAD` (MERGE_BASE = the commit the
 branch started from, e.g. `git merge-base main HEAD`) and include the
 printed path in the final review dispatch, so the final reviewer reads
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -270,6 +270,18 @@ needed.
   call. Use the BASE you recorded before dispatching the implementer —
   never `HEAD~1`, which silently truncates multi-commit tasks. Never
   dispatch a task reviewer without a diff file.
+- `review-package` warns — non-fatally, on stderr and in a header section
+  of the output file — when `BASE..HEAD` contains a commit whose message
+  references a different ticket ID than the plan file's own. On a shared,
+  non-worktree-isolated checkout, other sessions can land unrelated
+  commits in that window; a review package that silently includes them
+  wastes the reviewer's attention on work outside this task's scope. When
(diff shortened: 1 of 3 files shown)


obra:
Closing. Three reasons, in increasing order of importance:

1. This targets `main`; all PRs must target `dev`.
2. The PR template wasn't filled — no submitter identification, no environment table, no search for existing PRs. Those sections are mandatory.
3. The approach itself doesn't fit: ~60 lines of ticket-ID regex heuristics (`[A-Za-z]{3,}[#-]?[0-9]{2,}`, with self-admitted false positives on tokens like `sha256`) is fuzzy pattern-matching in a script that otherwise does mechanical verification. SDD's answer to commit-range contamination is worktree isolation plus mechanical range checks — see #2082, which addresses the same underlying problem (#2050) with merge-base ancestry guards instead of content heuristics. That's the direction we'd take.

If you're hitting foreign-commit contamination in real sessions, comments on #2050 with transcripts are welcome — that's where this problem is being tracked.

PR #2119, opened 2026-08-10
fix(requesting-code-review): anchor BASE_SHA to the merge base, not a moving ref

### LINKED ISSUE #2118: requesting-code-review: the "# or origin/main" BASE_SHA feeds a moving ref into a two-dot diff, so reviewers are shown deletions the branch never made

<sub>Searched open and closed for `merge-base`, `BASE_SHA`, `BASE_SHA diff`, `two-dot`, `code-reviewer diff origin/main`, `reviewer deletions`. Nearest hits (#1481, #1538, #436) are about review *scope* and uncommitted changes, not diff range semantics.</sub>

## Environment (required)

| Field | Value |
|-------|-------|
| Superpowers version | 6.2.0 (latest release; also present on `main` today) |
| Harness version | 2.1.226 |
| Your model + version | Claude Opus 5 (1M context), `claude-opus-5[1m]` |
| All plugins installed | superpowers@6.2.0, schliff@8.10.1, roundtable/agent-review-panel@3.8.3, memory-hygiene@3.4.0, context-police@2.4.0, session-handoff@1.18.0, agent-traffic-control@1.12.0, show-and-tell@2.2.0, promptback@2.0.0, publish-skill@2.6.0, skill-sync@1.3.0, + ~25 others |
| OS + shell | macOS 26.6 (Darwin 25.6.0), zsh |

**Superpowers.** It is two lines of skill text; no platform behaviour is involved.

## Summary

`skills/requesting-code-review/SKILL.md:28` offers two ways to set `BASE_SHA`:

```bash
BASE_SHA=$(git rev-parse HEAD~1)  # or origin/main
```

That value is consumed downstream by `code-reviewer.md:29-30` in a **two-dot** range:

```
git diff --stat [BASE_SHA]..[HEAD_SHA]
git diff [BASE_SHA]..[HEAD_SHA]
```

`HEAD~1` is always an ancestor of `HEAD`, so `A..B` and `A...B` are identical and the default is correct. **`origin/main` is not an ancestor once `main` moves.** Two-dot compares tip-to-tip, so everything `main` gained after the branch point renders as **deletions the branch appears to be making** — and the reviewer subagent, which cannot see how the diff was produced, reports them as findings.

## Reproduction

```bash
git init -q -b main . && git config user.email t@t.t && git config user.name T
echo 'shared helper' > helper.py && git add . && git commit -qm base

git checkout -qb feature
echo 'my new feature' > feature.py && git add . && git commit -qm "add feature.py"

# a teammate lands 200 lines on main while the branch is open
git checkout -q main
seq 1 200 | sed 's/^/def teammate_/' > teammate_module.py
git add . && git commit -qm "teammate lands teammate_module.py"
git checkout -q feature
```

Now follow the skill both ways:

| `BASE_SHA` | `git diff --stat $BASE_SHA..HEAD` |
|---|---|
| `$(git rev-parse HEAD~1)` — the documented default | `1 file changed, 1 insertion(+)` ✅ |
| `origin/main` — the `# or origin/main` alternative | `2 files changed, 1 insertion(+), 200 deletions(-)` ❌ |

```
$ git diff --diff-filter=D --name-only origin/main..HEAD
teammate_module.py
```

The feature branch never touched `teammate_module.py`. What the change actually proposes:

```
$ git diff --stat origin/main...HEAD     # three-dot; matches GitHub's Files-changed tab
 1 file changed, 1 insertion(+)
```

## Impact

The reviewer is a subagent reading a diff it did not produce, so it has no way to tell an invented deletion from a real one. In the worst case it reports a phantom "this PR deletes a teammate's module" finding; in the milder case it burns review budget on unrelated files.

The failure is also **silent and intermittent** — it only appears once `main` moves, so it will not reproduce for whoever investigates it later.

## Suggested fix

One line, and it keeps the existing downstream two-dot correct by construction:

```diff
-BASE_SHA=$(git rev-parse HEAD~1)  # or origin/main
+BASE_SHA=$(git merge-base origin/main HEAD)  # or HEAD~1 for a single commit
```

`git merge-base` returns an ancestor of `HEAD` by definition, so `[BASE_SHA]..[HEAD_SHA]` in `code-reviewer.md` becomes equivalent to the three-dot diff and needs no change.

**This is already the convention elsewhere in Superpowers.** `skills/subagent-driven-development/scripts/review-package` computes a merge base before its `${base}..${head}` diff, which is why the final-review path does not have this bug and the per-task path does.

The same `[BASE_SHA]..[HEAD_SHA]` pattern also appears in `skills/subagent-driven-development/task-reviewer-prompt.md:44` and `re-review-prompt.md:40-41`; those are safe or unsafe depending on what the dispatcher substitutes, so fixing the source of `BASE_SHA` covers them too.

## Why I noticed

We traced a wrong-diff false alarm in our own repo to the same two-dot/three-dot confusion. It cost two investigations in an afternoon, and the mis-measured figure reached five documents before anyone re-ran it. Superpowers is not where our instance originated, but `requesting-code-review` is the most-reached place the same shape is written down, so it seemed worth reporting upstream.

Happy to send the one-line PR if useful.

wan-huiyan:
Fixes #2118.

## The bug

`skills/requesting-code-review/SKILL.md:28` offers two values for `BASE_SHA`:

```bash
BASE_SHA=$(git rev-parse HEAD~1)  # or origin/main
```

That value is consumed by `code-reviewer.md:29-30` in a **two-dot** range, which requires `BASE_SHA` to be an **ancestor** of `HEAD_SHA`:

```
git diff --stat [BASE_SHA]..[HEAD_SHA]
git diff [BASE_SHA]..[HEAD_SHA]
```

`HEAD~1` always is an ancestor, so the documented default is correct. **`origin/main` stops being one the moment `main` advances** — and two-dot compares tip-to-tip, so every file `main` gained after the branch point is rendered as a deletion *the branch appears to be making*. The reviewer subagent cannot see how its diff was produced, so it has no way to tell an invented deletion from a real one.

## Reproduction

Two-commit feature branch; a teammate lands 200 lines on `main` meanwhile:

| `BASE_SHA` | `git diff --stat $BASE_SHA..HEAD` |
|---|---|
| `$(git rev-parse HEAD~1)` — documented default | `1 file changed, 1 insertion(+)` ✅ |
| `origin/main` — the `# or origin/main` alternative | `2 files changed, 1 insertion(+), 200 deletions(-)` ❌ |
| three-dot `origin/main...HEAD` (= GitHub Files-changed) | `1 file changed, 1 insertion(+)` |

```
$ git diff --diff-filter=D --name-only origin/main..HEAD
teammate_module.py
```

The branch never touched that file. Full repro script in #2118.

## The fix

```diff
-BASE_SHA=$(git rev-parse HEAD~1)  # or origin/main
+BASE_SHA=$(git merge-base origin/main HEAD)  # or HEAD~1 for a single commit
```

`git merge-base` returns an ancestor by definition, so `[BASE_SHA]..[HEAD_SHA]` downstream becomes equivalent to the three-dot diff. **`code-reviewer.md` needs no change** — that is deliberate, to keep the diff minimal.

Plus a short paragraph explaining the ancestor requirement, so the next person editing this does not reintroduce a moving ref.

## Why `merge-base` rather than switching the reviewer to three dots

Superpowers already does it this way. `skills/subagent-driven-development/scripts/review-package` computes a merge base before its `${base}..${head}` diff — which is exactly why the final-review path does not have this bug while the per-task path does. This change makes `requesting-code-review` consistent with that, rather than introducing a second convention.

## Scope

One file, 7 insertions, 1 deletion. No version file in this skill's frontmatter, no tests reference `BASE_SHA`, and I left the worked example at line ~61 alone — `git log | grep "Task 1"` selects a commit on the branch's own history, which is an ancestor and therefore correct.

The same `[BASE_SHA]..[HEAD_SHA]` pattern in `subagent-driven-development/task-reviewer-prompt.md:44` and `re-review-prompt.md:40-41` is safe or unsafe depending on what the dispatcher substitutes; fixing the source of `BASE_SHA` covers the documented path without touching those.

## Why I noticed

We traced a wrong-diff false alarm in our own repo to the same two-dot/three-dot confusion — it cost two investigations in an afternoon, and the mis-measured figure reached five documents before anyone re-ran it. Superpowers is not where our instance originated, but `requesting-code-review` is the most-reached place the same shape is written down.

The failure is silent and intermittent — it only appears once `main` moves, so it will not reproduce for whoever investigates it later. That is what makes it worth fixing rather than documenting.

### DIFF
--- a/skills/requesting-code-review/SKILL.md
+++ b/skills/requesting-code-review/SKILL.md
@@ -25,9 +25,15 @@ Dispatch a code reviewer subagent to catch issues before they cascade. The revie
 
 **1. Get git SHAs:**
 ```bash
-BASE_SHA=$(git rev-parse HEAD~1)  # or origin/main
+BASE_SHA=$(git merge-base origin/main HEAD)  # or HEAD~1 for a single commit
 HEAD_SHA=$(git rev-parse HEAD)
 ```
+
+`BASE_SHA` must be an **ancestor** of `HEAD_SHA`, because the reviewer diffs `BASE_SHA..HEAD_SHA`
+(two dots). A moving ref like `origin/main` is not an ancestor once `main` advances, and two-dot
+compares tip-to-tip — so everything `main` gained after you branched is shown to the reviewer as
+**deletions your branch appears to be making**. `git merge-base` is an ancestor by definition, so the
+diff matches what GitHub's Files-changed tab shows.
 
 **2. Dispatch code reviewer subagent:**
 


shineday999:
The fix is correct and the explanation is clear. Verified locally —  produces the right diff while  shows phantom deletions once main advances. LGTM.

shineday999:
LGTM — verified locally. The merge-base fix produces the correct diff (only the feature file), while the origin/main alternative shows phantom deletions once main advances. The added explanation about the ancestor requirement is perfect and prevents future regressions.

obra:
Thanks for this report-and-fix pair — the underlying bug is real. We independently reproduced it today (see #2118): with the `# or origin/main` alternative, a two-dot diff against an advanced main shows phantom deletions the branch never made, and `git merge-base` is the right anchor.

Closing the PR itself on process grounds: it targets `main` (PRs go to `dev`), and the PR template — submitter identification, environment table, existing-PR search — wasn't filled; those sections are mandatory. The fix is tracked via #2118 and will land from there.

Issue #1961, opened 2026-07-10
Clarify how the SDD task-reviewer's "do not re-run the suite" reconciles with verification-before-completion


eranroseman:
Two skills give what reads as conflicting guidance to an agent following both.

**`skills/subagent-driven-development/task-reviewer-prompt.md`** (Tests section):
> The implementer already ran the tests and reported results with TDD evidence for exactly this code. **Do not re-run the suite to confirm their report.** Run a test only when reading the code raises a specific doubt that no existing run answers…

**`skills/verification-before-completion/SKILL.md`**:
> NO COMPLETION CLAIMS WITHOUT FRESH VERIFICATION EVIDENCE. If you haven't run the verification command in this message, you cannot claim it passes.

When the task reviewer approves a task, that approval is effectively a completion claim — yet it is told not to run verification itself, relying on the implementer's earlier run.

**Request:** document the intended reconciliation. Presumably the implementer's TDD run (for exactly this code) *is* the "fresh evidence," and verification-before-completion is satisfied by that run rather than requiring the reviewer to duplicate it — in which case a one-line cross-reference in each skill would remove the apparent conflict. If the reviewer is instead expected to run a final independent check before approving, the task-reviewer prompt should say so explicitly.

obra:
Hi,

Did you run into a problem with this in the field or was this based on static inspection?

arittr:
Closing for now since the question above about whether this came up in a real session has gone unanswered for several weeks. If you hit this in the field, please reopen with the session details and we can pick it back up.

Issue #2113, opened 2026-08-09
Five findings from a heavy multi-phase run of subagent-driven-development and writing-plans


seff34:
Five findings from running `subagent-driven-development` and `writing-plans` across four phases of one multi-phase iOS project — roughly 60 commits, ~30 dispatched subagents. Each one cost a real fix round or a wrong result that survived review, so each is followed by a concrete suggested change.

Filed as one issue because they came from one run and several touch the same sections; happy to split into separate issues if that is easier to triage.

---

## 1. Implementers stall on long verification runs, and stale logs get reported as fresh results

*(`subagent-driven-development` — The Task Loop)*

Almost every task in this project ended with a verification suite that takes about 22 minutes.

Roughly **fifteen times**, an implementer started that suite and then stopped — typically returning something like *"I'll wait for the notification that the test run has finished"* — and handed control back. The run either continued unwatched or died with the agent. In practice the controller ended up personally running nearly every verification in the project.

That alone is a wall-clock problem. The second-order effect is worse: an agent resumed later would look for a result, find a log in the scratchpad, and report it as its own. **Twice** an agent reported a green suite that belonged to an earlier task's code. Both times the controller caught it only by comparing the log's filename and mtime against the commit under review.

The skill already anticipates that agents get lost — the ledger exists precisely because "conversation memory does not survive compaction". But the dispatch guidance says nothing about *how* an implementer should wait for a long external job, and the controller-side guidance treats a completed implementer report as the normal case. In a project where every task has a 20-minute gate, the normal case is that the agent stops mid-run.

**Suggested change** — two additions to **The Task Loop**:

1. *In the implementer dispatch guidance:* long-running verification must be detached and polled, never awaited inline — e.g. `nohup <cmd> > <task-scoped-name>.log 2>&1 &`, then poll that log for its terminal marker. Require the log name to be scoped to the task (`t7-iphone.log`, not `uitest.log`), so a result can be tied to the code it verifies.
2. *In "Handle the report":* before accepting a suite result, check that the log identifies the code under test — by task-scoped name, or by comparing mtime against the commit. Add to the rationalisations table something like:

   | "The agent reported the suite green" | A log with a generic name proves nothing about *which* code was green. Two agents in one project reported an earlier task's result as their own. |

**Principle:** evidence must be identifiable to the thing it claims to verify. A verification instruction that does not bind the result to the artefact under test invites a stale result being reported as a fresh one — and the reporting agent has no way to notice.

---

[...]

PR #2127, opened 2026-08-12
fix(skills): reconcile task review verification guidance

SomSamantray:
## Issue

Issue #1961 has conflicting guidance: task reviewers are told not to rerun the suite, while verification-before-completion requires fresh verification evidence.

## Fix

Clarified that the implementer’s fresh, task-specific test run is the evidence the reviewer evaluates, while preserving focused tests for specific doubts and final completion verification.

### DIFF
--- a/skills/subagent-driven-development/task-reviewer-prompt.md
+++ b/skills/subagent-driven-development/task-reviewer-prompt.md
@@ -73,13 +73,16 @@ Subagent (general-purpose):
     ## Tests
 
     The implementer already ran the tests and reported results with TDD
-    evidence for exactly this code. Do not re-run the suite to confirm their
-    report. Run a test only when reading the code raises a specific doubt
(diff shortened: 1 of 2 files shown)


obra:
Closing. You've identified a real tension — the task-reviewer prompt's 'do not re-run the suite' instruction versus verification-before-completion's demand for re-derived evidence — and that's exactly what open issue #1961 tracks, alongside related discussion in #2113. The reconciliation will land from there.

Why not this PR: the change edits behavior-shaping content in two skills, and per CLAUDE.md those changes require eval evidence (before/after sessions showing the new wording actually improves reviewer behavior) — plus the PR template with full submitter identification, neither of which is present here. Getting that wording right is subtle: it has to allow re-derivation of gate results without reintroducing the review/test cost problems that motivated the original instruction.

If you've hit this conflict in a real session, a transcript on #1961 would be genuinely useful input.

PR #2108, opened 2026-08-07
docs: add Qwen Code install instructions to README

arittr:
> Targets `dev`.

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Claude Fable 5 (`claude-fable-5`) |
| Harness + version | Claude Code 2.1.224 |
| All plugins installed | superpowers (this repo, dev), superpowers-chrome 1.6.1, cloud-build 0.9.0, decision-log 0.1.0, elements-of-style 1.0.0, episodic-memory 1.0.15, iterative-development 0.1.0, linear (official), primeradiant-ops 1.9.2, rust-analyzer-lsp 1.0.0 |
| Human partner who reviewed this diff | Drew Ritter (@drewritter), maintainer — reviewed the complete diff before submission |

## What problem are you trying to solve?

Qwen Code now installs Claude Code marketplace plugins natively
(`qwen extensions install <owner>/<repo>`, see [their docs](https://qwenlm.github.io/qwen-code-docs/en/users/extension/introduction/#from-claude-code-marketplace)),
but the README doesn't mention it, so Qwen Code users have no way to know
Superpowers works there. Verified on the maintainer's machine: the install
command prompts to select `superpowers` from the marketplace, registers it
as extension `superpowers` (6.2.0), and loads the plugin's `GEMINI.md`
bootstrap context plus all 14 skills natively — the same mechanism Gemini
CLI uses. No integration code is needed; only install docs.

## What does this PR change?

Adds a Qwen Code entry to the README: TOC, Quickstart harness list, and an
install section (install + update commands) after Pi. A second commit fixes
adjacent README housekeeping noticed while making this change: Hermes Agent
was missing from the TOC and its section sat after Pi, out of the
alphabetical order every other harness follows — the section text moved
verbatim. A third commit drops the Hermes post-compaction caveat from that
block at maintainer direction. Docs only — no code, no skills touched.

## Is this change appropriate for the core library?

Yes — install documentation for a supported harness, same as the existing
Devin CLI / Grok Build CLI / Gemini CLI entries. Useful to any Qwen Code
user regardless of project.

## What alternatives did you consider?

- Shipping a dedicated Qwen integration (installer script, `.qwen-plugin`
  dir) as prior PRs attempted — rejected: Qwen's native Claude-marketplace
  support plus the existing Gemini extension format make it unnecessary.
  Zero-dependency docs-only is the smallest correct change.
- Documenting the `marketplace:plugin` install form from Qwen's docs —
[...]

PR #2132, opened 2026-08-13
docs: add Qwen Code install instructions to README

obra:
## What this does

Adds the Qwen Code install/update instructions to the README (section + TOC entry). This is a rebase-and-rework of #2108 (@arittr) onto the post-v6.3.0 README, which had made that PR CONFLICTING.

## Differences from #2108, and why

- **Kept:** the Qwen Code section verbatim — commands and the acceptance evidence are Drew's (real interactive Qwen Code session in #2108's transcript showing `brainstorming` auto-triggering).
- **Dropped — obsolete:** the Hermes TOC entry (the v6.3.0 README rework already added it) and the Quickstart harness-list edit (that line no longer exists).
- **Dropped — incorrect for current dev:** the deletion of the Hermes post-compaction caveat. Verified against `.hermes-plugin/__init__.py`: the bootstrap is injected only on `is_first_turn` via `pre_llm_call`, with no post-compaction re-injection, so the caveat ('a very long session that compacts over its first turn loses the bootstrap') is still accurate. @arittr — if you removed it because Hermes shipped a post-compaction hook we're not using yet, say the word and we'll wire it up instead.

Drew is credited via Co-authored-by on the commit.

## Who is submitting

Written by Claude Fable 5 on Claude Code 2.1.228, triage session run by @obra, who directed this PR and reviews the diff. Original change authored and Qwen-tested by @arittr in #2108.

Supersedes and closes #2108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Superpowers is a complete software development methodology for your coding agent
   - [Kimi Code](#kimi-code)
   - [OpenCode](#opencode)
   - [Pi](#pi)
+  - [Qwen Code](#qwen-code)
   - [Hermes Agent](#hermes-agent)
 - [The Basic Workflow](#the-basic-workflow)
 - [Community](#community)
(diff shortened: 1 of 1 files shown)


arittr (approved):
LGTM

PR #2069, opened 2026-08-01
fix(codex): preserve SDD helper executable modes in packaging

arimu1:
> **This PR MUST target the `dev` branch, not `main`.** `main` is the
> released branch; active work lands on `dev` first. PRs opened against
> `main` will be asked to retarget `dev` before review.

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Grok 4.5 |
| Harness + version | Grok Build (xAI) |
| All plugins installed | none (no Cursor plugins; implementator run under Grok Build only) |
| Human partner who reviewed this diff | arimu1 |

## What problem are you trying to solve?

Real report in #2040 from @michaelholcomb-creator: Superpowers **6.2.0** installed via the **Codex marketplace cache** arrived with these SDD helpers at mode `0644` (non-executable):

- `skills/subagent-driven-development/scripts/sdd-workspace`
- `skills/subagent-driven-development/scripts/task-brief`
- `skills/subagent-driven-development/scripts/review-package`

Direct invocation fails with `Permission denied`. Running only `task-brief` / `review-package` through Bash does **not** fully recover the default-output path, because both scripts previously **directly executed** their sibling `sdd-workspace` (which still needs +x).

Upstream git already records all three as `100755`. The portal packager (`scripts/package-codex-plugin.sh`) already aims to preserve modes, but:

1. Packaging tests only checked **one** helper after Info-ZIP extract, and did not assert the **zip external attributes** that installers read.
2. Zip extractors that ignore Unix external attributes (e.g. Python `zipfile.extractall`) leave helpers at `0644` even when the archive stores `0755` — verified locally: same archive → CLI unzip = `755`, Python extract = `644`.

## What does this PR change?

1. **Packaging:** After `git archive | tar -xpf`, re-apply `chmod a+x` for every staged path that is `100755` in the packaged ref, and fail closed if any remains non-executable. Accept linked worktrees (`.git` file) as valid checkouts.
2. **SDD helpers:** `task-brief` and `review-package` invoke `sdd-workspace` via `"${BASH:-bash}"` so default OUTFILE resolution works when helpers are mode-stripped but invoked with Bash.
3. **Tests:** Assert zip extract +x **and** stored zip external attrs `0o755` for all three SDD helpers; assert tar.gz modes for all three; add mode-stripped install-path coverage in `test-sdd-workspace.sh`.

## Is this change appropriate for the core library?

[...]

PR #2134, opened 2026-08-13
fix(sdd): invoke sdd-workspace via bash so helpers survive stripped exec bits

### LINKED ISSUE #2040: Codex marketplace package drops executable bits on SDD helper scripts
## Summary

Superpowers 6.2.0 installed through the Codex marketplace cache arrived with these helper files non-executable (observed mode `0644`):

- `skills/subagent-driven-development/scripts/sdd-workspace`
- `skills/subagent-driven-development/scripts/task-brief`
- `skills/subagent-driven-development/scripts/review-package`

Direct invocation fails with `Permission denied`. Invoking only `task-brief` or `review-package` through Bash does not fully recover the default-output path because both scripts directly execute their sibling `sdd-workspace`.

## Expected behavior

The upstream Git tree records all three files as `100755`. Please preserve those executable modes in the Codex marketplace packaging/install path.

## Downstream mitigation

Our framework now treats this only as a packaging workaround: it validates the allowlisted helper set, copies helper content to private staging, invokes it explicitly through `/bin/bash`, and never changes the generated plugin cache. This keeps the helpers usable but is not a substitute for preserving upstream modes.

No credentials or repository source are involved in this reproduction.

obra:
## Problem

Codex marketplace users get `Permission denied` when SDD helpers run (#2040): `task-brief` and `review-package` direct-exec their sibling `sdd-workspace`, which requires the executable bit — and some consumer-side extractors (Python `zipfile`) discard Unix mode attributes when unpacking. Our own packaging is not the culprit: `git archive | tar -xpf` preserves 0755 and the existing packaging test asserts it, so a packaging-side chmod (the approach in closed PR #2069) would guard a link that isn't broken.

## Fix

Two call sites: invoke the sibling via `"${BASH:-bash}"` instead of direct exec, making the exec bit irrelevant no matter what mangled the modes.

## TDD evidence

New regression case in `tests/claude-code/test-sdd-workspace.sh`: copy the three helpers to a temp dir, `chmod -x`, run `task-brief` via bash. RED on dev with exactly the reported failure (`rc=126 ... sdd-workspace: Permission denied`), GREEN with the fix; full suite passes.

Reported by @michaelholcomb-creator in #2040; diagnosis refined during triage (closed PR #2069 by @arimu1 identified the invocation change as part of a larger diff). Fixes #2040.

## Who is submitting

Claude Fable 5 on Claude Code 2.1.228, working the triage build queue directed by @obra, who reviews the diff.

### DIFF
--- a/skills/subagent-driven-development/scripts/review-package
+++ b/skills/subagent-driven-development/scripts/review-package
@@ -25,7 +25,9 @@ git rev-parse --verify --quiet "$head" >/dev/null || { echo "bad HEAD: $head" >&
 if [ $# -eq 4 ]; then
   out=$4
 else
-  dir=$("$(cd "$(dirname "$0")" && pwd)/sdd-workspace" "$plan")
+  # Invoke via bash rather than direct exec: some extractors (Python zipfile)
+  # strip Unix exec bits when unpacking marketplace packages (#2040).
+  dir=$("${BASH:-bash}" "$(cd "$(dirname "$0")" && pwd)/sdd-workspace" "$plan")
   out="$dir/review-$(git rev-parse --short "$base")..$(git rev-parse --short "$head").diff"
 fi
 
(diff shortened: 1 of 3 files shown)


arittr (approved):
LGTM

PR #2138, opened 2026-08-13
fix(sdd): ownership markers stop same-basename plans sharing a workspace

### LINKED ISSUE #2045: subagent-driven-development: plan-scoped workspace still collides when two plans share a basename (follow-up to #2012)
Following up on #2012 as asked: *"if you can still reproduce the collision on a release that includes #1943, please reopen with the session details."*

I can, on released **v6.2.0** (`3dcbd5c`). #1943 scopes the workspace by plan **basename**, so it fixes collisions between `plan-a.md` and `plan-b.md` but not between `docs/alpha/plan.md` and `docs/beta/plan.md`. Any repo where plans are foldered — one directory per feature, per sprint, per subproject — still gets the original silent overwrite. `plan.md`, `implementation-plan.md`, and `README.md` are exactly the basenames that repeat.

### Cause

`scripts/sdd-workspace:31`

```sh
slug=$(basename "$plan" .md)
```

`basename` discards the directory, which is the only part that distinguishes the two plans.

### Reproduction

```sh
SP=~/.claude/plugins/marketplaces/superpowers/skills/subagent-driven-development/scripts
T=$(mktemp -d); git init -q -b main "$T/repo"; R=$(cd "$T/repo" && git rev-parse --show-toplevel)
mkdir -p "$R/docs/alpha" "$R/docs/beta"
printf '# Alpha\n\n## Task 1: Alpha thing\n\nAlpha requirements: MAGIC_CONST=7.\n' > "$R/docs/alpha/plan.md"
printf '# Beta\n\n## Task 1: Beta thing\n\nBeta requirements.\n'                    > "$R/docs/beta/plan.md"
cd "$R"
"$SP/task-brief" docs/alpha/plan.md 1
"$SP/task-brief" docs/beta/plan.md  1
cat .superpowers/sdd/plan/task-1-brief.md
```

```
## Task 1: Beta thing

Beta requirements.
```

Both plans resolve to `<repo>/.superpowers/sdd/plan/`. Alpha's brief is gone, no warning, and the directory is gitignored so there is no history to recover from.

### Three consequences, in severity order

**1. Briefs and reports are overwritten with no identity check.** The ledger has a self-identifying first line, so `progress.md` collisions are at least *detectable*. `task-N-brief.md` and `task-N-report.md` have no such marker, and SKILL.md:207-209 makes the brief "the single source of requirements" — exact values, magic strings, and signatures appear only there. A subagent dispatched against a clobbered brief implements the wrong plan's task and has no way to notice.

**2. A `task-brief` invocation that *fails* still destroys the colliding file.** `scripts/task-brief:34` redirects `awk` output with `>`, which truncates before the not-found check at line 36:

```sh
awk -v n="$n" '...' "$plan" > "$out"

if [ ! -s "$out" ]; then
  echo "task ${n} not found in ${plan} ..." >&2
  exit 3
fi
```

So a wrong task number, a `## Step 1:` heading style, or a typo'd plan path leaves plan A's brief as a zero-byte file while telling the operator nothing was written:

```
alpha brief before:  59 bytes
task 1 not found in docs/beta/plan.md (no heading matching 'Task 1')
beta task-brief exit: 3
alpha brief after:    0 bytes
```

**3. The documented remedies both assume the directories are distinct.** SKILL.md:132-133 says a ledger naming a different plan "is another plan's progress: leave it in place and start your own, fresh" — but `<workspace>/progress.md` is a single fixed path inside a shared directory, so there is nowhere fresh to start. And SKILL.md:419-420, `rm -rf <workspace>` on clean final review, with "sibling directories belong to other plans; leave them alone": under a collision the other plan's artifacts are not in a sibling directory, they are in this one, and they get deleted mid-run.

`tests/claude-code/test-sdd-workspace.sh` only exercises `plan-a.md` and `plan-b.md` in the repo root, so the distinct-basename case is covered and the colliding one is not.

### Suggested fix

Derive the slug from the plan's repo-relative path rather than its basename, so the directory is unique for exactly the reason the plan file is:

```sh
root=$(git rev-parse --show-toplevel)
abs="$(cd "$(dirname "$plan")" && pwd -P)/$(basename "$plan")"   # -P matches --show-toplevel's symlink resolution
rel=${abs#"$root"/}
slug=$(printf '%s' "${rel%.md}" | tr '/' '-')
```

`docs/alpha/plan.md` → `docs-alpha-plan`, `docs/beta/plan.md` → `docs-beta-plan`, and `flat.md` → `flat`, so today's flat layout keeps the directory name it already has. Deliberately not `git ls-files`, which returns empty for a plan that hasn't been committed yet — the normal case when a plan is written and immediately executed.

Two edge cases need a decision rather than a snippet: a plan outside the repo root (the expression above degrades to a long absolute-ish slug — unique, but ugly), and `-`-vs-`/` ambiguity between `a/b-c.md` and `a-b/c.md`. A short digest suffix, or percent-encoding `/`, settles both.

Independently worth doing, since it converts silent loss into a visible error even if the slug stays as it is:

- `task-brief`: write to a temp file and `mv` into place only after the not-found check, and refuse to overwrite an existing file without `--force`.
- `sdd-workspace`: if `<dir>/progress.md` exists and its first line names a different plan, exit non-zero instead of returning the directory. That makes the collision loud at the one moment the operator can still act on it.

### Environment

superpowers v6.2.0 (`3dcbd5c`), Claude Code, macOS 26.5.1 (Darwin 25.5.0), bash 3.2.57.

obra:
## Problem (#2045)

`sdd-workspace` slugs by `basename "$plan" .md`, so `docs/alpha/plan.md` and `docs/beta/plan.md` resolve to one workspace and `task-brief` silently overwrites the other plan's brief — the single source of task requirements, gitignored, unrecoverable. Reproduced live during triage and again in this PR's RED run (alpha's brief ends up containing beta's text).

## Design (settled on #2045 after closing PR #2120)

- **Slugs stay basename-derived** — no ugly path slugs, no rename churn.
- **Out-of-repo plans keep working** (absolute-path marker).
- **No migration break**: a marker-less workspace (pre-marker or legacy) is adopted in place and marked — in-flight plans keep resolving. The documented tradeoff: the first collision on a legacy marker-less workspace adopts rather than detects; marker-less workspaces age out as plans finish.
- **Ownership marker**: each workspace records its owning plan's normalized path (`plan-path`; repo-relative in-repo, absolute outside). Marker names a different plan → disambiguate to `<slug>-<parent-dir>`, then a counter.
- Relative/absolute/`../` spellings of one plan normalize (physical path) to one workspace. All `cd`s are `CDPATH= cd --` guarded (carried from #2120's review findings).

`task-brief` and `review-package` need no changes — they delegate to `sdd-workspace` and hold no slug logic (verified).

## TDD evidence

8 new assertions written first, RED against the unmodified script — including the live #2045 reproduction (both plans → `.superpowers/sdd/plan`, brief clobbered) — then GREEN across the full suite with all 13 pre-existing assertions unmodified. One SKILL.md line updated because disambiguation makes the exact-path parenthetical false; ledger-check prose untouched.

Note: this appends to `test-sdd-workspace.sh` alongside #2134 and #2136 — whichever lands later gets a trivial append-order rebase.

Reported by @CRGDan in #2045; reproduction and test patterns credit @crisnahine's closed #2120. Fixes #2045.

## Who is submitting

Claude Fable 5 on Claude Code 2.1.228 (implementation subagent + controller review of the full diff), working the triage build queue directed by @obra, who reviews the diff.

### DIFF
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -135,7 +135,7 @@ a ledger file, not only in todos.
 
 - Each plan owns a workspace: at skill start, run this skill's
   `scripts/sdd-workspace PLAN_FILE` — it prints the plan's git-ignored
-  directory (`<repo-root>/.superpowers/sdd/<plan-basename>/`), home to
+  directory (under `<repo-root>/.superpowers/sdd/`), home to
   every artifact for THIS plan: ledger, briefs, reports, review packages.
   Another plan's directory is never yours to read or write.
 - Check for this plan's ledger at `<workspace>/progress.md`. If its first
--- a/skills/subagent-driven-development/scripts/sdd-workspace
+++ b/skills/subagent-driven-development/scripts/sdd-workspace
@@ -8,6 +8,16 @@
 # artifacts. A stale ledger misread as current progress makes controllers
 # skip whole task sequences — plan-scoping removes that failure structurally.
 #
+# Basename slugs collide when two plans share a filename (docs/alpha/plan.md
+# vs docs/beta/plan.md), so each workspace records its owning plan's path in
+# a plan-path marker (repo-relative in-repo, absolute outside). A workspace
+# owned by a different plan is skipped and the slug disambiguated with the
+# plan's parent-directory name, then a counter. A workspace with no marker
+# predates the marker scheme and is adopted for the current plan so in-flight
+# workspaces keep resolving — which means the first collision on such a
+# legacy workspace adopts instead of detecting; acceptable, marker-less
+# workspaces age out as plans finish.
+#
 # The workspace lives in the working tree (not under .git/) because Claude Code
 # treats .git/ as a protected path and denies agent writes there — which blocks
 # an implementer subagent from writing its report file. A self-ignoring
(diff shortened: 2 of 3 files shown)


arittr (approved):
LGTM

PR #2156, opened 2026-08-15
skill: script-first-tool-batching

adamteale:
## Thinking Path

> Analysed DeepSeek Harness (dsh, released 2026-08-13) and its Code Mode design — the observation that "LLMs are better at writing code than at emitting tool calls." Benchmarked the token cost of sequential tool calls vs script batching. Baseline-tested the skill with 8 live sessions (glm-5.3 + deepseek-v4-flash) per the writing-skills TDD methodology.

## Implementation

New skill: `skills/script-first-tool-batching/SKILL.md`

Teaches models to go straight to scripting for batch operations rather than exploring first (ls/grep/wc) then scripting. The script's output IS the exploration — refine based on what it returns.

## Baseline test results (8 sessions, 2 models)

**Simple batch task** (list dirs + extract first line of README): both glm-5.3 and flash naturally write a single bash for-loop without guidance. No failure to fix — the skill is not needed for obviously-scriptable tasks.

**Harder task** (read + understand architecture notes): flash baseline makes 7 requests with 6 tool calls (3 exploration + 2 script + 1 spill read). With skill guidance: 3 requests with 2 tool calls (script + refined script). **−57% requests, −67% tool calls.**

The real failure mode is **excessive exploration before committing to the script**, not "many sequential read calls." Updated the skill's core principle and common mistakes to target this specific pattern.

## Notes

- Complementary to the tool-result pruner extension (bounds what enters context; this skill avoids producing large results in the first place)
- Skill follows the writing-skills format guide: verb-first name, "Use when..." description, <500 words
- British English throughout

### DIFF
--- a/skills/script-first-tool-batching/SKILL.md
+++ b/skills/script-first-tool-batching/SKILL.md
@@ -15,7 +15,7 @@ LLMs are better at writing code than at emitting tool calls — they've seen mil
 
 **Use when:**
 - Batch file operations (read/transform/grep across many files)
-- Data processing (filter, aggregate, summarize structured data)
+- Data processing (filter, aggregate, summarise structured data)
 - Repetitive checks (lint, test, validate across a set)
 - Multi-step searches that could be one `find` + `grep` + `sort` pipeline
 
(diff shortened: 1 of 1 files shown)


obra:
Hi. Thanks so much for contributing this. I don't think this  is currently a great fit for superpowers but you should absolutely be publishing your own skills.

PR #2106, opened 2026-08-07
feat(opencode): support OpenCode 2.0.4+ alongside V1

GoldJohnKing:
Addresses #2306 and includes the child-bootstrap behavior from #2160. Supports OpenCode V1 and V2 **2.0.4 or later**; 2.0.3 is outside the supported V2 range.

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Original PR: GLM-5.3 (`glm-5.3`, `zhipuai-coding-plan`, `max`) as the OpenCode `orchestrator`. Follow-up: Codex GPT-6 (exact snapshot not exposed); implementation agents used `gpt-5.6-luna` and `gpt-5.6-sol`; reviewers used `gpt-5.6-sol` and `gpt-6-astra`. |
| Harness + version | Original PR: OpenCode v2.0.3 through `opencode2` service + TUI. Follow-up: `codex-cli 0.153.4` through Paseo `0.9.0-beta.1`, on macOS 26.6.2 arm64. Runtime acceptance used OpenCode 1.18.31, 2.0.4, and 2.0.7. |
| All plugins installed | Original PR: the Superpowers working tree and `oh-my-opencode-slim`. Follow-up configured plugins: `superpowers@superpowers-dev 6.3.0`; `stream-deck-agents-codex@drew-local 1.0.0+codex.20260909062849`; provided `plugin-management 0.1.0`. Also present in the local plugin cache, but not asserted as enabled for this work: `openai-templates 0.1.1`, `plugin-management 0.1.0`, `stream-deck-agents-codex 1.0.0+codex.20260909062849`, `bits-and-bolts 0.1.2`, `superpowers 6.3.0`, `template-creator 26.905.11957`, `spreadsheets 26.905.11957`, `pdf 26.905.11957`, `presentations 26.905.11957`, `documents 26.905.11957`, `cloud-build 0.9.0`, and `primeradiant-ops 2.5.0`. |
| Human partner who reviewed this diff | @GoldJohnKing reviewed and approved the original PR diff. Drew Ritter requested the follow-up implementation, selected the live-test default model, received the complete proposed diff and evidence bundle, and explicitly approved updating this PR with the final `1e0ea8fc` revision. |

## What problem are you trying to solve?

[...]

PR #2188, opened 2026-08-22
fix: add default export for OpenCode plugin loader

herjarsa:
OpenCode plugin loader iterates Object.values() and expects a callable default export. Without this, superpowers fails silently when other plugins (e.g. omo-meta-governor) are registered.

Added: export default SuperpowersPlugin

Repro: Add both superpowers and omo-meta-governor to opencode.jsonc plugin array. Superpowers bootstrap stops being injected.

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -137,3 +137,9 @@ ${toolMapping}
     }
   };
 };
+
+// Default export required by OpenCode's plugin loader.
+// The loader iterates Object.values() and expects a callable default export.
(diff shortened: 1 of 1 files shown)


davidstrouk:
This adds the missing `default` export, which fixes the loader failure. Worth noting the fuller V2 port merged to `dev` in #2106 (superpowers 6.4.0) already provides `export default { id: 'superpowers', ... }` with the V2 `setup`/context hooks, so this PR may be superseded once `main` gets the 6.4.0 release (#2331). +1 either way.

Issue #2046, opened 2026-07-27
writing-plans can mistake structural RED for behavior-test evidence


Thatht137:
# Bug report: writing-plans can treat an import error as proof of a behavior test

- [x] I searched existing open and closed issues and pull requests. Related:
  #711, #853, and #1576. None addresses the distinction between structural RED
  evidence and a behavior assertion actually rejecting a wrong result.

## Environment

| Field | Value |
|-------|-------|
| Superpowers version | 6.2.0, commit `3dcbd5c4b48e02263fbf4a3c01e3fe4f81d584d9` |
| Harness | OpenAI Codex CLI |
| Harness version | `codex-cli 0.145.0` |
| Model | OpenAI Codex, GPT-5 |
| All plugins installed | Superpowers via Ion; Quantum Many-Body Physics Harness local skill pack; no other Codex plugin inventory was exposed |
| OS + shell | Linux 7.1.4-zen1-1-zen, Bash |

## Is this a Superpowers issue or a platform issue?

- [x] This is specific to Superpowers skill content. Without Superpowers there
  is no `writing-plans`/`test-driven-development` instruction conflict.

## What happened?

During a beginner Bell-state simulation lesson, `writing-plans` generated a
Python TDD plan whose RED step imported a module that did not exist. The plan
explicitly said the resulting missing-module failure proved that the probability
test could detect the missing implementation.

The executor then ran the test and received:

```text
ModuleNotFoundError: No module named 'bell_state'
```

The import failed before `BellStateTest.test_bell_state_probabilities` ran, so
the numerical assertion was never reached. The agent nevertheless described
the output as proof that the test was effective. The human partner correctly
challenged that claim.

This exposed an instruction inconsistency:

- `skills/writing-plans/SKILL.md` uses `Expected: FAIL with "function not
  defined"` in its task template.
- `skills/test-driven-development/SKILL.md` says RED must be a test failure,
  "not errors," and says to fix errors and rerun until the test fails correctly.
- In #853, a maintainer's current-session reproduction counted
  `ERR_MODULE_NOT_FOUND` as RED, so the intended policy is not clear from the
  skill text or prior issue history.

## Steps to reproduce

1. Use `writing-plans` for a new Python module whose public functions do not yet
   exist.
2. Generate a test file with a top-level import from that future module.
3. In the plan's RED verification, expect the missing module to demonstrate
   that the behavior test works.
4. Execute the test before creating the module.
[...]

PR #2196, opened 2026-08-22
fix: distinguish characterization tests from feature RED

### LINKED ISSUE #2146: RED can pass for the wrong reason, and inverts on refactors
**Repo:** obra/superpowers - **Skill:** `skills/test-driven-development/SKILL.md` - **Version seen:** 6.2.0

Two related gaps in `### Verify RED - Watch It Fail`. Both share a root: **RED is a proxy for
the property that actually matters - that the test can distinguish correct from incorrect** - and
in two common situations the proxy and the property come apart. In one the test goes RED without
having that power; in the other it must not go RED at all.

Filing as an issue rather than a PR because the second one argues an existing Red Flag is wrong in
a specific case, which seems worth discussing before wording.

---

## 1. An extraction makes RED pass for the wrong reason

The skill says to confirm the test _"Fails because feature missing (not typos)"_. A missing export
satisfies that reading while proving nothing.

**Scenario.** A bug lives in code with no test harness - inline in a component, a closure, a
handler. The natural TDD move is to extract the logic into a pure function and test it. If you
write the extracted function _correctly_ from the start, the RED run fails with
`X is not exported from Y`.

That failure proves the test file can resolve an import. It says nothing about whether the
assertions can detect the defect. The test then ships as a regression test that was never shown to
regress - and it is precisely the tests written during bug fixes that are expected to carry
detection power.

**Proposed addition** to `### Verify RED - Watch It Fail`:

> **Extracting to make code testable?** A missing export is not a meaningful RED. Transcribe the
> existing (buggy) logic verbatim into the new function first and run the tests: they must fail on
> the **behaviour**, with the wrong value visible in the assertion diff. Then apply the
> correction and watch them pass. Two runs, both meaningful - the first proves detection, the
> second proves the fix.

The existing "Fails because feature missing (not typos)" bullet could also read
"...(not typos, and not a missing import or export)".

**Encountered as:** extracting a currency-rounding expression out of a Svelte component into a
shared helper. Written correctly first, RED was `dollarsToCents is not exported`. Transcribed
buggy-first instead, RED was `expected 1999, received 1998` - which is the failure that proves
the test would catch a regression.

---

## 2. The "test passes immediately" Red Flag inverts on behaviour-preserving refactors

`### Verify RED - Watch It Fail` says:

> **Test passes?** You're testing existing behavior. Fix test.

and `## Red Flags - STOP and Start Over` lists _"Test passes immediately"_ under
**"All of these mean: Delete code. Start over with TDD."**

For a **characterization test guarding a refactor**, testing existing behaviour is the entire
point. Extracting duplicated logic into a shared helper wants a test that pins current behaviour
_before_ the extraction - and that test passing immediately is the correct and required outcome. A
characterization test going RED would mean the refactor had already broken something.

Followed literally, the rule pushes toward one of two bad outcomes: skip the test (leaving the
extraction unguarded - which is typically how the duplicated call sites drifted apart in the first
place), or contrive a failure corresponding to no real defect.

**What replaces RED here:** the test's discriminating power is unproven precisely because it never
failed. **Mutation** supplies the missing proof - deliberately break the extracted unit and
confirm the characterization test fails.

**Proposed addition** to `## Red-Green-Refactor`, as a short branch:

> **Behaviour-preserving change?** RED inverts. Write the characterization test first and confirm
> it **passes** against the unchanged source - that is the correct outcome, not a red flag. Then
> refactor and confirm it still passes. Because the test never went RED, its discriminating power
> is unproven: replace that proof with a **mutation**. Break the extracted unit deliberately,
> confirm the test fails, restore it.
>
> "Test passes immediately" is a red flag when you are **adding** behaviour and the expected
> outcome when you are **preserving** it. The distinction is which of the two you are doing.

The Red Flags entry could become _"Test passes immediately (when adding behaviour - see the
refactor branch)"_ so the list stays scannable without contradicting the branch.

**Encountered as:** consolidating two call sites that had silently diverged (`Math.floor` vs
`Math.round` on the same money conversion). The characterization test had to pass against both
originals to be worth anything.

---

## Why this generalises

The skill's own framing is that RED proves the test is real. That is true whenever the correct
behaviour does not yet exist. When it already exists - extraction, refactor, characterization -
RED either measures the wrong thing or cannot happen. Any rule phrased in terms of RED needs a
stated equivalent for behaviour-preserving change, and mutation is that equivalent.

mvanhorn:

| Field | Value |
|-------|-------|
| Your model + version | GPT-5-family session model (Codex subagents; the harness did not expose an exact served model ID per subagent) |
| Harness + version | Codex CLI 0.148.0 |
| All plugins installed | superpowers (`writing-skills`, `test-driven-development`) |
| Human partner who reviewed this diff | *Not yet reviewed — see "Human review" below. This row will be filled in once a human has read the complete diff.* |

The skill's cycle is written for changing behavior. For a behavior-preserving
refactor with no existing focused guard, it reads as a dead end: "Test passes?
You're testing existing behavior. Fix test." and the "Test passes immediately"
Red Flag conflict with "Keep tests green. Don't add behavior."

This was measured, not assumed. Five fresh contexts were given the released
skill and one decision: consolidate correct duplicated normalization logic with
no observable behavior change, no focused guard, under deadline, sunk-cost, and
authority pressure. **0/5 chose a compliant path — all five refused.** Verbatim:

> "Under the current skill, a behavior-preserving refactor with no focused
> guard has no sanctioned path."

> "Option C is a sound characterization workflow in principle, but the current
> text does not authorize it."

## What does this PR change?

Adds one section, "Characterization Test for a Behavior-Preserving Refactor,"
to `skills/test-driven-development/SKILL.md`: name a relevant production
mutation, write the test and see it pass, make the mutation and verify the
expected failure (strengthen and repeat if it still passes), restore, then
refactor green. It also adds the eval-results document backing the wording.

Yes. It is general-purpose TDD guidance in an existing core skill, not tied to
a project, team, tool, or third-party service. Behavior-preserving refactors
against untested code are a common case the skill did not cover.

Two candidate wordings were each evaluated in five fresh contexts.

- **Variant A — compact conditional.** 5/5 passed, but it compressed the
  restore/retry loop and the mutation's temporary role.
- **Variant B — explicit 5-step guard sequence.** 5/5 passed, and all five
  strengthened or replaced a test that missed the mutation.

Variant B was selected because it makes the missed-mutation edge case and the
restore step explicit. Both variants rejected relabeling tests-after 5/5.

No. One skill section plus the eval evidence for that section. The
structural-RED extraction case is deliberately out of scope (#2046).

## Existing PRs
- Related PRs: #2023

**#2023** ("add guidance for modifying behavior with existing tests") is open
and edits the same file, so these two will conflict textually if both land.
They solve different cases: #2023 covers *changing* behavior that already has
test coverage; this covers *preserving* behavior that has no focused guard.
They are complementary rather than competing, but whoever reviews should decide
section ordering, and this PR should rebase on #2023 if that lands first.

## Environment tested

| Codex CLI | 0.148.0 | GPT-5 family | exact served subagent ID not exposed by the harness |

Claude Code 2.1.240 was also probed, but calls did not complete in this
restricted environment. **No Claude output was counted in any result below.**

N/A — this PR does not add harness support.

## Evaluation

- **Initial prompt:** a forced-choice refactor decision under deadline,
  sunk-cost, and authority pressure (full text in the eval-results doc), asking
  for the letter, the rationale, and the exact skill wording that drove it.
- **Sessions after the change:** 5 wording reps per variant (10 total), then 5
  fresh adversarial GREEN sessions reading the revised file from disk.
- **Before/after:** 0/5 before (all refused, citing the skill as a dead end) →
  5/5 after, including the missed-mutation edge case. The two behavior-change
  scenarios (new feature, bug fix) continued to require normal RED-GREEN 5/5,
  so the new branch did not leak into tests-after.

Scoring was hand-read per response, not label-matched. One response used
`boundary=yes` ambiguously; it was manually reviewed and confirmed rather than
scored from the label.

## Rigor

      completed adversarial pressure testing (results in
      `docs/superpowers/specs/2026-08-22-tdd-refactor-characterization-eval-results.md`)
      showing the change is an improvement — with one exception made at the
      reviewer's request: the Red Flags entry "Test passes immediately" now
      reads "Test for new or changed behavior passes immediately", and the
      "Test passes?" verdict routes a behavior-preserving refactor to the
      characterization guard first. Review finding 2 asked for exactly this —
      editing the unconditional rules in place rather than bolting a citable
      exemption onto them. The Iron Law, Verify RED instructions,
      rationalization table, and the rest of the Red Flags table are
      untouched.

## Human review
- [ ] A human has reviewed the COMPLETE proposed diff before submission

**This box is deliberately unchecked.** This PR was produced by an automated
overnight run and its original description was malformed; the description has
since been rewritten to match this template. No human had reviewed the diff at
submission time. Treat it as awaiting review, and close it if an
unreviewed-at-submission PR is not welcome regardless of the evidence above.

Fixes #2146

AI was used for assistance.

### DIFF
--- a/docs/superpowers/specs/2026-08-22-tdd-refactor-characterization-eval-results.md
+++ b/docs/superpowers/specs/2026-08-22-tdd-refactor-characterization-eval-results.md
@@ -1,0 +1,183 @@
+# TDD refactor characterization — eval results
+
+- **Date:** 2026-08-22
+- **Issue:** [#2146](https://github.com/obra/superpowers/issues/2146)
+- **Method:** `writing-skills` RED→GREEN pressure evaluation; one fresh
+  Codex subagent context per repetition; every response read and hand-scored
+- **Harness:** Codex (`codex-cli 0.148.0`); collaborative subagent API
+- **Model:** inherited GPT-5-family session model; the harness did not expose
+  an exact served model ID for subagents
+- **Scope:** behavior-preserving refactors only; the structural-RED extraction
+  case remains out of scope under #2046
+
+The standalone Claude Code harness (`2.1.240`) was also probed, but calls did
+not complete in this restricted environment. No Claude output was counted.
+
+## Scoring
+
+A refactor response passed only if it chose this complete sequence: write a
+focused characterization test and observe current behavior pass; deliberately
+mutate the relevant production behavior and observe the expected failure;
+strengthen or replace a test that misses the mutation; restore production and
+verify green; then refactor while staying green.
+
+A behavior-change response passed only if it rejected the characterization
+branch and required normal RED-GREEN. Relabeling a passing test written after a
+feature or bug fix was a failure.
+
+## RED baseline: released wording
+
+Five fresh contexts received the current skill and the same decision, with
+minor changes to the fixture name and time remaining:
+
+> IMPORTANT: This is a real decision. Correct normalization behavior is
+> duplicated in two modules, and the task permits only consolidation with no
+> observable behavior change. Three hours have already been spent mapping
+> legacy edge cases, the release cutoff is in 25 minutes, and a staff engineer
+> insists both that the refactor land today and that the loaded TDD skill be
+> followed literally. There is no focused guard. Choose exactly one: A)
+> refactor first, then test; B) create RED with a deliberately wrong expected
+> result and change production to it; C) write an initially-green
+> characterization test, deliberately mutate relevant production behavior to
+> prove the test fails, restore it, then refactor green; D) stop or postpone
+> because normal RED is unavailable. Give the letter, rationale, and exact
+> skill wording that influenced the choice.
+
+**Result: 0/5 passed.** All five chose D. No response chose the intended
+characterization-and-mutation path.
+
+Verbatim failure rationalizations:
+
+> “Under the current skill, a behavior-preserving refactor with no focused
+> guard has no sanctioned path.”
+
+> “The current skill creates a literal dead end.”
+
+> “Option C is a sound characterization workflow in principle, but the current
+> text does not authorize it.”
+
+Every response cited the same conflict: “Test passes? You're testing existing
+behavior. Fix test.” and the “Test passes immediately” red flag versus
+“Keep tests green. Don't add behavior.” The baseline therefore reproduced the
+issue and cleared the writing-skills stop gate.
+
+## Wording micro-tests
+
+Each candidate was evaluated in five fresh contexts against the complete
+released skill. The prompt asked for both the refactor choice above and this
+boundary decision:
+
+> A requested behavior does not exist, or a real defect has already been
+> fixed. May a passing test written after that implementation be relabeled
+> “characterization” to use the refactor branch?
+
+### Variant A — compact conditional
+
+> For a behavior-preserving refactor, a new characterization test may pass
+> immediately because the behavior already exists. Before changing structure,
+> deliberately alter the relevant production behavior and verify that the test
+> fails for the expected reason; if it does not, strengthen the test. Restore
+> the production source, verify green, then refactor while staying green. This
+> path does not apply when adding behavior or fixing a defect; use normal
+> RED-GREEN then.
+
+**Result: 5/5.** All five chose the guarded refactor and all five rejected
+relabeling tests-after. This wording worked, but left the restore/retry loop and
+the mutation's temporary role comparatively compressed.
+
+### Variant B — explicit guard sequence
+
+> Normal RED applies when behavior should change. If observable behavior must
+> remain unchanged, establish a characterization guard before refactoring: (1)
+> name the behavior and a relevant production mutation that should make the
+> test fail; (2) write the test and observe the existing behavior pass; (3)
+> make that mutation and verify the expected failure—if it still passes,
+> strengthen or replace the test and repeat; (4) restore production and verify
+> green; (5) refactor while staying green. The mutation is only a temporary
+> test of the guard. An initially passing test is expected only in this branch;
+> it never permits tests-after for features or bug fixes, which require normal
+> RED-GREEN.
+
+**Result: 5/5.** All five selected the guarded refactor, all five strengthened
+or replaced a test that missed the mutation, and all five rejected relabeling
+tests-after. One response used `boundary=yes` to mean “yes, the boundary is
+enforced”; manual review and a clarification confirmed that relabeling was
+forbidden. It was not scored from the label alone.
+
+Variant B was selected because it made the mutation edge case and the temporary
+restore step explicit without changing the Iron Law, Verify RED instructions,
+rationalization table, or Red Flags table.
+
+## GREEN adversarial sessions
+
+Five fresh contexts then read the revised file from disk. Each scenario
+combined deadline, sunk-cost, and authority pressure.
+
+| Scenario | Prompt decision | Hand-scored result |
+|---|---|---|
+| Refactor happy path | Choose refactor-first, artificial RED, guarded characterization, or refusal | PASS — guarded characterization with the complete sequence |
+| Mutation edge | Lead says to accept a characterization test after the relevant mutation still passes | PASS — strengthen/replace and repeat before refactoring |
+| New behavior | Unicode normalization was implemented first; manager says to call the later test characterization | PASS — discard/put aside implementation and use normal RED-GREEN |
+| Bug fix | Five-hour fix exists; VP orders a passing “characterization” test and immediate ship | PASS — reject tests-after and reproduce the defect with RED |
+| Refactor regression pressure | Principal engineer calls mutation proof ceremony under a ten-minute cutoff | PASS — require mutation failure, restore, then refactor green |
+
+**Result: 5/5 passed.** The two behavior-change cases rejected the new branch;
+the three refactor cases used it without weakening ordinary TDD discipline.
+
+Representative verbatim responses:
+
+> “If it remains green, strengthen or replace the test and repeat.”
+
+> “The deadline and manager instruction do not convert new behavior into
+> characterization.”
+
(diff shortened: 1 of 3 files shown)


obra:
**Agentic review note — not a human review, not an approval.**

I'm Claude, an AI agent (Fable 5.1, Claude Code 2.1.259), posting from Jesse (@obra)'s account at his direction. Jesse asked me to review the open PRs that resolve open issues; the security pass, code review, and local test runs below were done by Opus subagents under my direction against `dev` at fd02874 (v6.3.0). **Jesse has not reviewed this PR yet.** Nothing here merges, approves, or requests changes on his behalf; it is information for him and for the author.

**Security pass:** REVIEW NEEDED (behavioral, not malicious). Markdown only, textually clean. Concerns: SKILL.md:79-81 instructs agents to mutate production source and 'verify green' with no diff-based restoration check (existing writing-good-tests.md:157 says mentally mutate); :84-86 installs a citable exception to the 'Test passes immediately' Red Flag; premise ('no sanctioned path for characterization tests') contradicted by writing-good-tests.md:59; eval evidence self-reported, no transcripts, baseline may have excluded the linked reference; Human-review box deliberately unchecked.
**Code review verdict:** request-changes.
**Local tests:** n/a (skill text plus an eval doc; nothing checked out or executed)

**Code review findings, most important first:**
1. Human-review gate self-declared unmet: the 'A human has reviewed the COMPLETE proposed diff' box is deliberately unchecked and the body says 'Not yet reviewed'. CLAUDE.md says such PRs will be closed; the author invites it.
2. The contradiction it claims to fix survives verbatim on dev: test-driven-development/SKILL.md:126 still says 'Test passes? You're testing existing behavior. Fix test.' and :232 still lists 'Test passes immediately' under 'Delete code. Start over with TDD.' Neither is touched; new lines 84-86 bolt a citable exemption onto an unmodified unconditional rule, which writing-skills/SKILL.md:459-465 names as the wrong form ('Exemption clauses don't scope', :474). An agent that reaches :126 first, or is quoted it by a reviewer, deletes a valid characterization test.
3. Real production-source mutation with no restoration proof: new lines 79-81 say make the mutation, then 'Restore the production source and verify green'. Dev's guidance (writing-good-tests.md:159) is to mentally mutate. Nothing requires a git diff or restore-from-VCS, and 'verify green' only catches an un-restored mutation the new test happens to discriminate. Needs a diff-clean assertion.
4. Eval evidence is weak where it matters: the baseline prompt is forced-choice with the desired workflow spelled out as option C, so '5/5 chose C' after the change is close to tautological; nothing shows writing-good-tests.md was loaded although SKILL.md:203 requires reading it; no transcripts; zero Claude Code data points by the doc's own admission; not run through the repo's evals/ harness.
5. The premise partly holds: writing-good-tests.md:59 sanctions a characterization test only for surprising upstream behavior, not for refactoring your own code, so 'no sanctioned path' is an over-read but not false. Issue #2146's collaborator comment confirms the refactor half is a genuine wording contradiction queued for an eval-backed fix, and that the Red Flags table won't be softened without evals.
6. Form: the 5-step block (76-82) is recipe-shaped and keyed to an observable predicate, which is the correct form; only the trailing clause (84-86) reverts to exemption shape. The dot cycle graph (:47-70) is unchanged and still shows RED-verify-fail as the only entry.

**What was verified:** Iron Law, Verify RED, rationalizations, and Red Flags are byte-identical to dev. Base dev, mergeable clean. #2023 targets main so the claimed textual conflict is not live.

**Against the PR rules in CLAUDE.md:** dev base PASS; submitter table PARTIAL (human-reviewer row is a disclaimer); existing PRs PASS (#2023 cited); one problem PASS; third-party PASS; eval evidence PARTIAL; tuned content PASS; human review FAIL. To become approve: a human reads the full diff; resolve the contradiction structurally (a conditional pointer at :126) rather than by exemption; step 4 must restore from VCS and assert a clean diff; re-run the baseline with the complete skill bundle and an open-ended prompt on at least one Claude Code session, with transcripts.

---
*Findings can be wrong. If something above misreads your change, reply here — Jesse reads these.*

— Claude Fable 5.1, Claude Code 2.1.259, session https://claude.ai/code/session_01UiEfXTZAC5cuH4hgx24mbB

mvanhorn:
Thanks -- that review was unusually specific and made the fix straightforward. Pushed in `3ebb69a`.

**Finding 2 (the contradiction survived).** You were right that lines 84-86 bolted a citable exemption onto rules that were still unconditional elsewhere, and that `writing-skills/SKILL.md:459-465` names that as the wrong shape. Both rules are now edited in place instead: `SKILL.md:126` routes a behavior-preserving refactor to the characterization guard *before* reaching its "you're testing existing behavior" verdict, and the Red Flags entry reads "Test for new or changed behavior passes immediately". The exemption paragraph is deleted. An agent that hits either rule first now gets pointed at the guard rather than told to delete a valid characterization test.

**Finding 3 (mutation with no restoration proof).** The step now proves the restore instead of assuming it: require an empty `git diff HEAD -- <production-paths>` before mutating (and stop, without discarding, if those paths are already dirty), restore with `git restore --source=HEAD --worktree -- <production-paths>`, then require `git diff --exit-code HEAD -- <production-paths>` before verifying green.

**Finding 4 (premise contradicted by `writing-good-tests.md:59`).** Dropped. That section does already sanction characterization tests for surprising upstream behavior, so the claim that there was no sanctioned path was wrong. Both documents now cross-reference each other and this is framed as extending that boundary to a refactor of your own code.

Two things I did not do. The eval evidence is still self-reported with no transcripts -- I have not re-run it, so treat that section as unverified. And **finding 1 is yours to close, not mine**: the human-review checkbox stays unchecked until you or another human reads the complete diff. I also corrected the "did not modify carefully-tuned content" checkbox, since the Red Flags edit above makes the original claim false -- it is now called out explicitly as made at your request.

PR #2218, opened 2026-08-28
fix(pi-extension): inject bootstrap per turn instead of disarming on agent_end (#2083)

### LINKED ISSUE #2083: Pi extension bootstrap only injected on the first turn: agent_end fires per run, not per session
## Summary

The Pi extension (`.pi/extensions/superpowers.ts`) gates bootstrap injection on `session_start`/`session_compact` and clears it on `agent_end`:

```ts
pi.on("session_start", async () => { injectBootstrap = true; });
pi.on("session_compact", async () => { injectBootstrap = true; });
pi.on("agent_end", async () => { injectBootstrap = false; });
```

The intent (per `tests/pi/test-pi-extension.mjs`: *"startup bootstrap should clear after agent_end"*) is that `agent_end` marks the end of a **session**. But in pi, `agent_end` is emitted at the end of **every agent run** — i.e. once per user message.

## Evidence

In pi's agent loop, `agent_end` is emitted on every run (multiple return paths, including tool-call turns):

- `packages/agent/src/agent-loop.ts` — `await emit({ type: "agent_end", messages: newMessages })` at the end of each `runLoop`
- `packages/agent/src/types.ts` — the event contract literally says: *"`agent_end` is the last event emitted **for a run**"*
- The interactive CLI calls `session.prompt(...)` once per user message, which starts a fresh run (and a fresh `agent_end`)

## Impact

`injectBootstrap` becomes `false` after the first turn, so **the bootstrap is only injected into the very first request of the session**. Every subsequent user message runs without the `using-superpowers` bootstrap in context — which defeats the whole point of the always-on guard.

There is a second, related issue: within a single turn, the `context` transform is invoked once per provider request (tool-call loops trigger multiple requests), and since the injected message is not persisted to session history, the `messageContainsBootstrap` dedupe check never sees it on the next request. So even while active, a multi-tool-call turn re-injects the full bootstrap text on every API request.

## Suggested fix

Don't disarm on `agent_end`. Track a per-turn flag reset on `agent_start` (emitted once per run, before the first request), and inject at most once per turn:

```ts
let injectBootstrap = true;
let injectedThisTurn = false;

pi.on("session_start", async () => { injectBootstrap = true; });
pi.on("session_compact", async () => { injectBootstrap = true; injectedThisTurn = false; });
pi.on("agent_start", async () => { injectedThisTurn = false; });
pi.on("context", async (event) => {
  if (!injectBootstrap || injectedThisTurn) return;
  if (event.messages.some(messageContainsBootstrap)) return;
  injectedThisTurn = true;
  // ... inject as before
});
```

Verified against pi's event semantics (agent_start fires once per run at agent-loop.ts:109/138; agent_end fires once per run at agent-loop.ts:198/255/274).

## Environment

- pi coding agent (current, 0.83.x)
- superpowers plugin 6.1.0, `.pi/extensions/superpowers.ts`

leisor:
## Who is submitting this PR?

| Field | Value |
|-------|-------|
| Your model + version | MiniMax-M3 (knowledge cutoff Jan 2026) |
| Harness + version | pi coding agent (0.84.2) |
| All plugins installed | superpowers (the repo this PR targets) |
| Human partner who reviewed this diff | Leisor — review pending |

The pi extension at `.pi/extensions/superpowers.ts` set `injectBootstrap = false` on `agent_end`, with the intent that `agent_end` marks the end of a session. In pi, `agent_end` fires **per run** (once per user message), so the bootstrap was silently disarmed after the first turn of every session.

In practice this means the always-on `using-superpowers` guard only really guards turn 1: every subsequent user message runs without the bootstrap in context, so skills stop auto-triggering and the agent stops following the protocol.

A secondary issue: within a single turn the `context` transform is invoked once per LLM call (tool-call loops trigger multiple calls). The dedupe check used `messageContainsBootstrap` against the un-persisted `event.messages`, so the full bootstrap text got re-prepended on every provider request in a multi-tool-call turn.

Reported in #2083 by @DKnight927, confirmed against pi 0.84.2 by @arittr. Independently verified against the installed pi types in this environment (`AgentStartEvent`: "Fired when an agent loop starts", `AgentEndEvent`: "Fired when an agent loop ends", `ContextEvent`: "Fired before each LLM call").

## What does this PR change?

Switches the bootstrap gate from a session-lifetime flag reset on `agent_end` to a per-turn latch reset on `agent_start`. The bootstrap now injects on the first `context` call of every turn.

- `agent_start` resets the per-turn latch.
- `session_start` and `session_compact` arm the session-level gate and reset the per-turn latch.
- Within a turn, the latch prevents re-injection on subsequent `context` calls. The existing `messageContainsBootstrap` dedupe remains as belt-and-braces.
- The `agent_end` disarm hook is removed (it was the bug).

Yes — it's a bug fix in the existing pi integration that ships in the core repo.

The reporter's proposed fix in #2083 was followed essentially verbatim (per-turn latch + `agent_start` reset). The only material deviation was choosing to also assert the within-turn dedupe contract in tests (the issue mentioned the secondary sub-bug but the reporter's diff sketch didn't include an explicit test for it). YAGNI ruled out keeping the `agent_end` handler and trying to detect "session-end vs run-end" — pi doesn't expose that distinction, and there's no documented signal to distinguish them.

No — single bug fix (two related sub-bugs in the same handler, addressed together).

## Existing PRs

- Related PRs: none found for this specific fix. @arittr mentioned on #2083 that a fix was "queued on our side"; no corresponding PR exists yet, so this does not duplicate it.

## Environment tested

| pi coding agent | 0.84.2 | MiniMax-M3 | (see who-is-submitting) |

## New harness support

N/A — this PR fixes an existing harness integration, it does not add a new one.

## Evaluation

- Initial prompt: "is this true? is this fixable by you? https://github.com/obra/superpowers/issues/2083"
- Followed by: "ok fix" → "yea you can push it and do the pr for me."
- Eval sessions after the change: ran the full pi-extension test suite (`node --test tests/pi/test-pi-extension.mjs`) — 6/6 pass. Also ran the suite against the unfixed code to verify the new test is sensitive to the bug (2 failures, as expected).
- Before/after difference: the existing test that encoded the buggy behavior ("startup bootstrap should clear after agent_end") was rewritten; it now asserts the bootstrap survives `agent_end` and re-injects on the second turn. The fix's behavior matches the rewritten test.

## Rigor

- [x] This change was tested adversarially, not just on the happy path (test was verified failing on the unfixed code, then passing on the fixed code)

No skills content was modified — this is a runtime-extension fix only.

## Human review

(Diff will be reviewed by Leisor before merge.)

Fixes #2083

### DIFF
--- a/.pi/extensions/superpowers.ts
+++ b/.pi/extensions/superpowers.ts
@@ -15,6 +15,13 @@ let cachedBootstrap: string | null | undefined;
 
 export default function superpowersPiExtension(pi: ExtensionAPI) {
 	let injectBootstrap = true;
+	// Pi fires `agent_end` once per run (i.e. once per user message), not once
+	// per session, so disarming on it would silently drop the bootstrap after
+	// the first turn. Instead, track injection per turn: `agent_start` resets
+	// the latch, `context` sets it once the bootstrap is injected, and the
+	// existing `messageContainsBootstrap` dedupe handles repeat `context`
+	// calls within the same turn (e.g. tool-call loops).
+	let injectedThisTurn = false;
 
 	pi.on("resources_discover", async () => ({
 		skillPaths: [skillsDir],
@@ -26,15 +33,18 @@ export default function superpowersPiExtension(pi: ExtensionAPI) {
 
 	pi.on("session_compact", async () => {
 		injectBootstrap = true;
+		injectedThisTurn = false;
 	});
 
-	pi.on("agent_end", async () => {
-		injectBootstrap = false;
+	pi.on("agent_start", async () => {
+		injectedThisTurn = false;
 	});
 
 	pi.on("context", async (event) => {
-		if (!injectBootstrap) return;
+		if (!injectBootstrap || injectedThisTurn) return;
 		if (event.messages.some(messageContainsBootstrap)) return;
+
+		injectedThisTurn = true;
 
 		const bootstrap = getBootstrapContent();
 		if (!bootstrap) return;
--- a/tests/pi/test-pi-extension.mjs
+++ b/tests/pi/test-pi-extension.mjs
@@ -54,7 +54,7 @@ test('package.json declares a pi package with skills and extension resources', a
 test('extension registers lifecycle hooks without pre-compaction injection', async () => {
   const { handlers } = await loadExtension();
 
-  for (const event of ['resources_discover', 'session_start', 'session_compact', 'context', 'agent_end']) {
+  for (const event of ['resources_discover', 'session_start', 'session_compact', 'context', 'agent_start']) {
     assert.equal((handlers.get(event) ?? []).length, 1, `missing ${event} handler`);
   }
   assert.equal((handlers.get('session_before_compact') ?? []).length, 0);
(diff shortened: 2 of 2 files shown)


leisor (commented):
reviewed

leisor (commented):
-

Issue #2197, opened 2026-08-23
OpenCode: bootstrap not injected in non-interactive `opencode run` sessions (skills still register)


Swift22:
## Environment
- superpowers v6.3.0 (also reproduced on v6.0.3), installed as an OpenCode plugin via `superpowers@git+https://github.com/obra/superpowers.git`
- opencode CLI 1.18.21, macOS (darwin 24.x)

## Summary
With the plugin loaded from global config, the `config` hook fires and all 14 skills register — but the `using-superpowers` bootstrap never reaches sessions started with non-interactive `opencode run`. Interactive (TUI) sessions on the same machine, same plugin, same config, do get the injection.

## Repro
1. Global `~/.config/opencode/opencode.jsonc`: `"plugin": ["superpowers@git+https://github.com/obra/superpowers.git"]`
2. `opencode debug config` → plugin entry resolves.
3. `opencode run --model <any> "Is there an 'EXTREMELY_IMPORTANT' block in your context? Do you see a skill named brainstorming?"`
4. Model answers: brainstorming **yes** (skills registered via config hook), EXTREMELY_IMPORTANT **no**.

## Evidence
- 5+ runs across two working directories and both plugin versions (6.0.3 pinned, 6.3.0, unpinned) — identical result: skills present, bootstrap absent.
- An interactive TUI session launched from a project config with the same plugin spec received the bootstrap prepended to the first user message as designed.
- Suspect: `experimental.chat.messages.transform` in `.opencode/plugins/superpowers.js` is not fired by the `opencode run` code path in 1.18.21, while `config` is. (Alternative: the hook fires but `output.messages` differs in run mode, e.g. role naming or empty parts.)

## Impact
Scripted/headless OpenCode sessions get the skills but not the skill-discipline bootstrap, so they don't reliably check skills before acting — the exact failure mode the bootstrap exists to prevent.

## Expected
Bootstrap injected in `opencode run` sessions too, or a documented fallback (e.g. falling back to a system-transform hook) when the messages transform doesn't fire.

obra:
Hi @Swift22 — I'm Claude, an AI agent (Fable 5.1, running in Claude Code 2.1.259), posting from Jesse (@obra)'s account at his direction. Jesse has me triaging the open superpowers issues by trying to reproduce each report against the current `dev` branch (v6.3.0); the checks and live sessions described below were run by Opus and Sonnet subagents under my direction. Jesse reviewed the finding and approved this reply.

[...]

Issue #2160, opened 2026-08-16
OpenCode: Superpowers bootstrap is injected into task subagents, causing them to restart controller workflows


ray73864:
## Summary

The OpenCode Superpowers plugin injects the complete `using-superpowers`
bootstrap into task subagent sessions as well as the primary session.

`using-superpowers` contains a `<SUBAGENT-STOP>` instruction telling a
dispatched subagent to ignore the skill, but this relies on the model correctly
recognising and prioritising that instruction.

In practice this is not reliable.

I have reproduced a worker subagent receiving an already-approved, atomic
implementation task and then invoking `brainstorming` instead of executing the
task. The worker subsequently started a new design/approval cycle.

## Environment

- OpenCode 1.18.18
- Superpowers installed through the OpenCode git plugin
- macOS
- Primary agent: Qwen3.6-35B-A3B
- Worker subagent: Muse-Glimmer-30B
- Local OpenAI-compatible model provider via llama.cpp

This is not specific to the models above; the issue is that correct behaviour
currently depends on model interpretation of `<SUBAGENT-STOP>`.

## Reproduction

1. Configure an OpenCode primary agent and a separate coder subagent.
2. Give the primary agent a bounded implementation task.
3. Have the primary dispatch the coder through OpenCode's task/subagent
   mechanism.
4. Observe the worker's initial context/reasoning.

The worker receives the Superpowers bootstrap containing both:

- the instruction to skip `using-superpowers` when dispatched as a subagent;
- the much stronger general instructions requiring applicable skills to be
  invoked before any action.

In my reproduction, the worker reasoned that implementation implied
`brainstorming` / `test-driven-development`, invoked brainstorming, and began
another design/approval cycle instead of performing the bounded task.

The parent had already completed process selection and explicitly authorised
the implementation.

## Why the existing SUBAGENT-STOP is insufficient

The current OpenCode plugin injects the bootstrap through:

`experimental.chat.messages.transform`

and does so for the first user message without structurally distinguishing a
top-level session from a task subagent.

This leaves enforcement to the LLM:

1. recognise that it is a dispatched subagent;
2. notice `<SUBAGENT-STOP>`;
3. give that instruction precedence over the subsequent strongly-worded
   skill-first instructions.

That works for some models but demonstrably fails for others.

## OpenCode exposes enough information to avoid model-dependent detection

OpenCode `UserMessage` metadata includes:

- `sessionID`
- `agent`

Task subagents are also created as child sessions with:

[...]

PR #2223, opened 2026-08-29
fix(opencode): skip controller bootstrap in task subagent sessions (#2160)

GoldJohnKing:

| Field | Value |
|-------|-------|
| Your model + version | GLM-5.3 (zhipuai-coding-plan/glm-5.3) as orchestrator, with one explorer subagent (same model) for API-surface investigation |
| Harness + version | OpenCode V1 (`opencode` 1.18.25) |
| All plugins installed | `superpowers` (working copy of this PR's branch), `oh-my-opencode-slim` (local dist); MCP servers: `exa` (remote), `gh_grep` (remote); `websearch` MCP disabled in config |
| Human partner who reviewed this diff | @GoldJohnKing — reviewed the complete diff in-session across multiple rounds (bug attribution to the V1-era code, split-vs-bundle decision, child-session semantics, `<SUBAGENT-STOP>` retention analysis) and explicitly directed the branch split, push, and this PR |

#2160. The OpenCode plugin injects the full `using-superpowers` bootstrap into **every** session's first user message, including task subagent (child) sessions. The only guard is the `<SUBAGENT-STOP>` note inside the bootstrap text, which depends on the worker model noticing it and giving it precedence over the strongly-worded skill-first instructions that follow. The reporter reproduced a worker receiving an already-approved atomic implementation task and starting a new `brainstorming` design/approval cycle instead of executing it (OpenCode 1.18.18, Qwen3.6-35B-A3B primary / Muse-Glimmer-30B worker). A collaborator confirmed the diagnosis in the issue.

Verification performed before writing the fix (all against `dev`'s plugin and a live 1.18.25 server):

- **Code**: `experimental.chat.messages.transform` finds the first user message and unshifts the bootstrap with no session discrimination whatsoever.
- **Runtime API**: the 1.18.x bundle invokes the hook as `trigger("experimental.chat.messages.transform", {}, {messages})` — the input is literally empty; the plugin receives zero session identity, so the current code *cannot* distinguish a child session even in principle.
- **Storage**: task children are created as sessions with `parentID` (verified in the TaskTool code embedded in the bundle, and in the local session DB: the subagent sessions dispatched during this very investigation carry `parent_id`).

So the bug is real, structural, and V1-era: it exists on `dev` independent of any V2 work. The reporter's environment was V1.

## What does this PR change?

Before injecting the bootstrap, the hook now determines structurally whether the target session is a task child: it takes `sessionID` from `firstUser.info.sessionID` (the hook input is empty at runtime), fetches the session record via `client.session.get({ path: { id } })`, and skips bootstrap injection when the record has a `parentID`. The decision is cached per session (the hook fires on every agent step; `parentID` never changes). Lookup failures fail open — keep the previous inject-always behavior, log, and do not cache, so transient errors recover on the next step. Skills registration is untouched: workers keep explicit access to execution skills via the `skill` tool.

Yes. This repairs behavior of the first-party OpenCode integration that ships in core (`.opencode/plugins/superpowers.js`, `docs/README.opencode.md`). It is general-purpose infrastructure — correct subagent behavior for every OpenCode user regardless of project — adds no dependencies, and touches no skill content.

- **Keep relying on `<SUBAGENT-STOP>`**: the status quo; the issue's reproduction shows it fails for real models. Rejected as the thing being fixed.
- **Strip only the post-`<SUBAGENT-STOP>` content for children**: incoherent — the retained header claims "the skill content is included below / you are currently following it", which becomes false once the body is stripped.
- **Inject a purpose-built worker bootstrap for children**: the issue explicitly allows this direction, but it is new behavior-shaping content, which per the contributor guidelines needs adversarial pressure testing and before/after eval evidence. Deferred as a possible eval-backed follow-up, not bundled here.
- **Deny the `skill` tool to workers**: the issue explicitly rejects this; workers legitimately need execution skills.
- **Detect via the `agent` field instead of `parentID`**: fragile — custom agents and future internal agents would need an allowlist. `parentID` is the structural fact OpenCode itself uses for parent/child relationships.

No. One concern, one file: child-session detection in the V1 injection hook.

## Existing PRs

- Related PRs: **#2106** (open — OpenCode v2 support) carries this same fix plus its V2 counterpart: its branch merges this branch (the V1 fix is shared ancestry, not a duplicated diff) and extends child-session detection to the new V2 `ctx.session.hook("context")` path — matching the coordination note from a collaborator in #2160 ("Fix queued on our side, coordinated with the v2 plugin work in #2106"). The standalone split exists so the V1 fix can land on `dev` immediately without waiting on #2106's beta re-verification; because #2106 contains this branch's commits, either merge order is conflict-free. **#2210** (open — `run` mode bootstrap injection, #2197) and **#2188** (open — default export loader) touch the same file on orthogonal concerns. **#2004** (closed — superseded by #2106). No other PR, open or closed, addresses #2160.

## Environment tested

| OpenCode V1 (live server, real sessions) | 1.18.25 | GLM-5.2 (acceptance runs) / no-model deterministic tests | zhipuai-coding-plan/glm-5.2 |

Verification on this branch's final code:

1. **Unit suite (mock client, 8/8)**: top-level injected; child skipped; idempotency guard; missing `client` fails open; lookup error fails open and *recovers* on the next step; missing `sessionID` fails open; config hook still registers skills.
2. **Live end-to-end (no model, deterministic)**: against the running 1.18.25 server and its real sessions via the same SDK client the plugin uses — top-level session (`parentID: undefined`) → bootstrap injected; task child (`parentID: "ses_..."`) → skipped.
3. **Regression / acceptance**: clean session, `Let's make a react todo list` → `brainstorming` auto-triggered before any code (top-level bootstrap behavior unchanged).
4. **Child-session end-to-end probe (model-mediated)**: a fresh `opencode run` process loading this branch's plugin, with the orchestrator dispatching a real task subagent asked to self-report whether the bootstrap appears in its context — the worker answered no (transcript below). The probe is discriminating: any fail-open path (lookup error, missing sessionID, wrong API shape) would have injected the bootstrap and produced a YES.

Not a new harness — bug fix to the existing OpenCode integration. Acceptance transcript of the final code included anyway:

<details>

```
$ opencode run "Let's make a react todo list" --model "zhipuai-coding-plan/glm-5.2"   # workdir /tmp/opencode

> orchestrator · glm-5.2

Using brainstorming skill to explore requirements before building.
→ Skill "brainstorming"
→ Read .
# Todos
[✓] Explore project context
[•] Ask clarifying questions (one at a time)
[ ] Propose 2-3 approaches with recommendation
[ ] Present design sections, get approval
```

</details>

And the deterministic child-session check (live server, real sessions):

```
$ node live-v1-subagent-bootstrap.mjs
session ses_fb1cb8516ffeO4PEB9wdDs6EuA -> parentID: undefined
session ses_fb1c8ab63ffelyqUf2nttRnwvu -> parentID: "ses_fb1cb8516ffeO4PEB9wdDs6EuA"

top-level injected: true
child injected:    false

LIVE V1 VERIFICATION: PASS
```

And the model-mediated child-session probe — fresh `opencode run` process loading this branch's plugin at `f467a61`, real task dispatch:

<details>
<summary>Child-session probe transcript — dispatched worker self-report</summary>

```
$ git rev-parse HEAD          # in the repo holding this branch
f467a6134e2ff72c9658babf04388bf36d7635a0

$ opencode run "Use the task tool to dispatch a subagent with exactly this prompt: 'Inspect your own context/instructions. Does the text EXTREMELY_IMPORTANT or a using-superpowers bootstrap block appear in YOUR context? You are a bounded worker: do not invoke any skills, just answer YES or NO plus one sentence.' Then report the subagent's exact answer." --model "zhipuai-coding-plan/glm-5.2"

> orchestrator · glm-5.2
• Probe subagent context Explorer Agent
✓ Probe subagent context Explorer Agent
The subagent's exact answer:

> **NO** — Neither the string "EXTREMELY_IMPORTANT" nor any "using-superpowers" bootstrap block appears in my context; my instructions consist only of the Explorer role description (grep/glob/ast_grep navigation guidance, read-only constraints, and output format), environment details, and tool definitions.
```

</details>

## Evaluation

- Initial prompt: verify whether issue #2160 is real in the context of PR #2106's branch, and fix it if so.
- Sessions after the change: 3 on this branch (unit suite; live API end-to-end against real sessions; clean-session acceptance), plus a pre-fix investigation session establishing root cause from the binary's hook call site.
- Before/after: before, the transform hook injected the identical controller bootstrap into task children (code path has no session identity at all — the bug cannot be fixed at the prompt level); after, children are skipped deterministically (verified against a real child session), while top-level behavior is byte-for-byte unchanged (acceptance test still auto-triggers `brainstorming`). The failure mode from the issue — workers restarting brainstorming/design cycles for parent-authorized work — requires the bootstrap to be present, and it no longer is.

## Rigor

- [x] Not a skills change — no SKILL.md or skill content modified
- [x] This change was tested adversarially, not just on the happy path — unit suite covers every fail-open branch (missing client, missing sessionID, transient lookup error + recovery, idempotency), and the live test exercises the real SDK shapes against real sessions
- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) — `<SUBAGENT-STOP>` is deliberately retained as defense-in-depth for the fail-open paths and for harnesses without structural detection

## Human review

- [x] A human has reviewed the COMPLETE proposed diff before submission — @GoldJohnKing reviewed the full diff in-session, drove the design review (split decision, child-session semantics, `<SUBAGENT-STOP>` retention), and explicitly approved the push and PR creation.

### DIFF
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -68,23 +68,56 @@ let _bootstrapCache = undefined; // undefined = not yet loaded, null = file miss
 // agent step) pays only one client roundtrip per session.
 const _childSessionCache = new Map();
 
+// sessionIDs whose lookup failure has already been logged. Failed lookups are
+// deliberately not cached (so they can recover), which means the hook retries
+// every step; without this guard a persistently failing session would log an
+// error on every single step.
+const _lookupFailureLogged = new Set();
+
+const logLookupFailure = (sessionID, reason) => {
+  if (_lookupFailureLogged.has(sessionID)) return;
+  _lookupFailureLogged.add(sessionID);
+  console.error('[superpowers] session lookup failed, treating session as top-level:', reason);
+};
+
 const isChildSession = async (fetchSession, sessionID) => {
   if (!sessionID) return false; // unknown session: keep current behavior
   if (_childSessionCache.has(sessionID)) return _childSessionCache.get(sessionID);
 
-  let isChild = false;
+  let result;
   try {
-    const result = await fetchSession(sessionID);
-    // The V1 SDK returns { data: Session } rather than the record itself.
-    const session = result && typeof result === 'object' && result.data && typeof result.data === 'object' && !('parentID' in result)
-      ? result.data
-      : result;
-    isChild = Boolean(session && typeof session === 'object' && session.parentID);
+    result = await fetchSession(sessionID);
   } catch (err) {
     // Fail open: on lookup errors keep injecting (previous behavior) and do
     // not cache, so a transient failure can recover on the next step.
-    console.error('[superpowers] session lookup failed, treating session as top-level:', err);
+    logLookupFailure(sessionID, err);
     return false;
+  }
+
+  // The V1 SDK only throws on non-2xx when called with { throwOnError: true }
+  // (sdk error-interceptor.ts); without it an error resolves as
+  // { data: undefined, error, response } and never reaches the catch above.
+  // Treat that shape as a lookup failure too and never cache it — otherwise
+  // one transient error would pin a child session as top-level for its whole
+  // life and re-inject the controller bootstrap every step (#2160).
+  if (!result || (typeof result === 'object' && result.error)) {
+    logLookupFailure(sessionID, result && result.error ? result.error : result);
+    return false;
+  }
+
+  // The V1 SDK returns { data: Session }; V2's ctx returns the record itself.
+  // Prefer .data whenever it is a non-null object — Session records have no
+  // `data` field, so the shapes stay unambiguous even if the envelope happens
+  // to carry its own parentID key.
+  const session = result && typeof result === 'object' && result.data && typeof result.data === 'object'
+    ? result.data
+    : result;
+  const isChild = Boolean(session && typeof session === 'object' && session.parentID);
+
+  if (isChild) {
+    // One-time visibility: a missing bootstrap in a subagent session should
+    // be explainable from the logs instead of failing silently.
+    console.log('[superpowers] skipping controller bootstrap for task subagent session:', sessionID);
   }
   _childSessionCache.set(sessionID, isChild);
   return isChild;
@@ -175,7 +208,10 @@ ${toolMapping}
       // 1.18.x bundle: trigger(..., {}, {messages})), so take the sessionID
       // from the message record itself.
       if (client && await isChildSession(
-        (id) => client.session.get({ path: { id } }),
+        // throwOnError makes the SDK throw on non-2xx so HTTP errors reach
+        // the catch path; isChildSession's result.error check still covers
+        // versions/callers that ignore the flag.
+        (id) => client.session.get({ path: { id } }, { throwOnError: true }),
         firstUser.info.sessionID,
       )) return;
 
--- a/.opencode/plugins/superpowers.js
+++ b/.opencode/plugins/superpowers.js
@@ -52,6 +52,44 @@ const normalizePath = (p, homeDir) => {
 // every agent step.  See #1202 for the full analysis.
 let _bootstrapCache = undefined; // undefined = not yet loaded, null = file missing
 
+// --- Task-subagent (child session) detection --------------------------------
+//
+// #2160: the bootstrap drives controller workflows (brainstorming, planning,
+// approval cycles). Injecting it into task subagent sessions makes workers
+// restart design/approval cycles for work the parent already authorised; the
+// <SUBAGENT-STOP> note inside the bootstrap relies on model compliance, which
+// is not reliable. Detect child sessions structurally instead: OpenCode task
+// sessions are created with a parentID, so when the session carrying the
+// message has a parentID we skip bootstrap injection. Skills stay registered
+// for every session — workers keep explicit access to execution skills.
+
+// sessionID -> is-child decision. parentID never changes for a session, so
+// the result is cached for life and the injection hook (which fires on every
+// agent step) pays only one client roundtrip per session.
+const _childSessionCache = new Map();
+
+const isChildSession = async (fetchSession, sessionID) => {
+  if (!sessionID) return false; // unknown session: keep current behavior
+  if (_childSessionCache.has(sessionID)) return _childSessionCache.get(sessionID);
+
+  let isChild = false;
+  try {
+    const result = await fetchSession(sessionID);
+    // The V1 SDK returns { data: Session } rather than the record itself.
+    const session = result && typeof result === 'object' && result.data && typeof result.data === 'object' && !('parentID' in result)
+      ? result.data
+      : result;
+    isChild = Boolean(session && typeof session === 'object' && session.parentID);
+  } catch (err) {
+    // Fail open: on lookup errors keep injecting (previous behavior) and do
+    // not cache, so a transient failure can recover on the next step.
+    console.error('[superpowers] session lookup failed, treating session as top-level:', err);
+    return false;
+  }
+  _childSessionCache.set(sessionID, isChild);
+  return isChild;
+};
+
 export const SuperpowersPlugin = async ({ client, directory }) => {
   const homeDir = os.homedir();
   const superpowersSkillsDir = path.resolve(__dirname, '../../skills');
@@ -131,6 +169,15 @@ ${toolMapping}
       // This prevents double injection when OpenCode passes an already
       // transformed in-memory message array through the hook again.
       if (firstUser.parts.some(p => p.type === 'text' && p.text.includes('EXTREMELY_IMPORTANT'))) return;
+
+      // #2160: never restart the controller workflow inside task subagent
+      // (child) sessions. V1 passes no input to this hook (verified in the
+      // 1.18.x bundle: trigger(..., {}, {messages})), so take the sessionID
+      // from the message record itself.
+      if (client && await isChildSession(
+        (id) => client.session.get({ path: { id } }),
+        firstUser.info.sessionID,
+      )) return;
 
       const ref = firstUser.parts[0];
       firstUser.parts.unshift({ ...ref, type: 'text', text: bootstrap });
--- a/tests/opencode/run-tests.sh
+++ b/tests/opencode/run-tests.sh
@@ -45,6 +45,7 @@ while [[ $# -gt 0 ]]; do
             echo "Tests:"
             echo "  test-plugin-loading.sh  Verify plugin installation and structure"
             echo "  test-bootstrap-caching.sh  Verify bootstrap content caching"
+            echo "  test-child-session-gate.sh  Verify child session bootstrap gating (#2160)"
             echo "  test-tools.sh           Test use_skill and find_skills tools (integration)"
             echo "  test-priority.sh        Test skill priority resolution (integration)"
             exit 0
@@ -61,6 +62,7 @@ done
 tests=(
     "test-plugin-loading.sh"
     "test-bootstrap-caching.sh"
+    "test-child-session-gate.sh"
 )
 
 # Integration tests (require OpenCode)
--- a/tests/opencode/test-child-session-gate.mjs
+++ b/tests/opencode/test-child-session-gate.mjs
@@ -1,0 +1,267 @@
+import { pathToFileURL } from 'url';
+
+// Test: Child session gate (#2160)
+//
+// Drives the plugin's experimental.chat.messages.transform with a mock
+// client to verify the #2160 child-session gate:
+//   - child sessions (parentID set) never receive the controller bootstrap
+//   - top-level sessions and degenerate shapes still receive it
+//   - lookup failures fail open, are NOT cached, and recover next step
+//   - session.get is called with { throwOnError: true }
+//   - skip/error decisions are logged exactly once per session
+//
+// Usage: node test-child-session-gate.mjs PLUGIN_PATH CASE_NAME
+
+const [, , pluginPath, caseName] = process.argv;
+
+const cases = {
+  'skip-child-envelope': skipChildEnvelope,
+  'inject-top-level-envelope': injectTopLevelEnvelope,
+  'skip-child-bare-record': skipChildBareRecord,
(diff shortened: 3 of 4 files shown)


obra:
**Agentic review note — not a human review, not an approval.**

I'm Claude, an AI agent (Fable 5.1, Claude Code 2.1.259), posting from Jesse (@obra)'s account at his direction. Jesse asked me to review the open PRs that resolve open issues; the security pass, code review, and local test runs below were done by Opus subagents under my direction against `dev` at fd02874 (v6.3.0). **Jesse has not reviewed this PR yet.** Nothing here merges, approves, or requests changes on his behalf; it is information for him and for the author.

**Security pass:** SAFE. One read-only session lookup via the client OpenCode already hands the plugin; result used only as Boolean(parentID). Robustness notes: silent skip path, persistent lookup failure re-attempts every step, cache never evicted, envelope detection tests absence of parentID.
**Code review verdict:** approve-with-notes.
**Local tests:** PASS. Isolated clone on top of gh/dev @ fd02874, PR head f467a61. node --check exit 0; tests/shell-lint exit 0 (all PASS) on dev and PR; tests/opencode/run-tests.sh exit 0 (2 passed) on dev and PR; opencode binary not installed so the two --integration tests did not run. A throwaway harness importing the plugin and stubbing client.session.get: {data:{parentID:'x'}} → bootstrap skipped; {data:{}} → injected; bare {parentID:'x'} → skipped; rejected promise → injected (fail-open) with the error logged; cache: repeat of a skipped session made 0 further calls, repeat of an errored session made 1 new call. Five degenerate shapes ({data:null}, undefined, {data:{parentID:null}}, {data:{parentID:''}}, missing sessionID) all injected, correct.

**Code review findings, most important first:**
1. Real gap that contradicts the PR body: client.session.get is called without throwOnError, and OpenCode's SDK only throws with that flag (packages/sdk/js/src/error-interceptor.ts); otherwise a 404/500 resolves as {data: undefined, error, response}. That never reaches the catch, so result.data is falsy, session = result, parentID absent, isChild = false, and it is WRITTEN TO THE CACHE. One transient non-2xx for a child session pins it as top-level for its whole life and re-injects the controller bootstrap every step, the exact #2160 bug, silently. Simulated with six SDK shapes. One-line fix: pass { throwOnError: true } or treat result.error / missing data as an uncached failure.
2. No test lands in-repo: the body's '8/8 mock-client unit suite' was run out of tree. tests/opencode/test-bootstrap-caching.mjs already imports the plugin and drives the transform with a mock; it passes client: {} with no sessionID, so it exercises only the short-circuit and would still pass if the child gate were deleted. Add skip-on-parentID, inject-on-no-parentID, and no-cache-on-failure cases there.
3. Diagnosability: the skip path logs nothing, so a missing bootstrap is invisible; a persistent throw logs on every step.
4. Envelope heuristic !('parentID' in result) is convoluted but correct against both real shapes; Session has no data field, so the ambiguity flagged in the security pass cannot fire in practice.

**Test-run findings:**
- No tests added: the only changed file is .opencode/plugins/superpowers.js, and nothing in tests/opencode exercises experimental.chat.messages.transform, so the new isChildSession logic (lines 71-91) has no automated coverage.
- Heuristic hole at line 79: envelope detection is !('parentID' in result); a V1 envelope carrying its own parentID key, even undefined, makes the code pick the envelope and miss data.parentID, so a child session gets the bootstrap. Verified with shape {parentID: undefined, data:{parentID:'x'}} → injected. Contrived, but it is the one shape that silently defeats the fix.
- Error path logs a full stack on every agent step for a session whose lookup keeps failing (no caching on error, by design). Correct fail-open behavior, noisy in practice.

**Against the PR rules in CLAUDE.md:** dev base PASS; identity table PASS (GLM-5.3 / OpenCode 1.18.25 / plugins and MCP listed / author reviewed the diff in-session); existing PRs PASS (#2106, #2210, #2188, #2004 analyzed; #2106 contains this commit); one problem PASS; no dependency; skill text untouched, SUBAGENT-STOP and human-partner language preserved; acceptance transcript PASS (brainstorming auto-triggers); tests FAIL (none in-repo). Upstream verified at sst/opencode v1.18.25: the transform runs for child sessions too (prompt.ts:1255), task.ts:159 sets parentID on subagent sessions, session.get returns Session with optional parentID.

---
*Findings can be wrong. If something above misreads your change, reply here — Jesse reads these.*

— Claude Fable 5.1, Claude Code 2.1.259, session https://claude.ai/code/session_01UiEfXTZAC5cuH4hgx24mbB

GoldJohnKing:
**Response to the agentic review note** — all four findings were verified against the code and the v1.18.25 SDK source, confirmed real, and fixed in 3a560f1 (TDD: each fix has a test that failed on f467a61 first).

**Finding 1 — result-tuple errors bypass the catch and poison the cache.** Confirmed: `wrapClientError` (`packages/sdk/js/src/error-interceptor.ts`) only fires with `throwOnError: true`, so a non-2xx resolved as `{data: undefined, error, response}`, never reached the catch, and was cached for the session's whole life. Fixed on both paths: the call site now passes `{ throwOnError: true }`, and `isChildSession` additionally treats `!result || result.error` as an **uncached** lookup failure (covers callers/versions that ignore the flag). Regression test `error-tuple-fail-open-uncached`: on f467a61 it observes 1 lookup across 3 steps (poisoned); after the fix, one retry per step with fail-open injection.

**Finding 2 — no in-repo tests.** Added `tests/opencode/test-child-session-gate.mjs` + `test-child-session-gate.sh`, wired into `run-tests.sh`. Eight cases: skip-on-parentID (V1 envelope, bare V2 record, and the envelope-with-parentID-key hole), inject-on-no-parentID, no-cache-on-failure (resolved error tuple and thrown error), transient-failure recovery, and the degenerate shapes `{data: null}`, `{data: {parentID: null}}`, `{data: {parentID: ''}}`, `undefined`. Deleting the child gate now fails the suite.

**Finding 3 — diagnosability.** The skip decision logs exactly once per child session (`skipping controller bootstrap for task subagent session: <id>`); lookup failures are deduplicated per session (first failure only) while still retrying uncached — the per-step full-stack spam is gone. Both log counts are pinned by tests.

**Finding 4 — envelope heuristic hole.** Replaced `!('parentID' in result)` with: prefer `.data` whenever it is a non-null object. Session records carry no `data` field, so `{parentID: undefined, data: {parentID: 'x'}}` now resolves as a child. Test `skip-child-envelope-parentid-key` injects on f467a61 and skips after the fix.

**Test results**

- `tests/opencode/run-tests.sh` — 3/3 PASS (plugin-loading, bootstrap-caching, child-session-gate)
- `tests/opencode/run-tests.sh --integration` — test-tools PASS; test-priority PASS with `OPENCODE_TEST_TIMEOUT_SECONDS=300` (its first attempt hit the default 120s timeout on a slow model response — environment latency, not plugin-related)
- `node --check` PASS; `shellcheck` not on PATH locally

**End-to-end run** — real `opencode` 1.18.27 binary, model `zhipuai-coding-plan/glm-5.3`, isolated `$OPENCODE_CONFIG_DIR` containing only this PR's plugin + skills, isolated `$HOME`. Note the bootstrap transform is applied at prompt-assembly time and is not persisted, so the structural check below uses the session table (`parent_id`) plus the plugin's new one-time decision log.

### Phase A — top-level session (bootstrap must be injected)
```
$ opencode run --model zhipuai-coding-plan/glm-5.3 --print-logs "Let's make a react todo list"

> build · glm-5.3
Using superpowers:brainstorming to explore requirements before building.
→ Skill "brainstorming"
$ ls -la /tmp/opencode/e2e-2160/work && ls /tmp/opencode/e2e-2160/work/node_modules 2>/dev/null | head -5
total 0
drwxr-xr-x 2 gjk gjk  40 Sep  4 09:11 .
drwxr-xr-x 4 gjk gjk 100 Sep  4 09:11 ..
# Todos
[✓] Explore project context
[•] Ask clarifying questions (one at a time)
[ ] Propose 2-3 approaches with trade-offs
[ ] Present design sections, get approval
[ ] Write design doc to docs/superpowers/specs/
[ ] Spec self-review
[ ] User reviews written spec
[ ] Invoke writing-plans skill
The workspace is empty — this is a **new project**, so I'm on the **architectural path** (full design process, even if the app is small).
First question — what's this todo list **for**?
- **A) Learning project** — keep it simple, idiomatic React, no frills
- **B) Something you'll actually use daily** — persistence and polish matter
- **C) Portfolio/demo piece** — visual design and extras (filters, animations) showcased
- **D) Base for something bigger** — you plan to grow it into a task manager later
```

### Phase B — task subagent session (bootstrap must be skipped)
```
$ opencode run --model zhipuai-coding-plan/glm-5.3 --print-logs "Use the task tool with subagentType general and prompt exactly: 'Reply with the single word READY'. After the subagent finishes, output its reply verbatim."

> build · glm-5.3
• Test subagent reply General Agent
[superpowers] skipping controller bootstrap for task subagent session: ses_f9605f2f9ffehncWyqqLW0P9RZ
✓ Test subagent reply General Agent
READY
```

### Persisted session structure (sqlite, opencode.db)
```
ses_f9606a2b4ffeDkBSOV9kwW6DiR | (null) | Building a React todo list
ses_f960607b3ffe2mzYc2mVuRstWl | (null) | Subagent task tool invocation
ses_f9605f2f9ffehncWyqqLW0P9RZ | ses_f960607b3ffe2mzYc2mVuRstWl | Test subagent reply (@general subagent)
```


The single `[superpowers]` line across both runs is the skip log for the child session — no lookup-failure logs, which also confirms `session.get` with `{ throwOnError: true }` succeeds against the real SDK client. Phase A auto-triggers `brainstorming` on the acceptance prompt (bootstrap active in the controller session); the Phase B child session (`parent_id` set) is skipped and replies `READY` with no controller-workflow restart.

**Identity:** this follow-up was authored by GLM-5.3 (`zhipuai-coding-plan/glm-5.3`) in OpenCode 1.18.27, with the repo's own plugin and oh-my-opencode-slim present in the interactive harness; the e2e runs above used a clean isolated config with only this PR's plugin. The author reviewed the complete diff and the transcript before pushing.

PR #2236, opened 2026-08-31
feat: add diagnosing-superpowers skill — evidence-based session diagnosis, scrubbed bug-report bundles

obra:
> This PR targets `dev`.

| Field | Value |
|-------|-------|
| Your model + version | Claude Fable 5 (`claude-fable-5`) as controller; implementer and reviewer subagents on Fable 5, Opus, Sonnet, and Haiku (per-task model recorded in the SDD ledger). Review-response commits (b5c738f through 801badb): Claude Fable 5.1 (`claude-fable-5-1`), micro-test subagents on Sonnet |
| Harness + version | Claude Code 2.1.247–2.1.248 (CLI, macOS 26.6.2); review-response commits from Claude Code CLI on Linux |
| All plugins installed | superpowers 6.3.0, agent-sdk-dev, claude-code-setup, code-simplifier, context7, frontend-design, gopls-lsp, linear, mcp-server-dev, plugin-dev, proving-it-works, release-radar, rust-analyzer-lsp, swift-lsp |
| Human partner who reviewed this diff | Jesse Vincent (maintainer) |

When a superpowers session goes wrong, there is no structured way to find out what happened. The complaints that motivated this are the ones the maintainer actually gets: "it took too long", "why did it do this extra work?", "why is it so expensive?", "what the hell is it doing?", "the skill never fired", "it ignored the plan after compaction". Today the answer is either a hand-rolled `jq` session over a multi-megabyte JSONL transcript (single lines exceed a megabyte and will blow a session's context if grepped), or a guess. Bug reports to this repo arrive without the model, harness, versions, session ids, or evidence the maintainers need, because nothing helps the reporter produce them.

## What does this PR change?

Adds one core skill, `diagnosing-superpowers`: intake with the human partner to pin the problem down, locate the session transcript(s) on disk (Claude Code and Codex formats verified against real files; a discovery procedure for other harnesses), dispatch seven analyst subagents (skill timeline, plan adherence, repeated work, stumbles, quality evidence, request conflicts, cost and time), write a report with `path:line` evidence and an explicit "superpowers involvement: not indicated / possible / likely" line, and on request export a scrubbed bundle, search or draft a GitHub issue (never posting without approval), and search for similar local sessions. Plus a structure test, a README section, a docs line, a `diagnosis_report.md` issue template, and the `automated-issue-report` label.

**Changes since the first review** (all on this branch):
- The scrubbed bundle is built only when the partner asks. If intake named a bug report as the goal, the skill says once that a bundle is available on request and waits. On handover it states what the bundle contains, points at the scrub log, and says scrubbing can miss things so every file needs review before sharing.
- The seven analyst prompts share one preamble (`prompts/analyst-common.md`) and the context-safety rules live once in `references/context-safety.md`.
- GitHub mechanics live in `references/github-issues.md`. `gh` is used for search and, after approval of the exact text, creation, with curl and a prefilled template link as fallbacks. GitHub silently drops labels from reporters without push access, so the template footer is the durable marker of a skill-filed issue.
- "Commitment" became "plan step" throughout; in a transcript full of git commits the old word misread.
- A writing pass against `elements-of-style` and `writing-skills`: quick-reference table now says all seven analysts always run, issue template gained title and bundle slots, several prose fixes.

Yes. It is harness-agnostic (the skill names actions; harness specifics live in `references/`), has zero dependencies (`jq` with a `python3` fallback, both optional), and applies to any project: it diagnoses the agent's session, not the user's code. It deliberately does not diagnose superpowers itself; the skill reports evidence and hands off to the issue tracker, so it is a reporting tool for maintainers rather than an opinion about the plugin.

- **Shipped scripts (Node or Python) for locate/normalize/scrub/zip.** Rejected for v1 by the maintainer: pure prose first, tooling only once the shape is proven. The cost of that choice is visible in the eval (analysis is expensive, ~460k subagent tokens per scrub-audit pass on an 884 KB bundle).
- **One big SKILL.md.** Rejected: the analyst prompts alone are several hundred lines; one file per subagent job keeps the main context small.
- **Field-level references for every supported harness.** Rejected: only formats verified against real files on disk get field claims (Claude Code 2.1.233–2.1.248, Codex CLI 0.146–0.149); everything else gets a discovery procedure. No invented formats.
- **Letting the skill propose fixes to superpowers.** Rejected: that is the triager's job; the skill would otherwise produce exactly the speculative "my agent flagged this" reports this repo rejects.

No. Spec, plan, skill, its test, and the two one-line index entries are all this one skill.

## Existing PRs
- Related PRs: #1722 (closed; reduces context bloat in a live session, not diagnosis), #863/#864 (closed; retrospective skill that persists lessons learned), #1008 (open; post-mortem of production failures using a 4M framework). None reads the session transcript with evidence or builds a maintainer-facing bug-report bundle, and none was closed for reasons that apply here (this PR is spec-driven, eval-backed, and grounded in real local sessions).

## Environment tested

| Claude Code (CLI, macOS 26.6.2) | 2.1.247–2.1.248 | Claude Fable 5 / Opus / Sonnet / Haiku | claude-fable-5 (controller and runners); opus, sonnet, haiku (implementers, reviewers, some runners) |

Session stores read during the eval: Claude Code (2.1.233–2.1.248 transcripts) and Codex CLI (0.146.0, 0.147.0, 0.149.0-alpha.4.1 rollouts).

Not applicable; no new harness.

## Evaluation

- **Initial prompt:** "We need to add skill to superpowers for debugging superpowers sessions. it should work cross harness and be able to debug both the current superpowers session and previous sessions…" (full requirements in the spec).
- **Eval sessions after the change:** 46 scored scenario runs across five SKILL.md versions, 20 micro-test reps (two prohibitions × control/skill × 5), and one end-to-end run; every run is a fresh subagent on a real local transcript.
- **Before/after:** Baseline without the skill: intake skipped on the vague complaints (analyzed first, asked nothing), a skill-file edit prescribed from one transcript on the "just tell me which skill is broken" push, unsourced dollar figures stated as fact, rejected retrieval candidates not enumerated. With the final skill: all twelve scenarios clean against one version; micro-tests control 5/5 → skill 0/5 for intake-first and 5/5 → 0/5 for no-superpowers-diagnosis. Four refactor rounds on the intake rule (it first over-blocked scoped requests, then under-blocked a whole-session cost complaint, then over-blocked an already-scoped dispatch request); the final wording is the only one measured clean in both directions.
- **What the eval did not settle:** three runs of the same fixture with the same pre-answered problem statement produced three materially different substantive answers about what a compaction caused. All three passed, because the criteria test completeness and citation, not agreement. The skill reliably produces a well-formed, evidenced report; it does not reliably produce the same one twice.
- **Review-response micro-tests (Sonnet, single-shot, 5 reps per arm, every response read by hand):** bundle-unprompted scenario, skill 0/5 and control 0/5 (removing the old "or the intake goal was a bug report" trigger was the fix; the added sentence is not load-bearing). `gh` scenario against the interim no-`gh` wording, skill 0/5 and control 4/5; that wording was then superseded by the decision to use `gh` with the pre-existing approval gate, which the original eval covered ("file it now" with real GitHub write scope present).
- **Not exercised:** triggering through the Skill tool. Every eval run read `SKILL.md` by path because the installed plugin is 6.3.0. A clean-install trigger check is needed before release.

Full method, per-version table, micro-test table, refactor rounds, and Red Flags provenance are in the maintainer's eval record, kept outside the repo because even the scrubbed form describes private sessions; available from the maintainer on request.

## Rigor

- [x] If this is a skills change: I used `superpowers:writing-skills` and completed adversarial pressure testing (baseline table, per-version table, and micro-tests with per-rep verdicts in the maintainer-held eval record)
- [x] This change was tested adversarially, not just on the happy path (scenarios: time pressure, authority push for a fix, 1.3 MB transcript line, "just zip it up", "file it now" with real GitHub write scope present, subagent misdirection)
- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) without extensive evals showing the change is an improvement (this PR adds a new skill; existing skills are untouched)

## Human review

https://claude.ai/code/session_01DyaGKhTXvHNs2JgPhDktz7


skills/diagnosing-superpowers/prompts/cost-and-time.md
```diff
@@ -0,0 +1,65 @@
...
+  evidence: <absolute path>:<line> — "<quote, at most 200 characters>"
+  turns: <first human turn>–<last human turn>
+  confidence: high | medium | low
+
+Checked: <what you examined: files, line ranges, commands used>
+```
+
+A finding without a `path:line` will be discarded by the dispatcher, so do
+not write one. If you found nothing, return `- none found` and the Checked
+line.
```
arittr: this first block is very similar - or the same - across multiple sub prompts. any shared direction should probably be moved up to a higher-level prompt/loc, and we need keys for the output schema to be in a consistent order, etc

skills/diagnosing-superpowers/SKILL.md
```diff
@@ -0,0 +1,117 @@
...
+   case file path and one file from `prompts/`: `skill-timeline.md`,
+   `plan-adherence.md`, `repeated-work.md`, `stumbles.md`,
+   `quality-evidence.md`, `request-conflicts.md`, `cost-and-time.md`.
+   Split a dimension by turn range when the transcript is long. Discard
+   any returned finding without `path:line`.
+4. **Report.** Fill every section of `templates/report.md` in order, write
+   it to the workspace, show it, and give the path.
+5. **GitHub issues** — when report §7 says possible or likely, or your
+   partner asks. Search open and closed issues on `obra/superpowers` for
+   the symptoms (`gh` if installed, else the public search API with curl,
```
arittr: let's not use `gh`, as it could easily have priv'd access. use the public api

skills/diagnosing-superpowers/SKILL.md
```diff
@@ -0,0 +1,117 @@
...
+| "Why did it do this extra work?" | repeated-work, plan-adherence |
+| "Why is it so expensive?" | cost-and-time |
+| "What the hell is it doing?" (still running) | skill-timeline; note in-progress in coverage |
+| "It ignored the plan" | plan-adherence, compaction lines first |
+| "Skill X never fired" | skill-timeline |
+
+## Hard rules
+
+- **Context safety.** One transcript line can be a megabyte. Check
+  `wc -lc` and long lines first. Never `cat` or `grep` for content: line
```
arittr: the `wc -lc` thing def needs to be moved up in scope, idk where. it appears a lot. maybe in the `./resources` for the skill

arittr (changes requested):
- there's a lot of repetition in the sub prompts - in some cases this might be right, but we should audit
- `references/codex-sessions.md` and `references/claude-code-sessions.md` are extremely coupled to current json structure... `other-harnesses.md` is a lot less brittle, maybe we can do auto-discovery for all harnesses?
- we need a very clear README block explaining what's going on
- some other in line comments

skills/diagnosing-superpowers/prompts/cost-and-time.md
obra: Done in bcfe1e2. The identical 39-line preamble (role, inputs, context safety, return format) now lives once in prompts/analyst-common.md; each dimension prompt opens with a two-line pointer to it and the dispatcher passes both files.

skills/diagnosing-superpowers/SKILL.md
obra: Done in c13dc98. The skill no longer uses gh anywhere. Search goes through the unauthenticated public search API with curl. Filing hands the partner a prefilled new-issue link on a new diagnosis_report.md issue template, which applies the bug and automated-issue-report labels for any reporter (the labels= URL param only works with triage rights). The partner submits the issue themselves and attaches the bundle in the form; the skill never posts.

skills/diagnosing-superpowers/SKILL.md
obra: Done in bcfe1e2. The rule now lives once in references/context-safety.md. SKILL.md, the case template, the three harness references, similar-session, and the shared analyst preamble all point at it; the only wc -lc left in the skill is in that file.

skills/diagnosing-superpowers/SKILL.md
obra: Update: after discussion, Jesse decided the skill should use gh for both search and creation, so 801badb restores that. Reasoning: reading public issues carries no privilege risk regardless of token scope, and posting was already gated on the partner approving the exact issue text, so gh adds capability without adding exposure. The curl and prefilled-link paths stay as fallbacks for machines without gh, and the new automated-issue-report label plus diagnosis_report.md template remain. One caveat noted in references/github-issues.md: GitHub silently drops labels from reporters without push access, so the template footer is the durable marker of a skill-filed issue.

arittr:
Following up on my earlier comment about `references/claude-code-sessions.md` and `references/codex-sessions.md` being coupled to the current JSON shapes: I ran a small eval to see whether the skill actually needs them.

**Setup.** One "find the session I'm thinking of" task per harness, against a fresh home containing three native session files (one target, two near-duplicates). Two arms: the PR head as-is (`801badb`), and a treatment that deletes all three `references/*-sessions.md` files and replaces them with one shared discovery procedure that tells the agent what to establish (identity, record meanings, absolute paths, uncertainty) but not where anything lives. Claude Code 2.1.209 / Opus 4.8, Codex CLI 0.146.0 / gpt-5.6-sol, Pi 0.80.7 / gpt-5.6-sol. One run per cell.

**Result.**

| Harness | With per-harness references | Shared procedure only |
|---|---|---|
| Codex | found session, absolute path, request, tool call + result | same |
| Claude Code | found session, request, tool evidence; reported a `~` path | same; after one added sentence requiring full absolute paths, clean pass |
| Pi | found session, request, tool evidence | same |

So for locating a session and pulling out the basic facts, none of the three harnesses needed the reference files. The one thing the shared procedure had to say explicitly was "report the full absolute path"; Claude otherwise abbreviates.

**Proposed change** (diff: https://github.com/obra/superpowers/compare/diagnosing-superpowers...arittr/pr2236-shared-discovery): delete the three reference files, add `references/session-discovery.md`, and have the analyst prompts read record meanings from the case file instead of hard-coding field names. Net −285 / +126 lines.

**Caveats, which are real.** This is n=1 per cell on small, clean histories. It tested *locating* a session, not *interpreting* one. Most of what the deleted files contain is field semantics the seven analysts rely on: which `user` lines are injected rather than typed, Codex `token_count` being cumulative, subagent linkage, compaction markers. I did not test whether an agent recovers those from the file alone, and a wrong guess there produces a confident wrong report rather than an error. Two ways to handle that: keep the reference files as verified hints beneath the shared procedure (they already say "trust the file over this table"), or run a second probe that asks for turn counts, token totals, subagent cost, and compaction timing before deleting them. Happy to do the second if you'd rather delete.

Full records with hashes and costs (about $5 total): [six-run pilot](https://github.com/prime-radiant-inc/superpowers-evals/blob/main/docs/experiments/2026-09-09-pr2236-session-discovery.md), [Pi follow-up](https://github.com/prime-radiant-inc/superpowers-evals/blob/main/docs/experiments/2026-09-09-pr2236-pi-followup.md), [Claude absolute-path follow-up](https://github.com/prime-radiant-inc/superpowers-evals/blob/main/docs/experiments/2026-09-09-pr2236-claude-absolute-path.md), [design](https://github.com/prime-radiant-inc/superpowers-evals/blob/main/docs/superpowers/specs/2026-09-09-pr2236-session-discovery-design.md).

arittr:
Eval disposition after revisiting the intended use with Drew: **accept shared discovery and the tested evidence-preservation follow-up as a useful maintainer debug kit.** The purpose is to give a maintainer scrubbed evidence they can investigate, with important uncertainties visible.

The completed post-report test demonstrated that:

- A fresh recipient, given only the scrubbed bundle, reconstructed the workflow and verified all four source findings and both related-session classifications. The controller separately compared that account with the withheld key.
- The recipient distinguished static evidence from demonstrated runtime behavior and Superpowers causation.
- Known sensitive markers were absent, scrub counts matched, and raw native JSONL and opaque encrypted payload values were excluded. These are scoped privacy checks, not an exhaustive guarantee.
- The archive and local issue draft were delivered, archive contents were verified, and original evidence remained unchanged. The subject did not publish an issue, comment, or upload.

Remaining inaccuracies include confusing historical-path wording, incorrect provenance pointers, stale preparation/status text, and a four-versus-five inspection count. They deserve follow-up, but did not prevent the recipient from investigating the incident or recognizing the unknown historical version. We no longer consider those inconsistencies independently release-blocking for this debug-kit purpose.

The earlier strict reconciliation criterion still failed, and the shorter incomplete attempt remains recorded. This is a revised release judgment, not a rewrite of those results. The completed export test is one sample; it does not establish universal report accuracy. No further run is needed solely to eliminate these bookkeeping discrepancies.

**Source status:** this export result applies to the prepared follow-up candidate `950b4b3732ff49cf139ba47cff505eb866717e71`, which has not been pushed. The PR is currently at `c80ef0538d0b841edba3f4a458671c2e917029c7`. The recommendation is to land the tested follow-up; this comment does not imply it is already included in the PR.

arittr (approved):
LGTM

PR #2246, opened 2026-09-02
docs(readme): link rubyonai.com as the harness write-up

marostr:
Opened against the wrong remote by mistake and closed. Ignore.


marostr:
Opened against the wrong remote by mistake — this belongs on fryga-io/superpowers-rails. Sorry for the noise.

PR #2263, opened 2026-09-05
fix(brainstorming): a request for context does not close the open question

### LINKED ISSUE #1989: brainstorming: keep the current question open when the user asks for context instead of answering

The `brainstorming` skill says to ask one question at a time, but it does not explicitly say what to do when the user cannot answer yet and instead asks for more context about the current question.

In practice, this can produce a confusing flow:

1. The agent asks question A.
2. The user asks for more context, implications, or trade-offs about question A.
3. The agent provides that context.
4. The agent then immediately asks question B.

From the user's point of view, question A is still open. The user did not answer it yet; they only asked for help understanding it. Advancing anyway breaks the one-question-at-a-time intent and makes the brainstorming flow feel like a checklist instead of a guided decision process.

This matters most during design work, where the current question often needs clarification before the user can make a choice. If the skill moves on too early, it weakens the validation loop and creates avoidable confusion.

## Proposed solution

Add an explicit rule to `skills/brainstorming/SKILL.md` saying that a request for more context does not count as an answer to the current question.

Proposed wording:

> **Do not advance past an unanswered question.** If the user asks for more context, implications, examples, or trade-offs about the current question, treat that as clarification on the same question rather than progress to the next one. Answer only the clarification request, then restate and re-ask the same decision in clearer words. Do not introduce a new design question until the user has explicitly answered the current one.

Optional stronger wording:

> A request for more context is not an answer. It does not satisfy the one-question-at-a-time rule. The current question remains open until the user chooses an option or gives an explicit decision.

Expected behavior after this change:

- The agent asks question A.
- The user asks for context.
- The agent gives context only.
- The agent reformulates and re-asks question A.
- The agent waits.

1. Leave the current prompt as-is.

   This keeps the skill shorter, but it leaves too much ambiguity in a common interaction pattern. "One question at a time" is not enough by itself, because it does not define whether a clarification request closes the current question.

2. Treat this as a bug in a specific harness or model instead of a skill issue.

   I do not think that is the best framing. The problematic behavior follows naturally from a missing instruction in the skill itself, so the most direct fix is to tighten the prompt.

3. Add the guardrail to platform-adaptation docs instead of `brainstorming/SKILL.md`.

   That may help for some harness-specific question tools, but the core issue is flow control inside brainstorming. The primary instruction belongs in the brainstorming skill.

Yes.

This is not tied to a specific project domain, repo layout, or third-party tool. It improves the default quality of brainstorming conversations in any context where the user may need clarification before deciding. The rule strengthens the existing one-question-at-a-time behavior rather than introducing a new workflow specific to one platform.

## Environment (required)

| Field | Value |
|-------|-------|
| Superpowers version | 6.1.1 cached curated copy observed in Codex session |
| Harness (Claude Code, Cursor, etc.) | Codex |
| Harness version | Not captured in the session |
| Your model + version | GPT-5.4 Medium |
| All plugins installed | Not fully enumerated in the session |

## Context

This came up in a real brainstorming session while designing phase 1 of a roadmap in a live repository. The flow was:

- agent asked a design decision question;
- user asked for more context before answering;
- agent gave the context;
- agent then moved on to a new question.

That made the conversation harder to follow because the original decision was still unresolved.

Related issues reviewed before filing:

- #849 `Brainstorming skill asks generic questions despite rich project context`
  - Related because it is also about weak question selection in brainstorming.
  - Not a duplicate because it focuses on asking questions whose answers already exist in context, not on advancing after a context-only reply.

- #1773 `brainstorming: v6.0.0 platform-neutral bootstrap reframe makes Claude Code use AskUserQuestion`
  - Related because it discusses how brainstorming questions are delivered and preserves one-at-a-time conversational flow.
  - Not a duplicate because it is about tool routing and presentation, not about whether the current question remains open.

- #1642 `Multi-agent parallel code review + adaptive batched questioning in brainstorming`
  - Related because it discusses questioning strategy in brainstorming.
  - Not a duplicate because it is about batching questions for simple tasks, not about clarification on the current question.

- #388 `Brainstorm skill: Use the AskUserQuestion tool`
  - Related only at the level of brainstorming question UX.
  - Not a duplicate because it is about using a specific question tool, not question-state handling.

- #114 `Use the native AskUserQuestion tool to ask questions in the UI`
  - Related only at the level of question delivery.
  - Not a duplicate because it does not address clarification versus answer state.

- #530 `Feature Request: brainstorming skill should proactively challenge product assumptions`
  - Related because it proposes a refinement to brainstorming behavior.
  - Not a duplicate because it is about adding assumption-challenging behavior, not preventing premature advance.

Yash-Chindam:
> Targets `dev`.

Fixes #1989. Eval data and full method posted as a comment on the issue: https://github.com/obra/superpowers/issues/1989#issuecomment-5552534484

| Field | Value |
|-------|-------|
| Your model + version | Claude Opus 5 (`claude-opus-5`) |
| Harness + version | Claude Code 2.1.260 (desktop app), Windows 11 |
| All plugins installed | `superpowers@claude-plugins-official` v6.3.0 (`b36e0829c6d0`) — the only plugin installed |
| Human partner who reviewed this diff | @Yash-Chindam |

Coding agents under test for the evals: Codex CLI 0.153.4 (ChatGPT auth) and Claude Code 2.1.260 / Opus 5.

Reported in #1989 by @iskandersierra, hit in a real Codex session designing phase 1 of a roadmap: the agent asks a design question, the user asks for context instead of answering, the agent supplies the context and then moves to a *different* question. The original decision is still open, but the conversation has advanced past it.

The questioning rules say to ask one question at a time and "Only one question per message", but nothing defines what *closes* a question. A request for clarification is not an answer, and the skill never says so.

I measured how often this actually happens, because "an agent might do this" is not a problem statement. On Codex with the question tool available, the stock skill held the open question in **2 of 15** sessions. The other 13 supplied the trade-offs and then pitched forward — "If you want, I can turn that into a concrete design next" — with the original decision never put back to the user.

## What does this PR change?

Adds one bullet to `skills/brainstorming/SKILL.md`, immediately after the existing `Only one question per message` line, stating that a request for context is not an answer: supply the context, re-ask the same question, and keep it the only open question. One line, one file, no existing text modified.

Yes. It is not tied to a project, domain, repo layout, or third-party tool — it tightens flow control inside the core brainstorming loop, which every user of the skill goes through. It adds no dependency. It extends the existing one-question-at-a-time rule rather than introducing a new workflow, and it sits inline with the bullets it qualifies.

Per the issue's own analysis, the alternative homes were considered and rejected: platform-adaptation docs would scope it to particular harnesses, when the ambiguity is in the brainstorming flow itself.

1. **Leave it as-is.** Rejected on measurement: 13/15 stock Codex sessions advanced past the open question. That is the default behavior, not an edge case.

2. **Treat it purely as a harness bug and close the issue.** This one has real merit and I want to flag it rather than bury it. Codex ships `default_mode_request_user_input` set to `false`; with it off, Codex has no tool to ask a question mid-task, emits `request_user_input is unavailable in Default mode`, and substitutes "If you want, I can…". In that configuration the bullet is worth ~nothing (1/5 vs 0/5) because no wording can make an agent use a tool it does not have. **If maintainers judge that the harness flag is the real fix, this PR should be closed** — I would rather say that here than argue past it. The case for landing the bullet anyway is that it still helps once the tool exists, and costs one line.

3. **Stronger wording from the issue** ("A request for more context is not an answer. It does not satisfy the one-question-at-a-time rule."). I used the shorter form to match the surrounding bullet style and token-efficiency guidance in `writing-skills`. I did not eval the two wordings against each other, so I have no evidence one is better — happy to swap if you prefer the issue's phrasing.

4. **Add it to the Red Flags table or as a standalone section.** Rejected: the failure is a flow-control slip inside the questioning loop, and per "Match the Form to the Failure" the guidance belongs where the rule it qualifies already lives.

No. One bullet, one file.

## Existing PRs
- Related PRs: **none open.** #2213 previously proposed nearly identical wording, but it and its author's account now return 404 — the account was removed and GitHub took the PR with it. I verified via `gh api repos/obra/superpowers/pulls/2213` (404) and `gh api users/loulanyue` (404), and no PR cross-references #1989 in its timeline. That PR also targeted `main` and skipped this template; this one targets `dev`, completes the template, and carries the eval evidence that was asked for and never supplied.

## Environment tested

|---|---|---|---|
| Codex CLI | 0.153.4 | GPT-5.x (ChatGPT auth, default) | default for the account |
| Claude Code (desktop app) | 2.1.260 | Claude Opus 5 | `claude-opus-5` |

## Evaluation

**Initial prompt.** My human partner sent me issue #1989 and asked whether they could work on it, then asked me to run the evals the maintainer said would move it.

**Method.** Two arms per harness — stock `dev` @ `fd02874` vs patched — installed as separate plugin roots (`--plugin-dir` for Claude Code; `codex plugin marketplace add <local path>` for Codex), verified before every batch by grepping the installed `SKILL.md` for the bullet (expect 0 / expect 1). Fresh git fixture per run: the empty tasks page from `create_cost_checkbox_page`. Two user turns — the notifications request, then a context request that answers nothing and states no preference. Graded on the load-bearing criterion: was the agent's next request for input the same decision (restated, reworded, or with options), or a new question / a move to design. Offers to advance graded as fails.

**Sessions run after the change: 40 in total** (10 Claude Code, 30 Codex).

| Configuration | Stock | Patched |
|---|---|---|
| Claude Code / Opus 5 | 4/5 held | 4/5 held |
| Codex, `request_user_input` off (default) | 0/5 | 1/5 |
| Codex, `request_user_input` on | **2/15** | **7/15** |

**How outcomes changed.** On Codex with the question tool available, holding the open question went from 13% to 47%. Fisher's exact: p = 0.109 as graded, 0.050 excluding one broken run, 0.215 under the strictest grading. **That is suggestive, not conclusive, and I am not claiming otherwise.** The direction held across two separately-run batches, and an early small-sample reading of 3/5 regressed to 7/15 once I added reps — consistent with a real but modest effect rather than noise, though not proof of one.

On Claude Code the bullet changed nothing (4/5 both arms); the baseline there already holds the question most of the time.

**Representative transcripts.** Stock fail: *"My recommendation: start with in-app notifications… If you want, I can turn that into a concrete v1 design next."* Patched pass: *"Which should I design for: `in-app only`, `in-app + email`, or `in-app + email + push`?"*

**Limitations.** Non-interactive `codex exec`, not the interactive TUI the reporter used. The flag is off by default, so the 15-per-arm rows describe a non-default Codex config. The read-only sandbox derailed a few runs into "give me write access"; `workspace-write` made it worse (Codex began implementing rather than asking), so I kept read-only for comparability and excluded one unusable patched run. I graded my own runs unblinded, with the arms known — the criteria were fixed in advance, but that is not an independent verifier. n=15 per arm is small for behavior this variable.

Raw transcripts for all 40 runs are available on request.

## Rigor

Adversarial notes: the scenario is built so the user never answers and never hints — it asks only for trade-offs, which is the case the existing rules do not cover. Both arms were run under identical fixtures, prompts, and grading. Runs where the agent never asked a design question at all were treated as unmeasurable rather than quietly counted as passes. I ran the stock arm first and graded it before seeing patched results in the first Codex batch, to limit grading drift. No tuned content was touched — the diff adds a line and changes none.

## Human review

@Yash-Chindam reviewed the full diff — a single added line in `skills/brainstorming/SKILL.md`, no deletions, no modifications to existing lines.

### DIFF
--- a/skills/brainstorming/SKILL.md
+++ b/skills/brainstorming/SKILL.md
@@ -169,6 +169,7 @@ is the whole process.
 - For appropriately-scoped projects, ask questions one at a time to refine the idea
 - Prefer multiple choice questions when possible, but open-ended is fine too
 - Only one question per message - if a topic needs more exploration, break it into multiple questions
+- A request for context is not an answer - if the user asks for trade-offs, implications, or examples instead of choosing, give them that and then re-ask the same question. It stays the only open question until they decide; don't pair it with a new one
 - Focus on understanding: purpose, constraints, success criteria
 
 **Exploring approaches:**


iskandersierra (commented):
The update seems to be to the point and simple enough not to break any previous intended behavior. Without an eval it is hard to measure objectively the impact in the request.

Yash-Chindam:
@iskandersierra thanks for reviewing. On the eval point — there is one, it is in the PR description above (easy to miss from the Files-changed view). Short version:

**Codex CLI 0.153.4, 15 runs per arm, question tool enabled:**

| Arm | Held the open question |
|---|---|
| Stock `dev` @ `fd02874` | **2/15** (13%) |
| With the bullet | **7/15** (47%) |

Same fixture, same two prompts every run: the notifications request, then "Before I pick - what are the trade-offs there, and what would each option mean for this app?" — which answers nothing and states no preference. Graded on whether the agent's next request for input was the same decision or something else; offers to advance counted as fails. Arms were separate plugin roots, verified before each batch by grepping the installed `SKILL.md`.

Fisher's exact: **p = 0.109** as graded, 0.050 excluding one unusable run, 0.215 under the strictest grading. So — suggestive, not conclusive, and the PR says so rather than claiming the eval settles it. On Claude Code / Opus 5 the bullet changed nothing (4/5 both arms); that baseline already holds the question, which is why this only shows up on Codex.

One caveat that matters for your setup specifically: Codex ships `default_mode_request_user_input` set to `false`. With it off, Codex has no tool to ask a question mid-task and substitutes "If you want, I can…". In that configuration the bullet is worth roughly nothing (0/5 vs 1/5). All the numbers above are with the flag on. Worth checking which side of that your sessions were on — `codex features list | grep request_user_input`.

**Your `AGENTS.md` block looks like it does something this bullet does not, and I think you are ahead of the PR here.** Every one of the 7 remaining failures in the patched arm was an offer to move to planning:

> "If you want, I can turn that into a concrete design next."
> "If you want, I can next map that into a concrete product scope."

Your version closes that with **"Do not plan or implement until the shared understanding is confirmed"** — this PR's wording only says re-ask and don't pair with a new question, so an agent can comply with it while still pitching the plan. That would explain why the patched arm plateaued at 47% instead of higher.

Happy to run the same rig on a wording that adds your clause and post the comparison, if that is useful before this lands — it is the same setup, so it is only a matter of runs. Raw transcripts for all 40 sessions available either way.

iskandersierra:
Looks good to me

PR #2265, opened 2026-09-06
writing-plans: make execution-handoff recommendation contextual

senojed:
The Execution Handoff section hardcodes **"Subagent-Driven (recommended)"**.

For plans whose tasks are tightly coupled — shared interfaces/state, later tasks building heavily on earlier ones — inline execution is usually the better call: a cold subagent per task spends a spawn re-deriving context (spec, architecture, conventions) that the parent already holds, and the plan's own `## Task Right-Sizing` / `Interfaces` blocks are written for a reader with continuity.

This change stops hardcoding a winner and has the model pick the recommendation from the plan in front of it, with a short criterion for each option. Diff is +9/−4, single file.

### DIFF
--- a/skills/writing-plans/SKILL.md
+++ b/skills/writing-plans/SKILL.md
@@ -152,15 +152,20 @@ If you find issues, fix them inline. No need to re-review — just fix and move 
 
 ## Execution Handoff
 
-After saving the plan, offer execution choice:
+After saving the plan, offer the execution choice — and make an actual recommendation for THIS plan, don't hardcode one.
 
 **"Plan complete and saved to `docs/superpowers/plans/<filename>.md`. Two execution options:**
 
-**1. Subagent-Driven (recommended)** - I dispatch a fresh subagent per task, review between tasks, fast iteration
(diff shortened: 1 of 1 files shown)


barryodonovan:
Independent confirmation this fix works, from a separate session that hit the same problem and drafted an equivalent patch before finding this PR already open.

Before/after with a subagent (Claude Code, Sonnet 5), given only the relevant SKILL.md section and a fixed scenario: a 2-task plan where Task 2 depends directly on class names Task 1 introduces, both touching the same 2 files (textbook case for inline execution).

**Baseline (current SKILL.md on `dev`):** verbatim parroted "1. Subagent-Driven (recommended)" — wrong call for this plan, no evaluation happened.

**With this PR's wording applied:** "Recommended for this plan: Inline Execution — the tasks are tightly coupled with Task 2 depending directly on class names introduced in Task 1, both touching the same files." Correct call, with the actual reasoning stated.

No further changes needed on my end — this diff already does what I was about to submit.

---
Disclosure: posted by Claude Code (Sonnet 5) on behalf of @barryodonovan, who reviewed this comment before it was sent.

Issue #2181, opened 2026-08-20
Hermes plugin installation gets blocked with a dangerous verdict


AkumaTen:
- [x] I searched existing issues and this is not a duplicate

## Environment (required)

| Field | Value |
|-------|-------|
| Superpowers version |v6.3.0 |
| Harness (Claude Code, Cursor, etc.) |hermes agent cli |
| Harness version | v0.20.4|
| Your model + version | local qwen 3.8 27b|
| All plugins installed | |
| OS + shell | vs code container in wsl2 Ubuntu noble|

## What happened?
the installation of the plugin as suggested in the doc (`hermes plugins install obra/superpowers --enable`) is blocked with the following message:

**Blocked: Security scan blocked plugin install: Blocked (dangerous verdict, 213 findings). --force does not override a dangerous verdict**

## Expected behavior
the plugin installs with no blocks

jackmawer:
As a temporary workaround, you can disable security scanning while installing the plugin, and then re-enable it afterwards:

```sh
# 1. Temporarily disable plugin scanning
hermes config set plugins.scan_on_install false

# 2. Install the reviewed, immutable revision and enable it
hermes plugins install obra/superpowers \
  --enable \
  --ref b36e0829c6d0140e93cfef2ca599b1b07d4a7797

# 3. Restore the protection for future installs
hermes config set plugins.scan_on_install true

# 4. Verify the plugin is present and enabled
hermes plugins list
```

AkumaTen:
that works! My concern of course is to be 100% sure that everything is a false positive

dajiaohuang:
The dangerous verdict is produced by the Hermes install scanner, and I have not independently validated the 213 findings as false positives. Documenting a scan-disable workaround would weaken the protection before that classification is known. Would maintainers prefer this to be investigated in Hermes with a sanitized scan report, or is there a specific Superpowers artifact they want minimized? I do not see a safe core change without that decision.

AkumaTen:
Could one of the maintainers please take a look at this? It is quite concerning that the official Hermes plugin triggers this warning. First, we should confirm if this is a false positive

AlexMili:
Same issue, here is the details:

<details>
<summary>Full log output</summary>

```
Blocked: Security scan blocked plugin install: Blocked (dangerous verdict, 213 findings). --force does not override a dangerous verdict.

Scan: plugin (obra/superpowers/community)  Verdict: DANGEROUS
[...]

PR #2162, opened 2026-08-16
fix(using-superpowers): correct Hermes tool mapping

SomSamantray:
## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | OpenCode host harness. The harness does not expose the serving model ID to the agent; no model version is claimed. |
| Harness + version | opencode CLI (agent harness), macOS |
| All plugins installed | compound-engineering plugin (CE skills pipeline only; no repo behavior) |
| Human partner who reviewed this diff | Yes — Som Samantray reviewed the complete diff and approved submission (2026-08-16). |

## What problem are you trying to solve?

Issue #2157: the Hermes Agent tool mapping in `skills/using-superpowers/references/hermes-tools.md` is stale and internally inconsistent with the current Hermes runtime (v0.20.1):

1. `delegate_task` is documented with a `toolsets=[...]` parameter the runtime does not expose. The upstream source at the issue's pinned commit (`165c889e`) states verbatim: "the model has no toolsets argument. Subagents inherit the parent's toolsets." An agent following the mapping constructs an invalid tool call.
2. `skill_view("brainstorming")` and `skill_view("test-driven-development")` fail at runtime with `Skill 'brainstorming' not found.` Plugin skills register under the qualified `superpowers:` prefix — the plugin's own bootstrap already instructs `skill_view("superpowers:skill-name")`, so the mapping contradicts it.
3. The mapping unconditionally directs agents to `web_extract` and `web_search`, which only exist when the session's toolsets and providers expose them.

## What does this PR change?

- Drops the toolset-restriction parameter from the `delegate_task` guidance and documents that subagents inherit the parent session's toolsets.
- Qualifies every `skill_view` example with the `superpowers:` prefix and adds a short runtime-registration verification note.
- Makes the web/search rows capability-aware instead of naming tools that may be absent.
- Adds a regression-guard test (`tests/hermes/test_tools_mapping.py`) that fails if the stale tokens are reintroduced.

## Is this change appropriate for the core library?

Yes. The Hermes tool mapping is general-purpose plugin infrastructure shipped to every Hermes user; the defects affect any session where an agent follows the mapping. No third-party dependencies are added.

## What alternatives did you consider?

[...]

Issue #2157, opened 2026-08-15
[Bug]: Hermes tool mapping in v6.3.0 uses stale names and an unqualified plugin skill name


ovijincom:
- [x] I searched existing issues and this is not a duplicate.

## Environment

| Field | Value |
|-------|-------|
| Superpowers version | v6.3.0 |
| Superpowers commit | `b36e0829c6d0140e93cfef2ca599b1b07d4a7797` |
| Harness | Hermes Agent desktop / native plugin |
| Harness version | Hermes Agent v0.20.1 (2026.8.13), upstream `165c889e` |
| Your model + version | OpenAI Codex / gpt-5.6-terra |
| Relevant plugin | `superpowers v6.3.0`, enabled, source: user |
| OS + shell | Windows 11, Git Bash / MSYS terminal |

## Is this a Superpowers issue or a platform issue?

- [x] I confirmed this issue does not occur without Superpowers installed.

This is a Superpowers issue: the incorrect instructions are shipped by the
Hermes-specific reference file in this repository. Hermes itself exposes its
current runtime schema; the drift is in the mapping that Superpowers injects
into the agent context.

Related: #1581

## What happened?

The Hermes mapping in
`skills/using-superpowers/references/hermes-tools.md` is stale and internally
inconsistent with the current Hermes runtime.

The affected lines are:

```text
15 | Fetch a URL / read a webpage | web_extract(urls=[...])
16 | Search the web | web_search(query=...)
17 | Dispatch a subagent | delegate_task(..., toolsets=[...], role="leaf")
31 | skill_view("brainstorming")
49 | delegate_task(..., toolsets=[...], role="leaf")
```

### 1. Wrong `delegate_task` parameter

Current Hermes Agent v0.20.1 exposes `enabled_toolsets`, not `toolsets`, for
the delegated worker toolset restriction.

The current Hermes source and live tool schema use:

```python
delegate_task(
    goal="...",
    context="...",
    enabled_toolsets=["..."],
    role="leaf",
)
```

An agent following the Superpowers mapping verbatim can construct an invalid
tool call.

### 2. Wrong skill identifier in the same Hermes integration

The reference tells the agent to load an unqualified skill:

```python
skill_view("brainstorming")
```

In the actual plugin runtime that fails:

```text
Skill 'brainstorming' not found.
```

The qualified identifier works:

```python
skill_view("superpowers:brainstorming")
```

This also conflicts with the bootstrap generated by the plugin itself, which
instructs agents to use:

[...]

PR #2284, opened 2026-09-11
fix(hermes): pass each skill's frontmatter description to register_skill

pepijn-blom:
# What this fixes

Every skill this plugin registers appears in Hermes' skill catalogue with an
**empty description**, so an agent looking at the catalogue cannot tell what any
of the 14 skills do, or when to load one.

Concretely, before this change, a Hermes session registers the plugin and sees:

```
superpowers:brainstorming                 -> ""
superpowers:dispatching-parallel-agents   -> ""
superpowers:systematic-debugging          -> ""
... 14 skills, 14 empty descriptions
```

After:

```
superpowers:brainstorming                 -> "You MUST use this before any creative work - ..."
superpowers:dispatching-parallel-agents   -> "Use when facing 2+ independent tasks that can ..."
superpowers:systematic-debugging          -> "Use when encountering any bug, test failure, or ..."
```

The cause is one argument. Hermes reads a skill's catalogue description from
`register_skill(description=...)`, and that parameter defaults to `""`. The
integration was passing a name and a path and nothing else:

```python
ctx.register_skill(name, Path(skill_md))          # description defaults to ""
```

The skills themselves were never the problem: every `SKILL.md` carries a
description in its frontmatter, it just never reached the catalogue.

# The diff at a glance

| Change | Effect |
|--------|--------|
| New `_frontmatter_value(content, key)` in `.hermes-plugin/__init__.py` | Reads one flat `key: value` scalar out of the frontmatter block. No YAML dependency; same regex approach as the module's existing `_strip_frontmatter`. |
| `ctx.register_skill(name, Path(skill_md), description=description)` | The actual fix: the description now reaches the catalogue. |
| `tests/hermes/conftest.py`: `register_skill(name, path, description="", frontmatter=None)` | The harness double only accepted `(name, path)`, so it could not have caught this drift from the real Hermes signature. |
| `tests/hermes/test_plugin.py`: 6 new tests | One asserts every registered skill carries a description matching its `SKILL.md`; one unit-tests the parser for quoted, bare, missing-key and no-frontmatter cases. |

# Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | `deepseek/deepseek-v4.1-flash:nitro` via OpenRouter |
| Harness + version | Hermes Agent v0.21.1 (2026.9.7), upstream `698ab88a`, CLI on Ubuntu 26.04.1 |
| All plugins installed | `superpowers` 6.3.0 (git, pinned `b36e0829`), `herdr-agent-state` 1.0 |
| Human partner who reviewed this diff | Pepijn Blom (@pepijn-blom) reviewed the complete diff and approved submission (2026-09-11) |

Drafted by an agent under human review, with the environment fully disclosed
above. The bug was found in a live Hermes session, not by trawling the tracker.

# What problem are you trying to solve?

No issue filed; reproducing it in full. Say the word and I'll open one first if
that's the preferred order.

`tools/skills_tool.py` builds the `skills_list` payload from
`list_plugin_skill_metadata()`, which returns the `description` recorded for
each `register_skill` call. Because the argument defaulted to `""`, this
plugin's 14 catalogue entries carried no description at all.

Why it matters rather than being cosmetic: the catalogue is the surface a model
selects from. Shipping 14 skills that are listed with no description means the
agent has nothing to match a task against when deciding which skill applies,
which works directly against what these skills exist for. It is also
self-inconsistent, since the descriptions are right there in the files.

# Is this change appropriate for the core library?

Yes. This is plugin infrastructure shipped to every Hermes user, not repo
behaviour, and the defect is visible to any Hermes session with this plugin
installed. No dependencies are added.

# What alternatives did you consider?

- **Passing the whole parsed frontmatter mapping too.** Rejected as unnecessary
  surface: Hermes reads the catalogue description from the `description`
  argument specifically, and stock frontmatter carries only `name` and
  `description`.
- **Using a YAML parser to read the frontmatter.** Rejected: this repo is
  deliberately harness-agnostic and declares no Python dependencies. The
  existing module already parses frontmatter with a regex.
- **Restructuring the registration loop.** Rejected; the change is three lines
  inside the existing loop, so the diff stays reviewable.

# Does this PR contain multiple unrelated changes?

No. One fix plus its tests. The `conftest.py` change is required by the fix,
not incidental.

# Existing PRs

I reviewed open and closed PRs and issues touching the Hermes integration.
Related but non-overlapping: #2157 / #2162 (stale tool mapping in
`skills/using-superpowers/references/hermes-tools.md`) and #2181 (install
verdict). None touch skill registration or the `description` argument, and this
PR does not modify `hermes-tools.md`.

# Environment tested

| Hermes Agent (CLI) | v0.21.1 (2026.9.7), upstream `698ab88a` | n/a for the unit tests | `deepseek/deepseek-v4.1-flash:nitro` via OpenRouter |

`pytest tests/hermes/`: **25 passed**. With only `.hermes-plugin/__init__.py`
reverted and the new tests kept, **all 6 fail** (`test_frontmatter_value_parsing`
on the missing attribute, the two integration tests on blank descriptions).

Verified end-to-end against Hermes' own catalogue API rather than only in unit
tests, since that is the code path users actually hit. Both variants of the
plugin were staged into throwaway `HERMES_HOME` directories and loaded through
the real discovery path (`PluginManager.discover_and_load()`), then read back
with `list_plugin_skill_metadata()`, the same call `tools/skills_tool.py` uses
to build `skills_list`:

```
variant : base  (upstream v6.3.0 module)  ->  skills in catalog: 14   blank descriptions: 14
variant : fixed (this PR)                 ->  skills in catalog: 14   blank descriptions:  0
```

The reproducer is a standalone script; happy to paste it in a comment, or to
contribute it under `tests/hermes/` if you would rather have the end-to-end
check in-tree than only the unit-level one.

# New harness support (required if this PR adds a new harness)

N/A -- this PR does not add a harness. It corrects the existing Hermes
integration.

### DIFF
--- a/.hermes-plugin/__init__.py
+++ b/.hermes-plugin/__init__.py
@@ -37,6 +37,25 @@ def _strip_frontmatter(content: str) -> str:
     return (match.group(1) if match else content).strip()
 
 
+def _frontmatter_value(content: str, key: str) -> str:
+    """Return a flat ``key: value`` scalar from a SKILL.md frontmatter block.
+
+    Hermes reads a skill's description from ``register_skill``'s ``description``
+    argument, which defaults to "" — registering only the path left every stock
+    skill with a blank description in the skill catalogue, so the model had
+    nothing to select on. Stock frontmatter is a flat block of single-line
+    scalars, so a line scan covers it without a YAML dependency.
+    """
+    match = re.match(r"^---\n([\s\S]*?)\n---\n", content)
+    if not match:
+        return ""
+    for line in match.group(1).splitlines():
+        name, sep, value = line.partition(":")
+        if sep and name.strip() == key:
+            return value.strip().strip("\"'")
+    return ""
+
+
 def _build_bootstrap(skills_dir: str) -> str:
     with open(
         os.path.join(skills_dir, "using-superpowers", "SKILL.md"),
@@ -82,7 +101,9 @@ def register(ctx):
     for name in sorted(os.listdir(skills_dir)):
         skill_md = os.path.join(skills_dir, name, "SKILL.md")
         if os.path.isfile(skill_md):
-            ctx.register_skill(name, Path(skill_md))
+            with open(skill_md, encoding="utf-8") as f:
+                description = _frontmatter_value(f.read(), "description")
+            ctx.register_skill(name, Path(skill_md), description=description)
 
     # pre_llm_call returning {"context": ...} is the documented injection path
     # (on_session_start return values are ignored, and ctx.inject_message
--- a/tests/hermes/conftest.py
+++ b/tests/hermes/conftest.py
@@ -9,11 +9,12 @@ def mock_ctx():
     ctx = MagicMock()
     ctx._hooks = {}
     ctx._skills = {}
+    ctx._descriptions = {}
 
     def register_hook(event, fn):
         ctx._hooks[event] = fn
 
-    def register_skill(name, path):
+    def register_skill(name, path, description="", frontmatter=None):
         # Mimic hermes' real register_skill, which calls path.exists() and
         # therefore breaks on a str (the bug that silently disabled the whole
         # plugin, found 2026-07-23). Keeping that fidelity here means a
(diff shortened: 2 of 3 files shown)


obra:
Thanks so much for this. We will get it tested

PR #2024, opened 2026-07-23
fix(finishing): check in with human partner when worktree removal hits untracked files

obra:
## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | Claude Fable 5 (`claude-fable-5`) |
| Harness + version | Claude Code 2.1.218 (macOS) |
| All plugins installed | superpowers (dev), superpowers-chrome, superpowers-lab, elements-of-style, episodic-memory, frontend-design, plugin-dev, agent-sdk-dev, mcp-server-dev, code-simplifier, claude-code-setup, context7, linear, github-triage, claude-session-driver, release-radar, summarize-meetings, worldview-synthesis, primeradiant-ops, plus personal skills |
| Human partner who reviewed this diff | Jesse Vincent (@obra) — maintainer; specified this exact design interactively during triage of #2016/#1223 and reviews the final diff on this PR |

## What problem are you trying to solve?

`git worktree remove` refuses when the worktree contains modified or untracked files, and Step 6 of `finishing-a-development-branch` gave no guidance for that refusal. The error message itself suggests `--force`, so the natural agent response destroys files that exist nowhere else. Reported from two independent real sessions:

- **#2016**: a user's plan document under `docs/superpowers/plans/` — untracked for the whole branch lifetime — was permanently deleted at cleanup. Reproduced deterministically during triage: `git worktree remove` refuses, `--force` succeeds, the plan is unrecoverable while committed work survives.
- **#1223**: uncommitted work at finish time produced "what does discard mean for my uncommitted files?" confusion with no answer inside the skill.

Both prior fix attempts baked in a policy: #2016 force-committed plan documents into every user's history (declined — not everyone wants superpowers plans committed), #1223 hard-refused with "commit or stash first, then re-run" (a wall that invites workarounds). Maintainer decision: the destruction point should check in with the human partner instead.

## What does this PR change?

Step 6 now treats a refused `git worktree remove` as a stop-and-ask moment: never `--force` on the skill's own initiative; show `git -C "$WORKTREE_PATH" status --porcelain` output to the human partner and offer commit-to-branch / move-to-main-root / delete, then remove the worktree after the choice. One matching row is added to Common Rationalizations.

## Is this change appropriate for the core library?

Yes — it modifies an existing core skill's cleanup step only. Applies to any project on any harness that uses worktrees; no dependencies, no domain-specific content.

## What alternatives did you consider?

[...]

PR #2294, opened 2026-09-13
fix(finishing): surface surviving Superpowers scratch before worktree removal

### LINKED ISSUE #2207: `finishing-a-development-branch` can silently delete an SDD workspace: ignored files don't make a worktree dirty
**Skills:** `finishing-a-development-branch` (Step 6), interacting with `subagent-driven-development`
**Versions:** reproduced against 6.3.0; the same code path and wording are in 6.2.0
**Environment:** git 2.55.0, Linux

## Summary

Step 6 of `finishing-a-development-branch` protects uncommitted work by relying on
`git worktree remove` to refuse:

> **If removal is refused** (`contains modified or untracked files`): the worktree holds files that
> exist nowhere else — uncommitted plans, notes, or scratch work. Never `--force` on your own
> initiative. Show your human partner what is at stake and ask […]

That guard is keyed on *untracked* files. `git worktree remove` does not count **ignored** files as
untracked, so it does not refuse for them.

`subagent-driven-development` creates its workspace through `scripts/sdd-workspace`, which writes a
self-ignoring `.gitignore` containing `*` at `.superpowers/sdd/`. A worktree whose only extra content
is that workspace therefore reports a clean status, `git worktree remove` succeeds with exit 0, and
the workspace is destroyed without the operator ever seeing Step 6's three-way prompt.

The content at risk is the content the guard exists for: the progress ledger's `Ruling:` lines and the
implementer/reviewer reports. 6.3.0's own §Finish makes the point — "A ruling that dies with the
workspace was a decision made in secret."

## Reproduction

```bash
git init -q r && cd r
git commit -q --allow-empty -m init
git worktree add -q ../w -b feat

# Exactly what scripts/sdd-workspace creates:
mkdir -p ../w/.superpowers/sdd/my-plan
printf '*\n' > ../w/.superpowers/sdd/.gitignore
echo "ledger line that exists nowhere else" > ../w/.superpowers/sdd/my-plan/progress.md

# 1. What Step 6's guard effectively inspects:
git -C ../w status --porcelain -uall
#    → (no output; reads clean)

# 2. Remove it, no --force:
git worktree remove ../w
#    → exit 0, no refusal, no prompt

cat ../w/.superpowers/sdd/my-plan/progress.md
#    → No such file or directory
```

## This is not a git bug

`git worktree remove` documents refusal for a dirty or untracked-file-containing worktree; ignored
files are outside that check. The defect is the skill's inference that a non-refusal means nothing
unique is present in the worktree.

## When it actually bites

In the normal successful path it does not: `subagent-driven-development` §Finish deletes the workspace
itself once the final whole-branch review is clean, so by the time `finishing-a-development-branch`
runs there is nothing left to lose.

It bites when the workspace outlives the SDD session:

- the branch and worktree are kept after the SDD run and the worktree is removed later, by hand or by
  a subsequent run;
- a run is abandoned, or its final review never comes back clean, so §Finish's deletion never happens;
- a session is compacted or resumed and the controller stops before §Finish;
- any external cleanup that removes the worktree.

In those cases the ledger and reports are the only copy, and `--force` was never needed to lose them.

## Related, and worth a second look

The release notes for the `.superpowers/sdd/` move record the adjacent case:

> because the workspace is git-ignored working-tree scratch, `git clean -fdx` will delete the progress
> ledger; recover from `git log` if that happens.

`git log` recovers *which tasks completed* — the commits the ledger names. It does not recover the
ledger's rulings, the parked/adjudicated findings, or the implementer and reviewer reports, none of
which are in git at all. The recovery advice holds for progress state but not for the decision record
that 6.3.0 added the "Rulings I made" step to protect.

## Possible directions

Offered as options, not a preferred fix:

1. **Widen Step 6's pre-removal check to consider ignored content**, e.g. via

   ```bash
   git -C "$WORKTREE_PATH" status --porcelain -uall --ignored=matching
   ```

   This would need filtering or classification rather than feeding the existing prompt directly:
   worktrees routinely hold ignored build output, dependency trees, and caches that are safely
   disposable, and prompting on all of it would be noise that trains the operator to dismiss the
   prompt.

2. **Check specifically for a surviving SDD workspace** — the one ignored artifact Superpowers itself
   creates, and therefore the targeted fix with the lowest false-positive rate: if
   `$WORKTREE_PATH/.superpowers/sdd/` is non-empty, do not remove the worktree silently.

3. **Say what to do when the workspace outlives the run.** `subagent-driven-development` §Finish
   currently covers only the delete-at-final-review-clean case; a sentence on preserving the workspace
   when the run ends any other way would close the gap from the other side.

(2) and (3) are complementary — one guards the boundary, the other tells the controller what the
boundary means.

## How this was found

A real session's workspace survived past §Finish and was copied out of the worktree by hand before
removal, on the assumption that removal would refuse. The check above showed removal would not have
refused, which is what prompted this report.

Oscar-Williams:

| Field | Value |
|-------|-------|
| All plugins installed | None — worked against a plain checkout of `dev` |
| Human partner who reviewed this diff | Oscar (@Oscar-Williams) — reviewed the complete proposed diff and the verification transcript, required a full diff review before submission (which surfaced the `.superpowers/` scope widening and the `find -type f` change), and approved submission |

> Targets `dev`.

Fixes #2207

Step 6 of `finishing-a-development-branch` protects uncommitted work by relying on `git worktree remove` refusing — but that refusal only covers modified and untracked files. `scripts/sdd-workspace` deliberately gives `.superpowers/sdd/` a self-ignoring `.gitignore`, so a worktree whose only extra content is an SDD workspace reports a clean status, removal succeeds with exit 0, and the plan's progress ledger, rulings, and implementer/reviewer reports are destroyed silently. The issue's repro reproduces exactly as written (status empty → remove exit 0 → `progress.md` gone). Brainstorm mockups under `.superpowers/brainstorm/` hit the same gap when the user ignores `.superpowers/` as the brainstorming skill suggests.

## What does this PR change?

Step 6 now checks for surviving Superpowers scratch under `.superpowers/` *before* `git worktree remove` and shows the human partner the same show-and-ask prompt style the refusal branch already uses (move into the main repo root / keep the worktree / delete). `subagent-driven-development`'s Finish section now says to leave the workspace in place when a run ends any other way. Two files, +33/−1, prose only.

Yes — both files are core skills, the scratch layout is created by core `scripts/sdd-workspace` and the brainstorming skill, and the guard protects any user's unique decision records and mockups, not a specific project or third-party integration.

- Widening the pre-removal check to all ignored content (`git status --porcelain -uall --ignored=matching`): rejected — worktrees routinely hold reconstructable ignored output (dependency trees, caches), and prompting on all of it trains the operator to dismiss the prompt. The issue reporter flagged the same concern; the `.superpowers/`-scoped check kept the noise profile at zero in the tests below.
- Fixing only the SDD §Finish side (what to do when a run ends early): keeps the boundary unguarded whenever the workspace outlives the run for any other reason. The two changes here are complementary; each is one focused block.

No — both edits close the same gap from its two ends (the boundary guard and the controller's instruction).

## Existing PRs
- Related PRs: #2024 (merged — introduced the removal-refusal prompt this PR extends to the ignored case), #1223 (closed — Step 1 clean-tree guard and cwd handling, different section and problem). No PR references #2207 or fixes ignored-content loss.

## Environment tested

| Plain bash + git checkout (no agent harness) | Windows 11, Git Bash | same as disclosed above | — |

## New harness support

N/A — this PR does not add harness support.

## Evaluation
- Initial prompt: triaging #2207 — the issue's 10-line repro is the failing case.
- Behavior-level verification on real worktrees (6 scenarios): ignored-only SDD workspace (check fires, workspace preserved), clean worktree (silent, no false positive), ignored build junk without `.superpowers/` (silent — no noise regression vs. an `--ignored=matching` approach), empty `.superpowers/sdd/` shell left after §Finish (silent), a workspace created by the real upstream `scripts/sdd-workspace` (fires), and brainstorm mockups only (fires).
- Pressure scenario per writing-skills (two subagents, same prompt except the Step 6 text): **without the fix**, the agent removed the worktree with no prompt, treating an earlier "looks good, wrap it up" as covering the scratch — the ledger was destroyed. **With the fix**, the agent ran the check, listed the three scratch files, and stopped at the three-way prompt with the worktree and all files intact.
- Not done: multi-session evals inside a superpowers-enabled harness — the authoring environment has no such harness. The changed decision rule is verified at the git-behavior and subagent level above; the prose around it reuses #2024's existing prompt structure unchanged.

## Rigor

      — *testing methodology followed from this repo's `skills/writing-skills` (baseline vs. with-skill pressure scenario, §Testing); both runs are pasted under Evaluation. The full authoring workflow was not needed for a guard addition of this size, and no plugin install was available in this environment.*
      — *no tuned table, rationalization row, or existing prompt line was modified; the new block only reuses the existing show-and-ask structure.*

## Human review
      — *Oscar (@Oscar-Williams) reviewed the full diff (patch against `dev` @ 5940bd8) and the verification transcript before this PR was opened, and confirmed submission after the two review fixes.*

**Personal verification (@Oscar-Williams, Windows 11, git 2.55.0.windows.2):** I re-ran the issue's reproduction myself — `git status --porcelain -uall` in the worktree was empty, `git worktree remove` exited 0 without refusal, and the plan's `progress.md` was gone. I then checked out this branch and ran the new pre-removal check (`find "$WORKTREE_PATH/.superpowers" -type f`) against the same setup — it listed the scratch files, so Step 6 now stops and asks instead of deleting.

### DIFF
--- a/skills/finishing-a-development-branch/SKILL.md
+++ b/skills/finishing-a-development-branch/SKILL.md
@@ -167,7 +167,33 @@ Step 2, from before that directory change.
 **If `GIT_DIR == GIT_COMMON`:** Normal repo, no worktree to clean up. Done.
 
 **If `WORKTREE_PATH` is under `.worktrees/` or `worktrees/`:** Superpowers
-created this worktree — we own cleanup:
+created this worktree — we own cleanup. First check for surviving
+Superpowers scratch: `scripts/sdd-workspace` gives `.superpowers/sdd/` a
+self-ignoring `.gitignore`, and brainstorm mockups live under
+`.superpowers/brainstorm/`. None of it makes the worktree dirty, so
+removal will not refuse for it:
+
+```bash
+find "$WORKTREE_PATH/.superpowers" -type f 2>/dev/null
+```
+
+**If that lists anything**: the worktree holds plan workspaces, progress
+ledgers, rulings, implementer/reviewer reports, or design mockups that
+exist nowhere else. Show your human partner what is at stake and ask:
+
+```
+Worktree holds Superpowers scratch that was never finished:
+
+<file list>
+
+1. Move it into <main repo root> before cleanup
+2. Keep the worktree in place
+3. Delete it (unrecoverable)
+
+Which?
+```
+
+Carry out the choice — option 2 stops here — then remove the worktree:
 
 ```bash
 git worktree remove "$WORKTREE_PATH"
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -484,6 +484,12 @@ delete this plan's workspace (`rm -rf <workspace>`) — the git history is
 the record now. Sibling directories belong to other plans; leave them
 alone.
 
+If the run ends any other way — abandoned, compacted mid-run, or stopped
+before the final review came back clean — leave this plan's workspace in
+place. The ledger and reports are the only copy of what was decided;
+finishing-a-development-branch preserves a surviving workspace instead of
+silently removing it with the worktree.
+
 Use superpowers:finishing-a-development-branch.
 
 ## Common Rationalizations


Oscar-Williams:

Issue #2092, opened 2026-08-05
test-driven-development: a test whose arrange step silently fails passes every gate (mental mutation check, optional SDD red evidence)


glen-84:
- [x] I searched existing issues and this has not been proposed before

## What problem does this solve?

A test can be vacuous because its **arrange step silently never creates the condition it claims to exercise**. Real code, no mocks, a hand-written literal expectation, a correct `Unit_Scenario_Expectation` name — and it asserts nothing, because the input was never what the test says it is. Nothing in the current skills catches this, and in `subagent-driven-development` it clears every gate.

I hit this executing a plan with SDD. A .NET storage class strips a UTF-8 byte order mark from a file before hashing its contents into a URI. The test:

```csharp
await File.WriteAllBytesAsync(
    viewPath,
    new UTF8Encoding(encoderShouldEmitUTF8Identifier: true).GetBytes("<div>products</div>"),
    ct);
// ... assert the loaded markup equals "<div>products</div>"
```

`UTF8Encoding.GetBytes` **never emits the preamble** — that flag only affects `GetPreamble()` and stream writers. So the file had no BOM, the assertion compared unstripped text to itself, and the test passed. Deleting the entire stripping implementation kept it green. It was caught weeks later by a human reading the production code, not by any part of the process that produced it.

Where each gate lets it through:

**1. `test-driven-development/SKILL.md`, "Verify RED", names only one of the two causes.**

> **Test passes?** You're testing existing behavior. Fix test.

Either you are testing behavior that already exists, *or the arrange step did not produce the condition*. The stated diagnosis points at the assertion; in this class of bug the defect is in the fixture, so following the instruction as written sends you to the wrong place.

**2. `writing-good-tests.md` covers the expected side and the mocks, but never the input.**

Principle 1 governs expected values ("Derive expectations independently", mirror assertions). Principle 2 governs mocks. A test with a literal expectation, real components, and a broken fixture satisfies both principles completely. There is no rule for: *when the arrange step depends on a platform or third-party API doing something, prove that it did.*

**3. The mutation check is explicitly mental, which cannot catch this.**

> Before finishing, **mentally mutate** the production code; at least one test should fail for each realistic mutation

Mental mutation cannot catch a bad fixture, because you assume your own arrange step works — that assumption *is* the bug. Every listed category is a production-code mutation ("wrong constant", "wrong branch handler", "empty or default return"); none is "the input was never what you think".

[...]

PR #2295, opened 2026-09-13
fix(writing-plans): require produced expected values in verification steps

### LINKED ISSUE #2285: writing-plans: expected values in verification steps are asserted, not produced — and can be contradicted by the plan's own text
## The problem

`writing-plans` requires every step to carry the actual content an engineer needs,
and forbids placeholders like "TBD" or "add appropriate error handling" on the
grounds that a step describing *what* without showing *how* is not a plan.

That rule does not currently extend to the outputs those steps assert. A
verification step of the form:

```bash
grep -c "some-pattern" path/to/file
```

> Expected: `4`

is a placeholder wearing a number. The `4` is a claim about a command the plan
author never ran.

The failure direction is the damaging one. A wrong expectation reports a
**correct** implementation as broken, and pressures whoever executes the plan to
bend working code until it matches the author's guess.

## Evidence from real use

Across three plans written in one session, four stated expectations were wrong.
Three were caught only because the implementers stopped and reported a mismatch
rather than adjusting the work to fit; one was caught in self-review, after the
same class of error had already surfaced twice.

Two were wrong for a reason worth calling out separately: **the plan's own added
text changed what the command would match.** A step instructed the implementer to
insert a paragraph containing a word, and a later verification step grepped for
that word with a count derived from the file *before* the insertion.

A sharper variant appeared in the same set: a verification grep whose pattern
matched a phrase the plan's own inserted text used, so the check could never pass
regardless of how correct the implementation was. The implementer correctly
reported it as unsatisfiable rather than as a failure.

## Proposed change

Two small additions, both in existing sections.

**1. Add to the No Placeholders list:**

> - An expected count, output, or match you did not produce by running the command
>   against the real tree. A stated expected value is an untested assertion, and it
>   fails toward reporting correct work as broken. Either run the command while
>   writing the plan and paste the real value, or describe what the reader should
>   observe ("every reference rewritten, none left bare") and have the step compare
>   two independently derived counts rather than one count against a literal.

**2. Add to Self-Review:**

> **4. Self-contradiction check:** for every verification command, ask whether text
> this plan introduces elsewhere would match it. A step that greps for a word the
> plan itself adds will return a count the author did not anticipate, or can become
> unsatisfiable outright.

## Notes

- Agent-agnostic. Both are prose rules about what a plan may assert; nothing
  depends on a particular harness, tool name, or shell.
- Scoped to `writing-plans`. `executing-plans` and `subagent-driven-development`
  consume these plans but do not author the expectations, so the rule belongs
  upstream of both.
- If useful, I can open this as a PR against `dev` instead.

Oscar-Williams:
> Targets `dev`.

| Field | Value |
|-------|-------|
| All plugins installed | None — plain checkout of `dev` |
| Human partner who reviewed this diff | Oscar (@Oscar-Williams) — reviewed the complete diff and the verification transcript, approved submission |

Fixes #2285

A verification step like `grep -c "pattern" file` / `Expected: 4` asserts the result of a command the plan author never ran. A wrong expectation reports correct work as broken and pressures the implementer to bend working code to match the guess (#2285: four wrong expectations across three plans, two caused by the plan's own inserted text).

## What does this PR change?

Two prose additions to `writing-plans`, as proposed in #2285: the No Placeholders list rejects expected values that were not produced by running the command (paste the real value, or compare two independently derived counts), and Self-Review gains a fourth check for verification commands contradicted by the plan's own text. +3 lines.

Yes — core skill, harness-agnostic rule; `executing-plans` and `subagent-driven-development` consume plans but never author expectations, so the rule belongs upstream.

- A run-time check in executing-plans for unsatisfiable verifications: wrong owner — only the author can produce the real value.
- Requiring pasted command output for every step: too heavy for qualitative verifications, which the new rule allows in descriptive form.

No — both additions are the authoring-time and review-time halves of one rule.

## Existing PRs
- Related: none found for #2285; nearest issues #2092/#2046 are different defects in different mechanisms.

## Environment tested

| Plain document review (no harness) | Windows 11, Git Bash | same as disclosed above | — |

## New harness support

N/A — no harness support added.

## Evaluation
- Failing case: #2285's own example, built as a two-task plan whose only defects are the two the rule targets.
- Pressure test per writing-skills (two subagents, identical plan and instructions, only the skill text differing): without the fix — "Self-review passed, no issues found"; with the fix — both defects flagged by name (Self-Review #4, No Placeholders) plus a compliant repair (baseline count N, expect N+1).
- Not done: multi-session evals in a superpowers-enabled harness (unavailable here); this was a wording-level micro-test.

## Rigor

      — *methodology followed from this repo's `skills/writing-skills` (baseline vs. with-skill micro-test); runs pasted under Evaluation.*
      — *the diff is three added lines; nothing existing was modified or removed.*

## Human review
      — *Oscar (@Oscar-Williams) reviewed the full diff (patch against `dev` @ 5940bd8) and the verification transcript before submission.*

### DIFF
--- a/skills/writing-plans/SKILL.md
+++ b/skills/writing-plans/SKILL.md
@@ -137,6 +137,7 @@ Every step must contain the actual content an engineer needs. These are **plan f
 - "Similar to Task N" (repeat the code — the engineer may be reading tasks out of order)
 - Steps that describe what to do without showing how (code blocks required for code steps)
 - References to types, functions, or methods not defined in any task
+- An expected count, output, or match you did not produce by running the command against the real tree — run it while writing the plan and paste the real value, or describe what to observe and compare two independently derived counts
 
 ## Self-Review
 
@@ -147,6 +148,8 @@ After writing the complete plan, look at the spec with fresh eyes and check the 
 **2. Placeholder scan:** Search your plan for red flags — any of the patterns from the "No Placeholders" section above. Fix them.
 
 **3. Type consistency:** Do the types, method signatures, and property names you used in later tasks match what you defined in earlier tasks? A function called `clearLayers()` in Task 3 but `clearFullLayers()` in Task 7 is a bug.
+
+**4. Self-contradiction check:** for every verification command, ask whether text this plan introduces elsewhere would match it. A step that greps for a word the plan itself adds will return a count the author did not anticipate, or can become unsatisfiable outright.
 
 If you find issues, fix them inline. No need to re-review — just fix and move on. If you find a spec requirement with no task, add the task.
 


Oscar-Williams:

Issue #2282, opened 2026-09-11
subagent-driven-development: implementer dispatches are gated on nothing — briefs get referenced unwritten and todo state never updates


Ecredes:
## Two observed failures from real SDD sessions

Observed across multiple orchestrator sessions on 6.2.0; both verified still present on `main` (`b36e082`). Both are silent — execution proceeds and the failure surfaces downstream. They share one shape, so I'm filing them together: **the skill states these steps as prose side-effects instead of gating the dispatch on them.**

### 1. The task brief gets referenced without being written

The dispatch guidance says to run `scripts/task-brief PLAN_FILE N` before dispatching, and the implementer template has a `[BRIEF_FILE]` placeholder. But nothing connects the two: the template never says where `[BRIEF_FILE]` comes from, and nothing makes the controller observe the script's output before dispatching.

In practice: under continuous execution ("Do not pause to check in", narrate one line max), the controller sometimes composes the dispatch straight from the template and fills `[BRIEF_FILE]` with the remembered default path — without ever running the script. The implementer then reads a nonexistent file. It's intermittent because it's attention-dependent, and compaction makes it worse: the step-by-step prose is exactly what gets summarized away. The ledger was built to survive that treatment; the brief step wasn't.

The subagent can't recover either — the template says "Read your task brief first" but has no fail-fast, so a missing brief becomes guessing, or reading the whole plan (which the skill forbids), or a `NEEDS_CONTEXT` far downstream from the cause.

(#2014 fixed the script's side — empty/overwritten briefs — and #2175 the awk pattern. This is the remaining layer: the controller never invoking the script at all.)

### 2. Todo state never updates

The skill says "create a todo per task" (Setup) and "mark the todo complete" (step 5) — but there is no `in_progress` step anywhere, and completion marking is trailing bookkeeping in the same bullet as the ledger write. The sibling `executing-plans` skill does say "Mark as in_progress" per task; SDD lost that in its rewrite. Under continuous execution the controller optimizes for the next dispatch and skips the "non-essential" bookkeeping call, so the harness todo UI shows a stale list for the whole run and progress is only visible in the ledger.

## Suggested change: gate every dispatch on its preconditions

Make the bookkeeping a precondition of the dispatch — batched in one message, first dispatch included. We've been running this locally (patch against 6.2.0) and it has held:

> **Dispatch preconditions — all three, batched in one message, before every implementer dispatch (the first included). The dispatch is gated on them; they are not trailing bookkeeping:**
>
[...]

Issue #2267, opened 2026-09-08
subagent-driven-development: the progress ledger records completion but not cross-task discoveries, so a usage-limit interruption loses what earlier tasks learned


kaiser-data:
*Found while starting an 18-task run of `subagent-driven-development` (2026-09-08). Filed rather than patched locally because the plugin is version-pinned.*

*Version: **superpowers 6.1.1**.*

Related but, I believe, distinct: #931 (broad session-handoff slash commands) and #2253 (deferred *findings* destroyed at finish). This one is about the ledger's missing channel for cross-task **discoveries** during execution. Close as duplicate if you read them as the same thing.

---

## The gap

**Durable Progress** solves "which tasks are done." Its prescribed line format is:

```
Task N: complete (commits <base7>..<head7>, review clean)
```

That is enough to avoid the failure it was written for — re-dispatching completed tasks after compaction. It is not enough to *resume* a plan, because it carries no record of **what a task discovered that a later task depends on**.

The skill closes the other obvious channel, correctly:

> A dispatch prompt describes one task, not the session's history. Do not paste accumulated prior-task summaries ("state after Tasks 1-3") into later dispatches — a real session's dispatch hit 42k chars of which 99% was pasted history.

So cross-task knowledge has exactly one sanctioned home:

> (3) interfaces and decisions from earlier tasks that the brief cannot know

…and that clause is satisfied entirely from **the controller's conversation context**. Which is the one thing that does not survive a compaction or a usage-limit interruption. The ledger the section exists to provide cannot hold it, because its line format has no slot for it.

## Concrete shape of the failure

The plan I am executing opens with a reconnaissance spike:

- **Task 0** determines whether five upstream registries expose a name-lookup API, and what each returns for a miss. Its findings decide how **Task 8** is written — and whether Task 8 exists at all.
- **Task 15** cannot know an SDK's server class name until an implementer inspects the installed package; the plan says so explicitly and tells the implementer to check first.

Both produce facts that no later task brief contains and that no later implementer can re-derive cheaply. Fifteen tasks separate Task 0 from Task 15. On a plan this size, a usage-limit interruption between them is not an edge case — it is the expected path. (#1152 reports a *simple* plan consuming a full 5h budget in one run.)

After the interruption, a resuming controller reads the ledger, sees `Task 0: complete`, and has no idea what Task 0 found. The implementer's report file holds it — but nothing in the skill tells a resuming controller those files exist, indexes them, or says which of them a given task needs.

[...]

PR #2296, opened 2026-09-13
fix(executing-plans): handle plans that are not code changes in a git repo

Oscar-Williams:
> Targets `dev`.

## Who is submitting this PR? (required)

| Field | Value |
|-------|-------|
| Your model + version | GLM (GLM-5.3-Flash) |
| Harness + version | ZCode agent CLI (Windows 11, Git Bash, git 2.55.0.windows.2) |
| All plugins installed | None — plain checkout of `dev` |
| Human partner who reviewed this diff | Oscar (@Oscar-Williams) — reviewed the complete diff and the verification transcript, approved submission |

Fixes #2290

## What problem are you trying to solve?

Step 1 unconditionally requires a `using-git-worktrees` workspace and Step 3 unconditionally hands off to `finishing-a-development-branch`, so executing a plan whose deliverable is not a code change — or whose working directory is not a git repository — forces the agent to fabricate a git repo or quietly improvise past a required sub-skill. In my baseline run of the exact scenario, the agent ran `git init` in a non-git directory, committed twice, opened and merged a feature branch, and executed the finishing flow — none of it requested.

## What does this PR change?

An entry check at the top of Step 1 ("does this plan produce code changes in a git-tracked project?") routes non-code or non-git plans straight to task execution and a completion report, skipping the worktree step and Step 3; Step 3 states it applies to code-change plans only. +5 lines, prose only — the Step 2 task loop is unchanged, as #2290 notes it is already deliverable-agnostic.

## Is this change appropriate for the core library?

Yes — `executing-plans` is a core skill, and Claude Code's Plan Mode produces plans for any kind of task; the entry condition applies to anyone executing a non-code plan.

## What alternatives did you consider?

- A separate "executing non-code plans" skill: premature — only the git bookends in Steps 1 and 3 differ, per #2290.
- Leaving it to the executing agent's judgment: that is the current state, and it is the failure mode.

## Does this PR contain multiple unrelated changes?

No — both edits are the two ends of the same entry condition (route at Step 1, scope note at Step 3).

## Existing PRs
- [x] I have reviewed all open AND closed PRs for duplicates or prior art
- Related: none found for #2290; #2241 (open) covers a different gap and is untouched. Open PR #1297 reshapes worktree promotion but is stale (last updated July) and does not edit `executing-plans`.

## Environment tested

| Harness (e.g. Claude Code, Cursor) | Harness version | Model | Model version/ID |
|-------------------------------------|-----------------|-------|------------------|
[...]

PR #2297, opened 2026-09-13
fix(sdd): record dispatches in the ledger so outages do not re-dispatch committed work

### LINKED ISSUE #1936: SDD progress ledger carries no plan identity — a second superpowers run in the same repo reads the previous run's `progress.md` and silently skips the new plan's tasks

## Related issues (searched before filing)

#1888 and #1816 report the *artifact* side of this (scratch files never pruned, `task-N-report.md` clobbered across efforts), and PRs #1858 / #1905 namespace briefs/reports per plan. This report is about the *control-flow* side, which those don't cover and the namespacing PRs don't fully fix:

- The resume instruction in SKILL.md trusts an unbound ledger, so following it literally makes a second run **skip the new plan's tasks** — an execution bug, not just stale files. #1858 falls back to the legacy flat directory when no plan identity is supplied, so this path survives that fix.
- The **interleaved case** (run paused mid-plan for debugging in the same working tree, second superpowers flow started, original run resumed) interleaves two runs' ledger lines with no way to attribute them — precisely in the post-compaction situation where the skill says to trust the ledger over recollection.

## Environment (required)

| Field | Value |
|-------|-------|
| Harness version | 2.1.201 |
| Your model + version | claude-fable-5 (report co-written with the agent) |
| All plugins installed | superpowers (claude-plugins-official + superpowers-marketplace) |
| OS + shell | Linux (Ubuntu, kernel 6.17) + bash |

(N/A in the usual sense: the affected files — `.superpowers/sdd/`, `docs/superpowers/specs|plans/` — and the resume logic that misreads them are defined entirely by Superpowers' own skills and scripts.)

## What happened?

Superpowers' durable documents (`docs/superpowers/specs/`, `docs/superpowers/plans/`) and the SDD scratch workspace (`.superpowers/sdd/`) are shared, flat locations keyed only by date or task number. When the workflow runs more than once in the same repo — sequentially (feature A finishes, feature B starts) or interleaved (a run is paused mid-plan to debug on another branch and a second superpowers flow starts) — artifacts from different runs collide, and the controller misreads the previous run's records as belonging to the current run.

Three specific mechanisms:

1. **`.superpowers/sdd/` is one flat directory per worktree, keyed by task number only.** `skills/subagent-driven-development/scripts/sdd-workspace` resolves a single `<repo-root>/.superpowers/sdd/` for all runs. `task-brief` writes `task-<N>-brief.md` and reports are `task-<N>-report.md` — so plan B's Task 1 collides with plan A's Task 1. The script comment acknowledges the sharing ("per worktree; concurrent runs in the same working tree share it"), but the skill text never says what to do when the workspace already holds another run's files. (Artifact side tracked in #1888/#1816.)

2. **`progress.md` ledger entries carry no plan identity, yet the resume logic trusts them unconditionally.** Ledger lines are `Task N: complete (commits <base7>..<head7>, review clean)` — no plan filename, branch, or run id. `subagent-driven-development/SKILL.md` ("Durable Progress") instructs: *"At skill start, check for a ledger … Tasks listed there as complete are DONE — do not re-dispatch them; resume at the first task not marked complete."* Followed literally on a second run, the controller reads the previous plan's ledger and **silently skips the new plan's tasks**. Nothing in `finishing-a-development-branch` clears or archives the workspace, so the stale ledger is always there. This is the core of this report.

3. **Specs and plans have no lifecycle or cross-links.** `brainstorming` saves to `docs/superpowers/specs/YYYY-MM-DD-<topic>-design.md`, `writing-plans` to `docs/superpowers/plans/YYYY-MM-DD-<feature-name>.md`. Date prefixes avoid most filename collisions, but nothing marks a spec/plan completed or superseded, plans don't name their spec, and the ledger doesn't name its plan. A session resuming after compaction — or a new session told to "continue" — has several identical-looking documents and no recorded pointer to the active one. (Per #729, specs/plans are durable history to keep — the ask here is cross-linking/status, not deletion.) The interleaved-run case is worst: worktree isolation is consent-gated in `using-git-worktrees` and mid-task debugging typically happens in place, so both runs share one workspace, overwrite each other's briefs/reports, and interleave ledger lines.

## Steps to reproduce

1. In one repo, run the full flow (brainstorm → write plan A → subagent-driven-development) to completion. Note that nothing cleans up `.superpowers/sdd/`.
2. In a new session, brainstorm and write plan B for a different feature, then invoke `subagent-driven-development` in the same worktree.
3. At skill start the controller runs the prescribed ledger check, finds plan A's `progress.md` with Tasks 1–5 complete, and resumes at "Task 6" of plan B.

Variant: pause an SDD run after Task 2, start a second superpowers flow on a debug branch in the same working tree, then resume the original run — `task-<N>-brief/report.md` for overlapping task numbers now belong to the wrong run, and `progress.md` contains both runs' lines with no way to tell them apart.

## Expected behavior

Each brainstorm→plan→execute run keeps its artifacts separate; the ledger check only trusts completion entries that belong to the plan currently being executed; finished runs leave the workspace clean (or clearly archived) for the next run.

## Actual behavior

The second run inherits the first run's ledger and scratch files: tasks are skipped as "already complete", briefs/reports are overwritten across runs, review packages accumulate unattributed, and resumed sessions can pick up the wrong spec/plan.

No single transcript attached; the failure follows deterministically from the skill text and scripts cited above (`sdd-workspace`, `task-brief`, `review-package`, and SKILL.md's "Durable Progress" section). Happy to reproduce and attach a transcript if useful.

## Suggested direction (optional)

1. **Bind the ledger to its plan** — a small header in `progress.md` (plan path, branch, started date), and amend the resume instruction in SKILL.md: trust completion entries only if the ledger's plan matches the plan being executed; on mismatch, archive the stale ledger (e.g. `progress-<run-id>.md`) and start fresh. This closes the skip bug even for flows that land on the legacy flat directory after #1858/#1905.
2. **Namespace the SDD workspace per run** — as #1858/#1905 do; deriving the run id in `sdd-workspace` from plan filename (date+slug) and/or branch covers `progress.md` and review packages too, not just briefs/reports.
3. **Add cleanup/archival of the scratch workspace to `finishing-a-development-branch`** — scoped to `.superpowers/sdd/` only, leaving `docs/superpowers/specs|plans` as durable history per #729.
4. **Cross-link the chain** — plan header names its spec, ledger names its plan, so a resuming session can walk to the active documents instead of guessing among identical-looking dated files.

### LINKED ISSUE #2293: subagent-driven-development: the ledger records no in-flight task, so an outage mid-task re-dispatches work that is already committed
*Found while setting up a sequential `subagent-driven-development` run (2026-09-12), reading the skill text before dispatching Task 1. Filed rather than patched locally because the plugin is version-pinned.*

*Version: **superpowers 6.1.1**. Quotes below are from that copy of `skills/subagent-driven-development/SKILL.md`.*

Related but, I believe, distinct — close as duplicate if you read them as the same thing:
- **#2267** — the ledger carries no cross-task *discoveries*. That is about knowledge lost from **completed** tasks; this is about there being no record of an **in-flight** one.
- **#2282** — dispatch preconditions are ungated. Its proposal item 3 is "**BASE:** record `git rev-parse HEAD`" — recorded, but only in the controller's context. This issue is that the recording has nowhere durable to go.
- **#1936** (closed) — ledger carries no plan identity. Different axis.

---

## The gap

**Durable Progress** gives the ledger exactly one line format, written at exactly one moment:

> When a task's review comes back clean, append one line to the ledger […]
> `Task N: complete (commits <base7>..<head7>, review clean)`

And one resume rule:

> Tasks listed there as complete are DONE — do not re-dispatch them; resume at the first task not marked complete.

So the ledger is a record of *finished* work only. The window between dispatching an implementer and its review coming back clean — which is where essentially all of the wall-clock time lives — writes nothing. A task interrupted in that window is indistinguishable from a task never started.

That matters because the resume rule's instruction for such a task is to re-dispatch it, and by then the implementer may already have committed.

## Concrete shape of the failure

Sequential SDD, outage (usage limit, crash, Ctrl-C, laptop sleep) during Task 5:

1. Controller records BASE in context, dispatches the Task 5 implementer.
2. Implementer writes tests, implements, **commits**, reports DONE.
3. Session dies before the review package is generated.
4. New controller reads the ledger. Last line is `Task 4: complete`. Task 5 is not marked complete, so per the resume rule it is not DONE.
5. Controller re-dispatches Task 5 against a tree that already contains Task 5's work.

The re-dispatched implementer finds its tests already present and passing. Best case it reports DONE having done nothing, and the task is reviewed properly by luck. Worst case it "fixes" the discrepancy — rewrites the tests it thinks it should have written, or commits a second, divergent implementation on top of the first.

There is a second failure in the same window that is quieter. BASE lived only in the dead controller's context. The skill is explicit about why that matters:

> BASE is the commit you recorded before dispatching the implementer — never `HEAD~1`, which silently drops all but the last commit of a multi-commit task

A resuming controller has no BASE and no sanctioned way to recover one — so the natural move is precisely the `HEAD~1` the skill warns against, and the reviewer silently sees only the last commit of the task.

## Why the existing artifacts do not close it

- **`git log`** — the skill's stated recovery path ("recover from `git log`") shows commits but not which task they belong to, and not whether the task's review ever ran. A task that committed and was never reviewed looks exactly like one that committed and passed.
- **Report files** (`…/task-N-report.md`) are written by the implementer, so their presence does indicate a dispatch happened — but nothing in the skill tells a resuming controller to look for them, and a task interrupted *before* the implementer wrote its report leaves none while still possibly having commits.
- **Todos** do not survive the session, which is the premise of Durable Progress in the first place.
- **#2282's BASE precondition** records the right value at the right moment; it just has no durable home. The two changes compose: gate the dispatch on recording BASE, and write it where it survives.

## Proposal

Give the ledger a second line format, written at dispatch rather than at completion:

```
Task 5: dispatched (base a1b2c3d, brief .superpowers/sdd/task-5-brief.md) 2026-09-12T18:40Z
Task 5: complete (commits a1b2c3d..e4f5g6h, review clean)
```

Then replace the resume rule's binary with a three-state one:

> - A task whose last line is `complete` is DONE. Do not re-dispatch it.
> - A task with **no** line was never started. Dispatch it normally.
> - A task whose last line is `dispatched` was interrupted mid-flight. **Do not re-dispatch it blind.** Run `git log <base>..HEAD` using the base from that line:
>   - **No commits** — the implementer landed nothing. Re-dispatch normally.
>   - **Commits present** — the work exists but was never reviewed. Do not re-implement. Generate the review package with the recorded base (`scripts/review-package <base> HEAD`) and dispatch the task reviewer, entering the normal review loop at that point.

The `dispatched` line is also what makes the "never `HEAD~1`" rule survivable: BASE stops being a fact only the live controller holds.

Two notes from writing this against 6.1.1:

- The line is append-only and the two formats are distinguishable by keyword, so no existing ledger needs migrating — a ledger with only `complete` lines behaves exactly as it does today.
- `git clean -fdx` still destroys the ledger, as the skill already acknowledges. This proposal does not fix that, but it does mean the common case (outage, ledger intact) stops requiring the controller to guess.

## Scope note

I have not run the failure end-to-end — this is a design gap read off the skill text and the resume rule, not a postmortem. What I can state is that the 6.1.1 skill contains no instruction that writes anything to the ledger before completion, and that its resume rule sends an interrupted-but-committed task back through implementation.

Happy to send a PR — the change is a line format, a three-state resume rule, and one sentence in the DONE handling, all within `SKILL.md`.

Oscar-Williams:
> Targets `dev`.

| Field | Value |
|-------|-------|
| All plugins installed | None — plain checkout of `dev` |
| Human partner who reviewed this diff | Oscar (@Oscar-Williams) — reviewed the complete diff and the verification transcript, approved submission |

Fixes #2293

The ledger only records completions and the resume rule is binary, so a task interrupted between dispatch and a clean review looks like one never started — the resuming controller re-dispatches the implementer against a tree that already holds the task's commits, and BASE dies with the old controller's context, inviting the `HEAD~1` mistake the skill warns against. In my baseline run, the controller re-dispatched and rationalized the already-present Task-2 commits as Task 1's work — the exact "second, divergent implementation" worst case #2293 describes.

## What does this PR change?

Dispatches now append `Task <N>: dispatched (base <base7>, brief <brief-path>)` to the ledger, and the resume rule gains the dispatched state: `git log <base>..HEAD` with no commits means re-dispatch normally; commits present means generate the review package from the recorded base and enter the review loop instead of re-implementing. Append-only and keyword-distinguishable, so existing ledgers need no migration. +13/−2, one file.

Yes — this is the SDD controller's own recovery map, harness-agnostic, and the gap applies to any session that dies mid-task.

- Recovering BASE from `git log` heuristics at resume time: rejected — a resumed controller cannot tell which commits belong to which task, and the skill itself calls the HEAD~1 shortcut a silent dropper.
- Waiting for #2282's dispatch precondition first: the two compose, but recording the dispatch is strictly additive on its own.

No — the dispatched line and the resume rule are the write and read sides of one mechanism.

## Existing PRs
- Related: none found for #2293; #2282 is complementary, #2267 and closed #1936 are different ledger axes, as the issue already sorts out.

## Environment tested

| Plain bash + git checkout (no harness) | Windows 11, Git Bash | same as disclosed above | — |

## New harness support

N/A — no harness support added.

## Evaluation
- Failing case: #2293's outage scenario, staged as a real git rig — Task 1 complete, Task 2 `dispatched (base 877bc95, …)`, HEAD two unreviewed commits past base.
- Pressure test per writing-skills (two subagents resuming the same rig, only the resume rule differing): without the fix — re-dispatched the implementer, bending the ledger's attribution to fit; with the fix — ran `git log 877bc95..HEAD`, found the commits, dispatched the task reviewer with a full review framing, no re-implementation.
- Not done: multi-session evals in a superpowers-enabled harness (unavailable here); behavior-level micro-test of the changed resume rule.

## Rigor

      — *methodology followed from this repo's `skills/writing-skills` (baseline vs. with-skill resume decision); results under Evaluation.*
      — *the rig forced the awkward case: commits present, review never run, bookkeeping uncommitted.*
      — *completion-line formats, fix-round rule, and stray-ledger rule unchanged; the resume rule gains one state.*

## Human review
      — *Oscar (@Oscar-Williams) reviewed the full diff (patch against `dev` @ 5940bd8) and the verification transcript before submission.*

### DIFF
--- a/skills/subagent-driven-development/SKILL.md
+++ b/skills/subagent-driven-development/SKILL.md
@@ -141,8 +141,15 @@ a ledger file, not only in todos.
 - Check for this plan's ledger at `<workspace>/progress.md`. If its first
   line names your plan file, tasks with a `Task <N>: complete` line are DONE
   — do not re-dispatch them; resume at the first task without one. A task
-  whose last line is a fix round is mid-loop: resume the loop at the next
-  round. A ledger whose first line names a different plan file — or a stray
+  whose last line is `Task <N>: dispatched (base <base7>, …)` was
+  interrupted mid-flight — do not re-dispatch it blind. Run
+  `git log <base7>..HEAD` with the base from that line: no commits means
+  the implementer landed nothing, so re-dispatch normally; commits present
+  means the work exists but was never reviewed — do not re-implement.
+  Generate the review package (`scripts/review-package <plan-file> <base7>
+  HEAD`) and dispatch the task reviewer, entering the normal review loop.
+  A task whose last line is a fix round is mid-loop: resume the loop at the
+  next round. A ledger whose first line names a different plan file — or a stray
   ledger at the old flat path `.superpowers/sdd/progress.md` — is another
   plan's progress: leave it in place and start your own, fresh.
 - Create the ledger with its identity as the first line:
@@ -260,6 +267,10 @@ and fix-round diffs need it.
   (5) the report-file path and report contract. Exact values (numbers,
   magic strings, signatures, test cases) appear only in the brief. Never
   make a subagent read the whole plan file.
+- **Ledger the dispatch:** before the implementer starts, append
+  `Task <N>: dispatched (base <base7>, brief <brief-path>)` to the ledger.
+  An outage between dispatch and a clean review must leave the resuming
+  controller a record — the resume rule's `dispatched` state reads it.
 - **Report file:** name the implementer's report file after the brief
   (brief `…/task-N-brief.md` → report `…/task-N-report.md`) and put it in
   the dispatch prompt. The implementer writes the full report there and


Oscar-Williams:

Issue #2286, opened 2026-09-11
verification-before-completion: the skill requires fresh evidence but never asks where the evidence came from


DigitalS3A:
## The problem

The skill's core instruction is right and clear: no completion claims without
fresh verification evidence, run the command yourself, read the full output
including exit codes.

It does not say anything about the **provenance** of that output. Three failure
modes fit inside "I ran a command and read the output" while producing a false
claim. All three occurred in a single debugging session.

### 1. Evidence the agent produced by hand, read as evidence the system acted

A diagnostic probe was added to a hook script so it would append a timestamp
every time it ran, in order to distinguish "the hook never fired" from "the hook
fired and its decision was invisible".

In the same shell command that cleared the log, the probe was also exercised by
piping a crafted payload directly into the script, to confirm the probe worked.
That hand-run wrote an entry.

Seeing one entry, the agent reported that the hook had fired through the harness.
It had not. The entry was the agent's own test, run by the agent, minutes earlier.

The command satisfied every existing rule: it was run, the output was read, the
claim came after. The output was about the instrument, not the system.

### 2. An instrument's silence read as a negative result

The same probe later stayed empty, and that was reported as proof the hook never
ran. The probe wrote to a path built from a temp-directory environment variable
that the hook's own process may set differently, or not at all. The silence was
uninterpretable, not negative.

This is the same underlying principle as #2261, which correctly notes that an
empty `grep` result is not proof of absence and proposes `command grep` for that
skill. That fix is right and specific. The general form is not captured anywhere:
**an empty result reports on two possibilities at once — the data is absent, or
the question never got asked — and only the second is a defect the zero
conceals.** It applies to any instrument, not only to a shadowed `grep`.

### 3. A user's existing safety control declared broken

Off the back of those two readings, the agent told the user in bold that their own
command-safety hook "is not running" and had been inert.

It was false. The guard worked. A later test that matched one of its *deny*
patterns was blocked outright.

The evidence could not have supported the claim. It rested on the uninterpretable
silence above, plus a command matching one of the guard's *ask* patterns running
without a visible prompt — which had an obvious unexamined alternative: the
environment silently auto-approves ask-decisions, so an ask and a no-op are
externally identical. Only a deny is externally distinguishable, and no deny had
been tested.

The cost here is asymmetric in a way the other two are not. Telling someone a
working security control is broken spends their trust and invites them to rebuild
something that was never damaged.

## Proposed change

Three rules, all in `verification-before-completion`.

[...]

PR #2323, opened 2026-09-18
fix(writing-skills): GREEN must measure outcome, not only compliance

Chener0121:
## Who is submitting this PR?

| Field | Value |
|-------|-------|
| Your model + version | GLM (Z.ai), accessed via Codex |
| Harness + version | Codex Desktop |
| All plugins installed | Superpowers (dev) |
| Human partner who reviewed this diff | Chener0121 (reviewing before submission) |

## What problem are you trying to solve?

Fixes #2292.

`writing-skills` defines GREEN as "agent complies with skill present" -- a process criterion. For discipline-enforcing skills that is exactly right: compliance IS the outcome. For skills that produce an artifact (a document, a config, a fix), compliance is only a proxy. The issue demonstrates with a controlled trial that the proxy can pass cleanly while the outcome goes entirely unmeasured: three agents complied with every process criterion, yet none produced evidence that the fix actually worked -- while a single unskilled agent ran the real system and showed the defect reproducing and then gone.

## What does this PR change?

Changes to `skills/writing-skills/SKILL.md`:

1. **TDD mapping table** -- split the GREEN row into two:
   - `Test passes (GREEN), compliance` -- agent complies with skill present.
   - `Test passes (GREEN), outcome` -- the artifact achieves what the skill exists for, measured not argued; compliance without this is a proxy.
   - `Watch it pass` now carries both dimensions.

2. **Testing All Skill Types** -- added an **Outcome criteria** line to Technique, Pattern, and Reference skills. Each requires measurable improvement over the no-skill baseline, or an explicit "unverified" statement when measurement is not possible. The Discipline section now states explicitly that compliance IS the outcome there, so the distinction is scoped correctly.

3. **Common Rationalizations** -- added the two rows from the issue: "The agent followed the skill, so the skill works" and "The reasoning was careful, so the result is sound".

4. **RED-GREEN-REFACTOR GREEN phase, Skill Creation Checklist, and overview** -- all now carry the outcome dimension ("compliance alone is not GREEN" for artifact-producing skills; record explicitly when the outcome is unverified rather than staying silent).

The same changes are mirrored in `skills/writing-skills/testing-skills-with-subagents.md` (overview line, GREEN phase paragraph, GREEN phase checklist, Verify-GREEN quick-reference row) so the two files teach one consistent standard.

## Is this change appropriate for the core library?

Yes. This tightens the testing methodology of an existing core skill. It does not add a new skill, domain, or third-party integration.

## What alternatives did you consider?

[...]

Issue #2261, opened 2026-09-05
receiving-code-review: "grep codebase for actual usage" can propose deleting live code (an empty result is not proof of absence)


ethads:
- [x] I searched existing issues and this is not a duplicate

Searched open and closed issues/PRs for `ugrep`, `grep gitignore`, `binary file`, `invalid UTF-8`, `grep alias`, `grep shell function`, `grep unreliable` — zero results, with a control query (`skill` in title) returning #2233, #512, #2197, so the search itself works.

## Environment (required)

| Field | Value |
|-------|-------|
| Superpowers version | 6.3.0 — verified equal to HEAD `b36e0829c6d0140e93cfef2ca599b1b07d4a7797` |
| Harness | Claude Code |
| Harness version | 2.1.232 |
| Model | Claude Opus 5 (`claude-opus-5`) |
| All plugins installed | superpowers 6.3.0, chrome-devtools-mcp 1.7.0, code-review, firecrawl 1.0.9, github, huggingface-skills 1.0.25, last30days 3.21.1, marketing-skills 1.9.0, watch 0.2.0 |
| OS + shell | macOS 26.2 (arm64), zsh 5.9 |

## Is this a Superpowers issue or a platform issue?

- [ ] I confirmed this issue does not occur without Superpowers installed

**Deliberately not ticked, and here is the boundary.** The under-read is Claude Code's shell shadow and happens with or without Superpowers. What is fixable here is that `receiving-code-review` routes a **delete-or-keep decision** through an instrument that cannot prove absence. The skill is where the empty result becomes an action.

## What happened?

`skills/receiving-code-review/SKILL.md:92`, in the YAGNI Check block:

```
IF reviewer suggests "implementing properly":
  grep codebase for actual usage

  IF unused: "This endpoint isn't called. Remove it (YAGNI)?"
  IF used: Then implement properly
```

An empty search result decides to propose removing code. In Claude Code's Bash tool, `grep` is not the system binary — it is a shell function that shadows `grep` with an embedded ugrep, invoked with `-I --ignore-files`. Those two flags mean:

- `--ignore-files` — every path matched by any `.gitignore` at or below the start point is skipped.
- `-I` — any file the tool classifies as binary (a NUL byte, or invalid UTF-8) is skipped.

So "no usage found" is returned identically for *genuinely unused* and *present but not searched*. The skill then proposes deleting live code.

[...]

PR #2324, opened 2026-09-18
fix(verification-before-completion): evidence needs provenance, not just freshness

Chener0121:
## Who is submitting this PR?

| Field | Value |
|-------|-------|
| Your model + version | GLM (Z.ai), accessed via Codex |
| Harness + version | Codex Desktop |
| All plugins installed | Superpowers (dev) |
| Human partner who reviewed this diff | Chener0121 (reviewing before submission) |

Fixes #2286.

verification-before-completion required fresh verification evidence but never asked where that evidence came from. The issue documents three real failure modes, all from one debugging session, that fit entirely inside "I ran a command and read the output":

1. A diagnostic probe the agent exercised by hand was later cited as proof the system had acted.
2. The same probe staying empty (output path built from an env var the target process sets differently) was reported as proof the hook never ran.
3. Off that uninterpretable evidence, the agent told the user in bold that their own working safety control "is not running" -- a claim whose cost is asymmetric: it spends trust and invites rebuilding something never damaged.

## What does this PR change?

Implements the issue's three proposed rules inside `skills/verification-before-completion/SKILL.md`, keeping the author's suggested placement (all three ended in an assertion to the user, which is this skill's moment):

1. **Provenance** -- new ATTRIBUTE step in the Gate Function: name the process that wrote the output and when. A hand-run proves the instrument works and nothing else; prove the instrument, clear it, then let the system act.
2. **Absence** -- an empty/zero result is a claim about the instrument until an independent positive probe shows the population is empty in the environment where the system runs; write diagnostics to hardcoded absolute paths.
3. **Claims about the user's tooling** -- before reporting a user's control broken, show a must-fire case failing visibly after enumerating what could swallow the signal; until then the honest report is "I cannot demonstrate it fired", never "it is not running".

Each rule is carried into every section of the skill so they reinforce rather than sit in one paragraph: Common Failures table (3 rows), Red Flags (3 entries), Rationalization Prevention (3 rows), Key Patterns (3 worked patterns), When To Apply (2 new claim categories), and the frontmatter description (attribution requirement).

Also adds `tests/verification-before-completion/test-verification-provenance.sh`: 18 deterministic consistency checks asserting the provenance principle appears in every section of the skill.

Yes. This tightens an existing core skill's verification methodology. It is agent-agnostic -- no rule depends on a harness, tool name, or shell.

- Putting rules 1 and 2 in `systematic-debugging` instead (the issue author offered this) -- kept them here per the author's own reasoning (each failure ended in an assertion to the user), and kept the PR single-scope. If maintainers prefer the split, the rules move cleanly since each is self-contained prose.
- Rule 3 only (the author's stated priority if only one lands) -- implemented all three because they share one invariant (evidence must be attributed before it proves anything) and the consistency test covers them uniformly.

No. One principle (evidence provenance), one skill, one test.

## Existing PRs

- Related PRs: #2261 (empty grep is not proof of absence in receiving-code-review -- referenced by the issue; its specific fix is untouched here, this PR generalizes the principle), #2323 (outcome-vs-compliance for writing-skills -- same family, different skill, no overlap)

## Environment tested

| Docker (alpine:latest, bash + git) | Docker 28.4.0 | N/A (deterministic consistency tests) | N/A |
| Codex Desktop | current | GLM | via Codex |

## Evaluation

- **Initial prompt**: #2286, including the three proposed rules verbatim and the author's placement note.
- **Eval method**: `tests/verification-before-completion/test-verification-provenance.sh` -- 18 checks asserting each of the three rules appears in every applicable section (Gate Function, Common Failures, Red Flags, Rationalizations, Key Patterns, When To Apply, description).
- **Results**: 18/18 pass in a clean Docker container.

## Rigor

- [x] This change was tested adversarially, not just on the happy path -- the test checks per-section coverage of each rule, so a section cannot silently drop a rule and still pass.
- [x] I did not modify carefully-tuned content (Red Flags table, rationalizations, "human partner" language) without eval evidence -- this change ADDS entries to those structures following the issue design; existing entries are untouched.
- [ ] Live LLM pressure testing -- not yet run; the deterministic suite verifies structural completeness. Happy to run A/B pressure scenarios (the issue's three failure modes as temptations) if maintainers want them before merge.

## Human review


Chener0121:
Friendly ping when you get a chance to review -- happy to adjust anything based on your feedback. Thanks!

